PR #5, opened 2014-10-11
Getting L10N Files

AntonBoch1244:
Czech and Russian L10N Files
and Again not all ready download resourcePacks for L10N


dan200:
It is my decision to support localisations only through resource packs

PR #156, opened 2017-05-01
Performance improvements to fs.find

### LINKED ISSUE #89: [1.78] fs.find performance issues
The copy, delete and move programs use `fs.find` to perform actions on files. However `fs.find` still searches the entire computer's filesystem even when a non wildcard argument is specified. For computers with a large number of files this can lead to a large delay.

**Version**
Tested in CC 1.78, but I'd expect this occurs in every version featuring `fs.find`.

**Expected results**
A non wildcard pattern (not containing `*`) would simply check if the file exists.

**Reproduction steps**

``` lua
local start = os.clock()
fs.find("rom")
print("Took: " .. (os.clock() - start)) -- Takes about 2 seconds on my computer
```

This only really matters when the computer has a lot of files - there are about 2500 on my computer.

[Copied from this issue](http://www.computercraft.info/forums2/index.php?/topic/26044-mc-189cc-178ssp-fsfind-is-slow-for-non-wildcard)

A further enhancement might be to start searching from the last folder before the wildcard.

SquidDev:
If the path includes no wildcards then it just checks it exists.

If it does, instead of scanning the entire tree, it works out the last
directory before the wildcard and starts scanning from there.

Closes #89

### DIFF
--- a/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
+++ b/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
@@ -466,9 +466,25 @@ public class FileSystem
     {
         // Match all the files on the system
         wildPath = sanitizePath( wildPath, true );
+
+        // If we don't have a wildcard at all just check the file exists
+        int starIndex = wildPath.indexOf( '*' );
+        if( starIndex == -1 )
+        {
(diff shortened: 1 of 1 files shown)


dan200:
bam

PR #158, opened 2017-05-01
Require the player to be interacting with the computer when typing

SquidDev:
Currently it is possible to fire events on any computer, irrespective of how far away you are from it. Whilst this has it's uses (such as MoarPeripherals's keyboard), it is subject to abuse.

This PR adds a `IContainerComputer` (not to be confused with `IComputerContainer`) which `Container`s can implement. When receiving a packet, the `ServerComputer` will ensure that the player has a container for this computer open, discarding the packet if not.

This will break mods which add additional computers or keyboard-esque items, but I don't see that as a major issue.

**Note:** I'm not 100% happy with this implementation, so any feedback welcome.

### DIFF
--- a/src/main/java/dan200/computercraft/client/proxy/ComputerCraftProxyClient.java
+++ b/src/main/java/dan200/computercraft/client/proxy/ComputerCraftProxyClient.java
@@ -20,6 +20,7 @@ import dan200.computercraft.shared.network.ComputerCraftPacket;
 import dan200.computercraft.shared.peripheral.diskdrive.TileDiskDrive;
 import dan200.computercraft.shared.peripheral.monitor.TileMonitor;
 import dan200.computercraft.shared.peripheral.printer.TilePrinter;
+import dan200.computercraft.shared.pocket.inventory.ContainerPocketComputer;
 import dan200.computercraft.shared.pocket.items.ItemPocketComputer;
 import dan200.computercraft.shared.proxy.ComputerCraftProxyCommon;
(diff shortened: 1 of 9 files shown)


Vexatos:
I agree it's not quite pretty, but it looks like the best possible implementation (at least one of the most sane).

Wojbie:
True. And now that this all is happening keyboard could be in future in mod anyways. 
More security is always good.

JLLeitschuh:
Does this compile with java 8? Perhaps `Optional<IComputer>` instead of using `@Nullable`.
If this is compiling with java 7 then use the guava `Optional`?

Vexatos:
As far as I am aware, this mod still compiles against Java 6.

JLLeitschuh:
The guava libraries that I think forge bundles with has an `Optional` type.

Vexatos:
This is precisely what `@Nullable` is useful for. I don't see the need.

dan200:
Will review this code later as it's a bit more substantial than the other commits so far. The principle seems sound though!

Lignum:
> This will break mods which add additional computers or keyboard-esque items, but I don't see that as a major issue.

Taking the keyboard as an example, would mods that add one be able to be adjusted to implement some sort of "proxy container" that permits them to interact with the computer? Or more simply, would those mods be able to be fixed or are they doomed?

dan200:
I don't have a big objection to breaking mods which weren't API compliant
anyway. The correct way to do a remote keyboard would probably be to have
some kind of "receiver" tileentity that invokes "key" events on the
computer.

PR #159, opened 2017-05-01
Various extensions to the HTTP API

SquidDev:
This adds a couple of additional features to the `http` API.

 - A response is returned on the event of a HTTP error (such as 404).
 - Responses include the response headers.

### Fetching an invalid page:
```lua
local response, err, err_response = http.get("https://google.com/404.txt")
if response == nil then
    printError("Cannot connect: " .. err)
   if err_response then print(err_response:readAll():find("<title>[^<]+</title>")) end
end
```
will display an error message and the title of the page.

### Accessing response headers
```lua
local response, err = http.get("https://google.com/404.txt")
if reponse == nil then
    printError("Cannot connect: " .. err)
else
    print("Connected OK.")
    print(textutils.serialize(response.getResponseHeaders()))
end
displays something like:
```lua
{
  [ "X-XSS-Protection" ] = {
    "1; mode=block",
  },
  Date = {
    "Mon, 01 May 2017 16:33:21 GMT",
  },
}
```

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/HTTPAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/HTTPAPI.java
@@ -69,7 +69,7 @@ public class HTTPAPI implements ILuaAPI
         }
     }
     
-    private static ILuaObject wrapBufferedReader( final BufferedReader reader, final int responseCode, final Map<String, Map<Integer, String>> responseHeaders )
+    private static ILuaObject wrapBufferedReader( final BufferedReader reader, final int responseCode, final Map<String, String> responseHeaders )
     {
         return new ILuaObject() {
             @Override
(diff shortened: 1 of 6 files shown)


dan200:
How common is it in practice for their to be multiple header values in a HTTP response for the same key? If it's rare, I'd prefer to just have a string->string map, concatenating multiples with newlines if necessary. print(headers.Date) is nicer than print(headers.Data[1])

SquidDev:
> How common is it in practice for their to be multiple header values in a HTTP response for the same key? 

I haven't ever noticed it, so probably not very common. From [what I can gather](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), it is preferred to join duplicate fields with commas instead of new lines, would that be OK?

dan200:
That seems fine. Making it a string->string map also mirrors how headers are passed *into* http functions. Also: as this is the first commit which adds a new feature, could you document it in the help files? "http", "changelog" and "whatsnew".

Wojbie:
Kind offtopic but that travis-ci is realy nice sanity check right here.

SquidDev:
@Wojbie Yep. Gotta love CI.

I've added some documentation. I wasn't entirely sure what to do about `rom/help/http` as it doesn't currently describe other methods on the response object (such as `.getResponseCode()`).

apemanzilla:
🎊 🎉

Issue #97, opened 2016-03-13
[1.79pr1/LuaJ] string.rep on negative numbers.


SquidDev:
This is kinda pedantic, sorry. `string.rep` should return an empty string for a negative count. It currently errors with a negative size exception.

moomoomoo309:
[Possibly related?](http://www.computercraft.info/forums2/index.php?/topic/27211-stringformat-2f/page__hl__string.format__fromsearch__1)

SquidDev:
@moomoomoo309 [Dan has said he won't fix that one](https://github.com/dan200/ComputerCraft/issues/91#issuecomment-191826041)

Issue #91, opened 2016-02-27
LuaJ String bugs


SquidDev:
This aren't bugs with ComputerCraft but with LuaJ. I understand that this isn't within the scope of this bug tracker, just thought it might be useful.

 - `string.gmatch` will loop forever on non-consuming patterns (such as `\n*`)
 - `\011` is not counted as whitespace.
 - `string.format` does not pad/truncate on string of floating point format codes (`%s`, `%f`, `%g`, etc...)
 - Balanced matches (`%b`) throw an out of bounds exception on no match

I've got a set of [fixes and test cases](https://github.com/SquidDev/luaj.luajc/commit/8e682b9f98d2246b8dc3b85d68d71bd272a95874) if you want to fix/test this. This doesn't correctly handle the `%g` modifier.

---

I've also [added a fix](https://github.com/SquidDev/luaj.luajc/commit/b009741977ebf3b7d5f998b289e562c82a5467df) for `string.find` returning the wrong index on a previous match (which I guess is what the `string.find` and `string.match` overrides in `bios.lua` are for).

dan200:
Fixed the string.find bug in my LuaJ fork and removed the bios.lua hacks. May merge in the other fixes later.

dan200:
Fixed the infinite loop, whitespace not matching \v (this is actually caused by a typo in the original code, which i corrected), and the out of bounds exception. I will not be implementing the string.format fixes due to code complexity and relative unimportance.

SquidDev:
Awesome! Thanks.

PR #163, opened 2017-05-01
Replace LuaJ with Cobalt

SquidDev:
> It does not add any new dependencies for compiling, running or using the mod.

I realise this is a bit of a long shot as it fails the first criteria, but I do still think this is worth considering.

Cobalt is a fork of LuaJ I started a long time ago to address a couple of bugs (#91, #97) in LuaJ. It's kinda expanded a bit since then, with the following improvements:

 - Many bug fixes - Cobalt passes a significantly larger proportion of the Lua tests than normal LuaJ. This includes fixes to string formatting and weak metatables.
 - No global state: - Unlike LuaJ, Cobalt does not have any global state. This means the string metatable no longer needs to be sandboxed. It also would allow for multi-threading computers, and (maybe) exposing the debug API without any issues.
 - Improved support for yield prevention.

However, this is nowhere near as used as LuaJ, so there may be some unknown issues. There is also no Lua 5.2/5.3 version, which may disrupt your future plans.

## Synopsis changes in this PR
 - Replace `LuaJMachine` with `CobaltMachine`
 - Remove sandboxing of string metatable and environment.
 - Change the timeout system to run on instructions instead of API calls. One issue with the original system was it didn't trigger in simple infinite loops such as `while true do end`. With the new system, you will get an error mid iteration. It should also fire errors when in very long pattern matches.

## Potential improvements
 - Allow running computers across multiple threads. CCTweaks includes a system for multi-threading computers, but it requires more thorough testing - there are still issues with deadlocks involving peripheral methods.
 - Inject the `debug` API into the environment. The `debug` API is immensely useful, so would be nice to have. However, there may be some way I am unaware of to abuse it - even if you cannot leak into other computers any more.

## Potential issues
 - I have tried to strike a balance between sticking as close to normal Lua as possible, whilst still maintaining comparability with LuaJ. I'm not aware of any issues Cobalt has caused (it is used on several CC servers without a problem), but there may be some.


apemanzilla:
This would also make it easier to distribute CC, and make it easier for other developers to compile against CC. Cobalt could be added directly to the `build.grade` rather than requiring a set of shell scripts, which would make it easier for everyone.

Lignum:
Since this is quite a monster of a PR, I think it's worth mentioning that Cobalt has been tested in battle quite extensively before, so merging immature code should not be a concern here. Notably, [SwitchCraft](http://www.computercraft.info/forums2/index.php?/topic/26974-1102-switchcraft-survival-krist-cc180pr0-griefprevention/), when it was still active, had this fork in use via CCTweaks, which allowed it to stabilise significantly.

viluon:
All I can say is :tada: :+1: :sparkles: :clap: :package: :shipit: :ok_hand: :wine_glass: :repeat:, please merge.

SquidDev:
@jamierocks I've deliberately avoided that in order to reduce noise and make it clear what changes this PR makes. It should be trivial to delete the LuaJ sources at a later date.

jamierocks:
@SquidDev I'd still say remove it tbh. Either way at least remove libs/luaj.*.jar

simon816:
(I know this is not the best place to mention it but) have you considered looking into merging your fixes into LuaJ itself? would befit more than just the CC community.

Restioson:
+1. How much code will this break?

SquidDev:
> +1. How much code will this break?

If I've done my job correctly, none. Several servers use Cobalt already, and haven't experienced any issues. It is a fork of LuaJ, and so shares most of its functionality.

Restioson:
Nice! That is very good. Will any *java* code break, or are the changes easily fixable?

SquidDev:
No code using the public API should break (`IPeripheral`, `ILuaObject`, etc...). To be accessing the Lua VM you'd have to be using all sorts of reflection trickery - [Dan has said](https://github.com/dan200/ComputerCraft/pull/158#issuecomment-298379933) that he isn't fussed about that.

Restioson:
That's cool. +1, definitely merge this

SquidDev:
@dan200 I realise there are other PRs out there which are objectively more important than this one, but I was wondering if you have any thoughts on this PR? I realise this is a massive change so isn't going to be an instant merge (or a merge at all for that matter), but I would be interested to know your thoughts.

apemanzilla:
Not trying to rush things, but it would be really nice to see some feedback on this @dan200. It would be a great change and fix a lot of headache with LuaJ, and open up a lot of new possibilities for features.

JustPingo:
If you're planning to change the Lua engine, why not use LuaJIT directly? It's extremely fast and supports most architectures.
I understand it's outside of the JVM and so would require additional binding work, but that's something that could be worth it.

SquidDev:
@JustPingo Cobalt is pretty much LuaJ with some bug fixes and minor improvements, meaning it isn't as dramatic a change as swapping to an alternative Lua runtime (such as LuaJIT or PUC Lua).

On the subject of LuaJIT itself, I can see a couple of potential issues:

 - In tight loops (such as `while true do end`), debug hooks are not run. You'd probably have to disable the JIT to prevent this (though the interpreter's performance is still incredible).

 - LuaJ's semantics don't entirely match up with PUC Lua/LuaJIT's. This is entirely LuaJ's fault, but many CC programs rely on its "broken" functionality. I don't know how much this matters, but worth bearing in mind.

 - `ILuaAccess.yield` and friends rely on Lua coroutines being backed by Java threads. This doesn't quite work with the coroutine implementation of normal Lua. It is possible to write an adaptor (which is what I've done for [Rembulan][1]), but the performance isn't great.

   I'd really like to rewrite how yields are implemented, but that would have to wait for a new Minecraft version to avoid breaking binary compatibility - I've got a mostly-working implementation of "true" coroutines for Cobalt but, as mentioned previously, it won't play nice with CC's current implementation.


[1]: https://github.com/SquidDev-CC/CCTweaks-Runtimes/blob/master/src/main/java/org/squiddev/cctweaks/runtimes/rembulan/RembulanMachine.java#L261

Wojbie:
What kind of broken functionality are we dependent on? If its a small thing it might make sense to drop it?

SquidDev:
@Wojbie The thing which first springs to mind is that `error` and stack levels function a little differently. A common pattern in CC is to use `pcall(error, "", i)` to get a traceback - this doesn't work in PUC Lua. Though if #216 gets implemented this isn't really a problem any more.

One thing worth noting is that using a Java implementation of Lua means we can execute bytecode without a security risk - allowing for programs like [JVML-JIT](http://www.computercraft.info/forums2/index.php?/topic/20601-) and [LuaLua](http://www.computercraft.info/forums2/index.php?/topic/15157-). Sure, this is a small price to pay, but it would be a shame to lose these programs.

JustPingo:
@SquidDev 

- Is this relevant? I mean, that's not a major feature nor an irreplaceable one. And as you say, the interpreter still is astonishing compared to LuaJ.
- Now I guess that's a matter of personal philosophy, but I feel like relying on inaccurate behaviours is pretty bad globally, and thus dropping compatibility for minor stuff seems reasonable to me. Someone new learning Lua through the doc could be confused by those problems, and an experienced Lua programmer might also be confused about some of those bugs they would encounter.
- I don't understand what you mean. Aren't native Lua coroutines doing everything that is needed?

SquidDev:
@JustPingo 

> I don't understand what you mean. Aren't native Lua coroutines doing everything that is needed?

Yes. It's more about how ComputerCraft currently interacts with coroutines and expects them to behave. Methods like `ILuaAccess.yield` block the current thread until the coroutine is resumed again. As "true" coroutines don't work the same way, you have to call each CC method in a separate thread, translating the blocking `ILuaAccess.yield` calls into the appropriate coroutine methods. It's possible, just a bit of a faff and rather inefficient.

I agree that the other two points aren't major, but I do still think they are worth bearing in mind. They aren't deterring factors, but should be considered if we do chose to use LuaJIT.

The broken behaviour [relating to `pcall`](https://github.com/dan200/ComputerCraft/pull/163#issuecomment-304537229) is actually required in order to replicate missing functionality, so you can't blame people for depending on it - though as mentioned it won't be needed if `debug.traceback` is added so I don't think it is worth worrying about.

I also maintain that using LuaJIT is a totally separate issue - this is just minor modifications to LuaJ, LuaJIT would require a rewrite of much of ComputerCraft's internals.

MoreRobustThanYou:
Please merge

hugeblank:
Hate me for reviving this, but why is this stagnant...? I would've assumed this'd be the first PR to get into CC, but it's still sitting here open. Like is there some hitch with this? It doesn't even seem like a dual-edge sword, like most PR's that stay open for as long as this one are...

SquidDev:
I've just finished doing a tiny bit of work on Cobalt with some optimising [and bug fixing](http://www.computercraft.info/forums2/index.php?/topic/29053-vm-error-array-index-out-of-bounds/). As I'm preparing for a release, I'm wondering whether it would be better to reference Cobalt as a git submodule instead of a maven artifact.

If Cobalt were a submodule, one wouldn't have to depend on an external maven server. Whilst Bintray provides one for free, it may be something we want to avoid. Submodules also simplify updating, as one doesn't need to bother with the whole "bump version, upload, update dependents" faff.

However, there is a downside of this: git submodules. These introduce all sorts of additional complications with cloning and pulling which we may want to avoid.

I'd really appreciate other people's thoughts on the matter - I'm not sure what the best course of action here is.

Vexatos:
Most IDEs nowadays handle submodules very well, including auto-updating them on pull and pushing changes to them (if changes are done).

apemanzilla:
You could just copy the source to the CC repo directly, rather than a submodule, akin to how LuaJ is currently set up, but using a submodule or maven artifact would be much better.

viluon:
I don't really see a problem with making Cobalt a submodule. How does it complicate pulling and cloning? Sorry if it's obvious to others, I don't frequently work with submodules myself. I presume that you only need to use a specific command when updating them, however.

SquidDev:
@viluon There's lots of small things, like having to use `git clone --recursive` instead of `git clone`. I believe there are also complications involved with pulling also updating the submodule when you don't want it to. None of these are *major* issues, but they complicate an already confusing system.

jamierocks:
Frankly, I see little benefit to using a submodule. ComputerCraft, will always want to be using release code - little bugs, more accountability (I see people update submodules all the time carelessly).

viluon:
@jamierocks the submodule can track a release branch, and frankly, tracking `master` isn't a problem with [the right branching model](https://docs.gitlab.com/ee/workflow/gitlab_flow.html),

@SquidDev can't you clone the submodule on build?

jamierocks:
@viluon clearly, I need to stop posting late at night - I did not articulate my point very well at all.

I was trying to get across that it would put more emphasis on ComputerCraft devs to know what is going on with Cobalt development - which would possibly result in Cobalt not getting updated. (this could be because not sure which releases/commits are going to break CC - appropriately versioned dep releases doesn't have this problem, although there quick fixes can't be merged quickly)

My other point, was that tracking a specific version is just a much more painful method than using a Maven dependency (having to recursively clone, etc) - the two get the exact same result, the easiest option IMO would be the desired one.

I totally see your point with using a branching model (that has a 'stable' branch), however in that case, Cobalt would have to make alterations - for something which can already be accomplished (through Maven dep).

Basically submodules are only useful if you know exactly what development is going on upstream. If you are tracking specific releases, you are just making life harder - having to combine the build scripts of Cobalt into CC, etc.

Don't get me wrong - I love submodules, however from my use of them, I have found that Maven dependencies (when they can be used) are far more appropriate - especially when you'd have to combine two builds. (although at least both use Gradle here)

PR #167, opened 2017-05-01
Impose additional limitations on disk and computer labels

### LINKED ISSUE #147: Disk labels have no length limit
One can set the disk label length to be a string as long as you like. This can result in crashing the client or world corruption as it cannot fit within an NBT Tag:

 - Create a new computer and place a disk drive + disk next to it.
 - Set the disk's label to `("x"):rep(1e7)`. 
 - My game hangs on a client world. Attempting to close Minecraft results in `encoded string to long: 10000000 bytes` when trying to serialise NBT.

It is also possible to use `\0` within labels which causes issues with Sponge. Whilst I realise Sponge is not officially supported by ComputerCraft it would be nice if labels were limited to the printable, ASCII range.

SquidDev:
- Disk labels are limited at 32 characters (like computer lables are)
 - Labels can now only include characters within the printable range (` ` to `~`).

I wasn't sure where to put the normalisation function so created a new class for it. Please say if there is a better place for it.

Closes #147

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/OSAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/OSAPI.java
@@ -8,6 +8,7 @@ package dan200.computercraft.core.apis;
 
 import dan200.computercraft.api.lua.ILuaContext;
 import dan200.computercraft.api.lua.LuaException;
+import dan200.computercraft.shared.util.StringUtil;
 
 import java.util.Arrays;
 import java.util.HashMap;
@@ -266,11 +267,7 @@ public class OSAPI implements ILuaAPI
                     {
                         throw new LuaException( "Expected string or nil" );
                     }
-                    label = (String)args[0];
-                    if( label.length() > 32 )
-                    {
-                        label = label.substring( 0, 32 );
-                    }
+                    label = StringUtil.normaliseLabel( (String) args[0] );
                 }
                 m_apiEnvironment.setLabel( label );
                 return null;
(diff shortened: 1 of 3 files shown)


Lignum:
I'm not great at deciphering regexes, but if, as you said in your issue, you restricted labels to the printable _ASCII_ range, then I have some doubts. What if I want äöü in my labels? If it doesn't already cover it, maybe [something like this](http://stackoverflow.com/a/418560) will do the trick.

SquidDev:
@Lignum Yeah, I'd forgotten about non-ASCII characters. Sorry. I'll have to check this - I don't _think_ CC characters entirely match up with their Java script representation.

dan200:
CC maps the 256 values in a lua string byte to the first 256 characters of unicode: https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Codepage_layout (with some custom glyphs if you try to render any of the control characters a screen). Therefore, it should be fine to just check if the character is not in the 0-31 or 128-160 range. My preference would be replacing the control characters with question marks in the strings instead of stripping them.

SquidDev:
Ahhh, I was aware it used IEC-8859-1, just didn't realise that mapped directly onto unicode. Good to know.

I've changed it to use `?` instead. I'm not too sure about allowing 128 (not defined) and 173 (soft-hyphen) as Minecraft renders them as unknown characters. I'm not too sure what to do about empty labels - currently it allows it, but I feel it should throw an error or clear the label entirely.

Wojbie:
Clearing the label entirely could pose the risk of loosing the turtle contents. setComputerLabel should only clear on nil i believe.

dan200:
127 is the DELETE control character, so don't allow that one. How does it look when MC renders soft hyphen? I'm also ok with setLabel("") acting the same as setLabel(nil)

SquidDev:
java_2017-05-02_13-48-10

dan200:
Ah that's not too pretty. Stick a question mark in there!

PR #168, opened 2017-05-01
Various model improvements

SquidDev:
This ensures the main modem texture is facing towards the screen. Sorry, I should have caught this in my earlier PR.

Before:
java_2017-05-01_22-38-00

After:
java_2017-05-01_22-36-24

We also use `item/generated` instead of `builtin/generated` for items, meaning they look better when in entity form and 3rd person.


dan200:
Does this remove the need for #164?

Issue #144, opened 2016-10-05
File Handle Limit


inklit:
File handles can be abused to cause corruption in the world, player data, and computers by simply opening too many of them.

The operating system will be unable to create new file descriptors which will cause all sorts of issues system wide.

See https://github.com/SquidDev-CC/CCTweaks/issues/118 for more information.

Can this be added into ComputerCraft?

PR #169, opened 2017-05-01
Configurable file descriptor limit

Lignum:
This adds a configurable file handle limit, defaulting to 128, to prevent players from exhausting the server's file descriptors. This is in response to issue #144.

The limit is always active and can be configured with `I:maximumFilesOpen`. To "disable" it, just set the limit to a really high value. Negative or zero values will simply prevent files from being read and written, making all computers useless.

Once the limit is reached, `fs.open` will return `nil`.

### DIFF
--- a/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
+++ b/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
@@ -651,12 +651,20 @@ public class FileSystem
         }
     }
 
-    private synchronized IMountedFile openFile( IMountedFile file ) throws FileSystemException
+    private synchronized IMountedFile openFile( IMountedFile file, Closeable handle ) throws FileSystemException
     {
         synchronized( m_openFiles )
         {
             if( m_openFiles.size() >= ComputerCraft.maximumFilesOpen )
             {
+                if( handle != null )
+                {
+                    try {
+                        handle.close();
+                    } catch ( IOException ignored ) {
+                        // shrug
+                    }
+                }
                 throw new FileSystemException("Too many files already open");
             }
 
@@ -721,7 +729,7 @@ public class FileSystem
                     throw new UnsupportedOperationException();
                 }
             };
-            return (IMountedFileNormal) openFile( file );
+            return (IMountedFileNormal) openFile( file, reader );
         }
         return null;
     }
@@ -773,7 +781,7 @@ public class FileSystem
                     writer.flush();
                 }
             };
-            return (IMountedFileNormal) openFile( file );
+            return (IMountedFileNormal) openFile( file, writer );
         }
         return null;
     }
@@ -811,7 +819,7 @@ public class FileSystem
                     throw new UnsupportedOperationException();
                 }
             };
-            return (IMountedFileBinary) openFile( file );
+            return (IMountedFileBinary) openFile( file, stream );
         }
         return null;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
```diff
@@ -649,6 +650,33 @@ private synchronized void copyRecursive( String sourcePath, MountWrapper sourceM
...
         }
     }
+
+    private synchronized IMountedFile openFile( IMountedFile file ) throws FileSystemException
+    {
+        synchronized( m_openFiles )
+        {
+            if( m_openFiles.size() >= ComputerCraft.maximumFilesOpen )
+            {
+                throw new FileSystemException("Too many files already open");
```
SquidDev: You're not closing the file handle here.

SquidDev:
Heh, this was next on my todo list - yah beat me to it 😛. 

One issue with this implementation is that `FileSystemException` is caught by `fs.open`. It just returns `null` instead, swallowing the error. You'd probably have to create a separate exception, or throw a `LuaException` instead. Neither solution seems especially elegant.

Lignum:
I think that's intentional? Since there's a "No such file" exception being thrown when you open a file that doesn't exist, but `fs.open` also just returns `nil`.

src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
Lignum: Oops, will fix in a second.

src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
```diff
@@ -697,11 +729,7 @@ public void flush() throws IOException
                     throw new UnsupportedOperationException();
                 }
             };
-            synchronized( m_openFiles )
-            {
-                m_openFiles.add( file );
-            }
-            return file;
+            return (IMountedFileNormal) openFile( file, reader );
```
SquidDev: Do you need the cast here? Can you not just do:

```java
openFile( file, reader );
return file;
```

Lignum: I can, but I wanted a one-liner, which is why `openFile` returns the handle in the first place. So unless that cast is bad for some reason, it's down to preference, I guess.

SquidDev: I generally don't like casts unless you need them. I guess you could make `openFile` generic:

```java
private synchronized <T extends IMountedFile> T openFile( T file, Closeable handle ) throws FileSystemException
```

Though I'm not sure that's any nicer 😄.

JLLeitschuh:
```
private synchronized <T extends IMountedFile> T openFile( T file, Closeable handle ) throws FileSystemException
```
I think that would accurately capture what is trying to be done here.
It actually seems like a better solution.

Lignum: I've gone with that solution now.

dan200:
This limit is per-computer, not per-server. Is that the intent?

Lignum:
Yes, that's the intent.

Lignum:
Cheers! :tada:

PR #172, opened 2017-05-02
Pocket computer upgrades

SquidDev:
## Internals
This PR adds a system for registering upgrades to pocket computers. The API mirrors that of turtles, with a `IPocketUpgrade` and `IPocketAccess` interfaces. 

[`IPocketUpgrade`][1] is composed of:
 - Some basic getters (upgrade id, adjective, crafting item)
 - Peripheral factory
 - Upgrade update
 - Right click handler (return true to not open the GUI).

This is registered through the `ComputerCraftAPI.registerPocketUpgrade(IPocketUpgrade)` method, which just stores it in a `Map<String, IPocketUpgrade>` in `ComputerCraft`. We will use this map to loop up upgrades by id and crafting item.

[`IPocketAccess`][2] exposes several methods to interact with the given computer:
 - Getter/setter for light colour
 - Current entity getter
 - Getter for NBT storage (like `ITurtleAccess.getUpgradeNBTData`). This is stored on the appropriate item stack.

`IPocketAccess` is implemented as a subclass of `ServerComputer`. It stores a reference to the owning `ItemStack` and holding entity, allowing it to get and set upgrades and upgrade data. Whilst this is rather ugly, it's the only solution I can come up with. Every time the item is ticked we will load the computer/pocket access (like normal), and tick the upgrade. 

## Brief summary of changes
 - Add API to register pocket upgrades.
 - Add ender pocket computer upgrade.
 - Add `.equip` and `.unequip` methods to the `pocket` API, in order to add and remove upgrades.
 - Allow pocket computer light to be any one of the 16 terminal colours.

## Potential issues/problems
 - As mentioned above, implementing `IPocketAccess` on the `ServerComputer` is rather ugly.

[1]: https://github.com/SquidDev-CC/ComputerCraft/blob/feature/pocket-upgrades/src/main/java/dan200/computercraft/api/pocket/IPocketUpgrade.java
[2]: https://github.com/SquidDev-CC/ComputerCraft/blob/feature/pocket-upgrades/src/main/java/dan200/computercraft/api/pocket/IPocketAccess.java


Wojbie:
Soo correct me if i am reading it right.

Would any upgrade have access to getModemLight and setModemLight? That would make for some intersting peripheral possibles.
Also you can swap upgrades by crafting them (and receive the old one back).
And the - onRightClick() - Right click handler (return true to not open the GUI). - I am kinda out of ideas what you would use that for. Could you enlighten me?

SquidDev:
> Would any upgrade have access to getModemLight and setModemLight? That would make for some intersting peripheral possibles.

Yep. The methods could potentially just be called `getLight` and `setLight` as they aren't specific for modems.

> Also can you swap upgrades by crafting them (and receive the old one back).

I'm in two minds about this one: it's more convenient than opening up the computer and running `pocket.equip()`. However, it does get rather messy: the crafting recipe would have to return two stacks: the removed upgrade and the modified pocket computer. I guess you'd probably leave the computer in the crafting grid and return the upgrade. I've done this before for other things, but it's not too elegant.

> And the - onRightClick() - Right click handler (return true to not open the GUI). - I am kinda out of ideas what you would use that for. Could you enlighten me?

This is there to allow players to interact with the upgrade without opening the GUI. For instance, you could have a "block click detector" which, when shift-right-clicking a block, will fire an event with the block's coordinates rather than opening the GUI\*. You could then use this upgrade to select a region of the world for your turtle to dig out.

\* You'd probably want to think about this a little more, as you don't want to be "leaking" the in-world coordinates. There is also only space for one upgrade so you couldn't then send these coords somewhere else. Allowing multiple upgrades would also be worth considering at some point.

Wojbie:
> I'm in two minds about this one: it's more convenient than opening up the computer and running pocket.equip(). However, it does get rather messy: the crafting recipe would have to return two stacks: the removed upgrade and the modified pocket computer. I guess you'd probably leave the computer in the crafting grid and return the upgrade. I've done this before for other things, but it's not too elegant.

Well turtles have equip and unequip cause they have internal inventory. In case of pocket it would grab stuff from player inventory? Soo like you type command and magical robot arm plugs it out your pocket and onto Pocket?

> This is there to allow players to interact with the upgrade without opening the GUI. For instance, you could have a "block click detector" which, when shift-right-clicking a block, will fire an event with the block's coordinates rather than opening the GUI. You could then use this upgrade to select a region of the world for your turtle to dig out.

This kinda feels like it needs some more polish. Cause as it is right now it feels like leaking coords into game. What if it returned a vector of 3 numbers that are "offset" from player to the clicked block. That sounds like something that modern day phone stuff can do. But you would still need a GPS coords to make it into a coordinate. But the only slot is blocked by "block click detector".  And that brings me to 2nd part of this post.

I also feel like its a good moment to think about the peripherals for pocket. And talk about where the system can be taken.
First - let me share my dream for this. 2 types of peripherals - communication and tool:

1. Communication - "back" - would be place where communication modems can be put. Be it wireless,ender or any other communication ones designed by some modder.
2. Tool - "bottom?" - Any other peripheral that adds functionality to the pocket like said "block click detector" or "flip out keyboard" (Shift-right-click anywhere while its on and you can send keyboard event to pocket with screen off.) or "Bell" or "Scanner for magical essence"

Reasons for this design? Because pockets are made to be connecting to the computers. And having only one peripheral slot stop people from making all those crazy peripherals they dream off cause it would not be able to communicate back (Or like in case of that chunk-loading peripheral for turtles they made it work as wireless modem too).

Sorry for long post but that is what i am dreaming off and last part of SquidDev comment triggered all my daydreams.

SquidDev:
>  Soo like you type command and magical robot arm plugs it out your pocket and onto Pocket?

When you put it like that, I guess it doesn't make much sense. 😛 I generally have have the attitude of "make everything automatable", but that shouldn't come at the cost of being nonsensical.

> This kinda feels like it needs some more polish. Cause as it is right now it feels like leaking coords into game. 

Definitely. I wasn't suggesting this specific item as something that could be added to ComputerCraft, just a potential use-case. Right-clicking is the only form of interaction we have for pocket computers, so I feel we should make the most of it. However, if [pocket computer map-like rendering](http://i.imgur.com/l3BKvFi.gifv) gets added, right-click actions will need to be reconsidered.

> First - let me share my dream for this. 2 types of peripherals - communication and tool

I do like this, though I'm not too sure whether distinguishing between the two is necessarily a good idea. While turtles distinguish between peripherals and tools, you can use either one on either sides, allowing turtles to have a sword and pick, two modems, or any other combination. Limiting to one of each doesn't seem the correct choice. You then need to consider what to do about the modem light when you have two modems. There is room on the texture to have multiple lights.

Another possibility would be to allow upgrades to take the crafting item's NBT into account. This could allow for "container" items which hold multiple upgrades. Though that doesn't really make much sense in ComputerCraft "lore".

Another thing to consider is swapping upgrades through crafting. I'm not sure of a nice way to handle multiple upgrades at once.

Thank you very much for the feedback @Wojbie! I'd really appreciate hearing what other's have to say, especially other peripheral mod devs.

apemanzilla:
Maybe instead of having functions to un/equip upgrades, shift right clicking a pocket computer would remove all upgrades and then you can craft them together to add them back?

Wojbie:
> Maybe instead of having functions to un/equip upgrades, shift right clicking a pocket computer would remove all upgrades and then you can craft them together to add them back?

But then you loose on possibility of right click getting used for peripherals.

dan200:
I like this! Let me know when you think the code is clean enough for a proper review.
Only thing I think needs a bit more thought is where "pocket.equip()" API finds the item. Currently it just scans the inventory from the start.. should it take a slot number instead? or scan starting at the computer location? or look in the offhand slot? it would be good for the player to have some intuition about what thing will get equipped.

SquidDev:
> Only thing I think needs a bit more thought is where "pocket.equip()" API finds the item. 

It currently starts from the current slot and loops its way round the inventory. Starting from the computer slot would be more logical - I think this was what I originally intended and then mucked up. Not quite sure what to do about the offhand slot.

Obviously, whatever behaviour I decide on, I'll add some documentation to `help/pocket`.

Wojbie:
Just a suggestion. If you go with .equip() way how about calling it .equipBack()

That way its consistent with turtles and if there ever are more spots on pocket you just add a side.

dan200:
Ooh I didn't spot that squiddev. If it uses the players selected slot then it's fine, and equivalent to turtle.equipLeft/Right() using the turtle.select slot

SquidDev:
@dan200 I'm wondering what your thoughts are on some of the above comments, namely [this one](https://github.com/dan200/ComputerCraft/pull/172#issuecomment-298605228)? I feel crafting support for swapping out pocket upgrades would make sense, though then we might also want to add that for turtles too.

dan200:
> Yep. The methods could potentially just be called getLight and setLight as they aren't specific for modems.

Agreed. Could even let you pass in "Colour" enum if you felt like changing the item rendering a bit!

> Also can you swap upgrades by crafting them (and receive the old one back).

I would only do this if you could think of an elegant way to do it for turtles as well.

> This kinda feels like it needs some more polish. Cause as it is right now it feels like leaking coords into game.

I have no problem giving a peripheral all the info it wants about the Minecraft world! It's up to it what it actually passes into lua.

> First - let me share my dream for this. 2 types of peripherals - communication and tool

Seems a bit unnecessary, not sure what a "tool" could do that a peripheral couldn't, there's no equivalent to the turtle.dig() on pocket computers.

> And the - onRightClick() - Right click handler (return true to not open the GUI). - I am kinda out of ideas what you would use that for. Could you enlighten me?

I don't mind the existence of the right click handler, as long as it's only invoked when the computer is on, and you can hold shift to bypas it.

SquidDev:
> Agreed. Could even let you pass in "Colour" enum if you felt like changing the item rendering a bit!

The `Colour` enum is currently outside the API classes. Are you OK me moving it?

> I don't mind the existence of the right click handler, as long as it's only invoked when the computer is on, and you can hold shift to bypas it.

The general idea was that the upgrade handled how it was bypassed: for some peripherals it might make sense to use shift-click, others it might make sense to normal right-click. However, I guess it being consistent would be better.

dan200:
I guess moving Colour to API is fine, only if you feel like adding the feature though!.

Wojbie:
> Seems a bit unnecessary, not sure what a "tool" could do that a peripheral couldn't, there's no equivalent to the turtle.dig() on pocket computers. 

Ok going back to basic. @dan200 what is your opinion on 2 peripherals on pocket. Any combination player wants?

dan200:
Actually: changed my mind on that last comment. Just take an integer in the range 0-15 and call Colour.fromInt() on our side. We don't need to expose the whole class and the colour constants etc.

dan200:
If you supported two upgrade slots, do the same as turtle where which side you craft it on determines where it goes. We haven't enforced tool=left, peripheral=right on turtles for quite a few versions now.

One thing though: the current peripheral is "back", and moving it would break code? How do we resolve that? Only clean way I can think is supporting *3* upgrades, with the back one crafted by placing above the computer in the grid.

Wojbie:
Well second one could be on "bottom" side? And that way you craft it by placing it below the pocket. Would need a second led on pocket itself through.

dan200:
I would stick to one tbh.

SquidDev:
So I've added support for various colours - this was just me testing it out, not an actual feature. One thing worth noting is red is significantly lighter than the original modem colour (red is `0xCC4C4C`, modem was `0xBA0000`).

dan200:
It looks fine :)

SquidDev:
I've done a little bit of a clean up. I'm not entirely happy with it (handling computers on items always ends up a little ugly), but I feel it is still an improvement.

I wonder if it would be worth adding a pocket computer equivalent of the `equip` program. I guess it wouldn't do much more than run `pocket.equip()`, so is rather less useful.

SquidDev:
Just rebased this branch back onto master to remove merge conflicts. I'd really appreciate any feedback from other mod devs on the API - anything they feel is missing or could be improved. A bit of me would quite like to change the "modem" light to allow full RGB colours - there isn't any technical reason we couldn't, but it may be better to keep it at the base 16 colours.

dan200:
Just gave this a new look over and didn't see anything I didn't like, so merged in in. Time to test it out!

PR #174, opened 2017-05-02
Send termination signals unless the computer is off

SquidDev:
If a shutdown has been queued, then the abort message is not set. This allows for programs to run for a significantly longer period of time, bypassing the "too long without yielding" error.

## Example
Run the following code, and press Ctrl+S after a couple of seconds.

```lua
local i = 1
while true do
    print(i, os.clock())
    i = i + 1
end
```

Without this patch, the computer will run forever. This can be observed by opening VisualVM and looking at the thread view: there will be a coroutine which is still running.

Note that using `os.shutdown()` does not reproduce this issue. I believe there is a threading issue somewhere, this is something to investigate further.

### DIFF
--- a/src/main/java/dan200/computercraft/core/computer/Computer.java
+++ b/src/main/java/dan200/computercraft/core/computer/Computer.java
@@ -285,7 +285,7 @@ public class Computer
     {
         synchronized( this )
         {
-            if( m_state == State.Running )
+            if( m_state != State.Off && m_machine != null )
             {
                 if( hard ) 
                 {


JLLeitschuh:
Something to be aware of, I noticed that there are several places where the Interrupted Exception isn't handled correctly.

There are several places where you see
```java
} catch (InterruptedException e) {
  // do nothing
}
```
or:
```java
} catch (InterruptedException e) {
  continue;
}
```

What the code really should be is this:
```java
} catch (InterruptedException e) {
  Thread.currentThread().interrupt();
  continue;
}
```
This resets the interrupt flag so that it will actually die.
http://www.yegor256.com/2015/10/20/interrupted-exception.html

You may also want to check for loops in threads that are `while(true)`
They may want to be changed to `while (Thread.currentThread().isInterrupted())`

DON'T do: `while (!Thread.interrupted())` as it will clear the interrupt flag and will cause problems.

PR #176, opened 2017-05-02
German language support

LuaXLab:
What it does is simple, isn't it?

This PR simply add's german language (de-DE) support to computercraft.

I hope that a pull is ok for that


Lignum:
[Dan has stated before](https://github.com/dan200/ComputerCraft/pull/5) that language support should be provided by resource packs.

LuaXLab:
Oh, dont seen that

PR #177, opened 2017-05-02
Make monitors emit light

SquidDev:
[As mentioned in this suggestion](http://www.computercraft.info/forums2/index.php?/topic/15568-) Yes, it is rather silly. However, I do think it looks rather cool. The values monitors give off is very much up for debate - it is set rather high currently.

### DIFF
--- a/src/main/java/dan200/computercraft/ComputerCraft.java
+++ b/src/main/java/dan200/computercraft/ComputerCraft.java
@@ -128,6 +128,9 @@ public class ComputerCraft
     public static int computerSpaceLimit = 1000 * 1000;
     public static int floppySpaceLimit = 125 * 1000;
 
+    public static int monitorLight = 7;
+    public static int advancedMonitorLight = 10;
+
     // Blocks and Items
     public static class Blocks
     {
@@ -184,6 +187,9 @@ public class ComputerCraft
 
         public static Property computerSpaceLimit;
         public static Property floppySpaceLimit;
+
+        public static Property monitorLight;
+        public static Property advancedMonitorLight;
 
     }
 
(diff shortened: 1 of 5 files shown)


apemanzilla:
Would be really cool if the light level was based on the contents of the monitor, i.e. full white would emit 15 light and full black 0 light.

InsomniaKitten:
I'd suggest only producing light when the monitor is active/has screen contents

SquidDev:
My one worry with dynamically adjusting the light level (either based off of it having contents or it's whiteness) is that Minecraft light computation is expensive - monitors can cause performance problems as is, adding light level changes would make it worse.

InsomniaKitten:
Having it only produce light when active wouldnt cause significant issues, it'd only be as bad as turning on a redstone lamp when you use the computer :P

Lignum:
> My one worry with dynamically adjusting the light level (either based off of it having contents or it's whiteness) is that Minecraft light computation is expensive 

As far as I know, it's not _computing_ the lighting that's expensive. Since Minecraft doesn't use any shaders (yet?), all lighting is uploaded to the graphics adapter as part of the vertex data. This means that when lighting changes, the entire chunk needs to be refreshed, which shouldn't be something that most modern graphics cards can't handle. Especially considering that this also happens when you break/place a block. So if anything, you could only do that when the graphics are set to fancy? Or maybe make it a separate setting?

> monitors can cause performance problems as is, adding light level changes would make it worse.

Their rendering is generally quite wonky... they shouldn't lag (and also not act weird when shaders are in use), so maybe we should generally look over what can be done about that.

SquidDev:
> As far as I know, it's not computing the lighting that's expensive.

I will confess I haven't looked into the lighting code at all, I'm just going off word of mouth here. Lighting isn't done via a shader though: it is calculated in Java and stored the block. This is done on both the client and server, as servers need access to light levels for mob spawning. From what I can tell from the code, it'll loop in a sphere around each light source, recalculating it's light. This isn't too bad, but if done multiple times a tick, could prove problematic.

> Their rendering is generally quite wonky... they shouldn't lag (and also not act weird when shaders are in use), so maybe we should generally look over what can be done about that.

In normal use they don't, but they can be problematic when sending lots of terminal updates in a short space of time. I'm not sure what the issue is here, but it is worth investigating.

Lignum:
> Lighting isn't done via a shader though: it is calculated in Java and stored the block.

Yep, meaning it has to be reupload the block's vertex data as part of a chunk reupload. A shader would prevent this, but that's not something we have to worry about.

 > There isn't really a measure of whether the terminal is on or not. I think the only way would be to determine whether the terminal is blank, or only have it on if it is wrapped as a peripheral. Neither is an especially satisfactory solution.

Perhaps you could check how many pixels are non-black and determine an appropriate light value from that? A monitor that's painted half white would have half of its maximum luminosity. If you're feeling adventurous, maybe you could make dark colours affect the value less than light colours, but I don't think people will notice that much.

JLLeitschuh:
I have no idea if this is even possible in Minecraft but should the back produce light? Shouldn't only the front face produce light?

SquidDev:
> I have no idea if this is even possible in Minecraft but should the back produce light? Shouldn't only the front face produce light?

Sadly that isn't possible, though it would be nice. If I lower the light levels a little, the fact that it goes in all directions shouldn't be as much of an issue.

dan200:
Make the intensity a config option and set the default value to something low and i'll merge this.

dan200:
(also document in changelog/whatsnew)

Vexatos:
How about they won't actually emit light but just the text renderer would ignore lighting, i.e. glow in the dark? Would match real screens a bit better.

SquidDev:
I've added config options, but I think @Vexatos's suggestion is worth considering as an alternative to this.

dan200:
I investigated additive blending for monitor text back in the day and couldn't get it to work. That was a lot of versions of the Minecraft rendered ago though! Feel free to give it a crack.

SquidDev:
So this is what it looks like with colour bleeding and without emitting any light. I don't know what people prefer. It would be totally possible to have both I guess.

Wojbie:
It does look good. Personally i would prefer this to glowing monitors.

PR #179, opened 2017-05-02
Monitor and terminal rendering improvements

Lignum:
I've made some changes to how monitors are rendered, and by changing how text is rendered, terminal rendering got some of these improvements as well.

In particular:
- I removed all usages of `GL_QUADS` and replaced them with something more appropriate. This primitive type has been deprecated because OpenGL doesn't know what quads are. Thus, all primitives rendered with `GL_QUADS` are triangulated on the CPU before drawing, which is just a waste of time.
- I replaced any direct OpenGL code I could find with Minecraft's equivalents. This includes:
  - Replacing calls like `glTranslatef` with `GlStateManager.translate`
  - Substituting the display list conundrum with calls to `Tessellator`, which allows Minecraft to use VBOs here!
- Microoptimisations here and there

These changes only apply to monitor-related code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ deploy
 *.ipr
 *.iws
 *.iml
-.idea/
+.idea
 .gradle
 luaj-2.0.3/lib
 luaj-2.0.3/*.jar
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ deploy
 *.ipr
 *.iws
 *.iml
+.idea/
 .gradle
 luaj-2.0.3/lib
 luaj-2.0.3/*.jar
--- a/src/main/java/dan200/computercraft/client/gui/FixedWidthFontRenderer.java
+++ b/src/main/java/dan200/computercraft/client/gui/FixedWidthFontRenderer.java
@@ -57,10 +57,16 @@ public class FixedWidthFontRenderer
 
     public void drawStringBackgroundPart( int x, int y, TextBuffer backgroundColour, double leftMarginSize, double rightMarginSize, boolean greyScale )
     {
-        // Draw the quads
         Tessellator tessellator = Tessellator.getInstance();
         VertexBuffer renderer = tessellator.getBuffer();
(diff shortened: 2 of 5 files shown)


KingofGamesYami:
Does this make rendering fast enough to handle [this](http://www.computercraft.info/forums2/index.php?/topic/27098-monitors-are-excessively-hard-on-fps/)?

apemanzilla:
@KingofGamesYami there's one way to find out!

Lignum:
I don't really expect this to be a huge performance gain, but yeah, give it a go.

dan200:
I'm not convinced this is a speedup for monitor renderer.. by losing the code which keeps track of whether the geometry needs to be regenerated, you're creating new geometry and sending it to the GPU every single frame. Vertex buffers are better than display lists, sure, but a display list that never changes is better than constantly refilling a vertex buffer.

Lignum:
@dan200 While writing this, I figured that Minecraft would do the clever things it does to chunk rendering to optimise this away secretly. Though @SquidDev informed me that that's not the case. So, I was planning to experiment with caching the VBO... and from looking into Minecraft's source, making our own `Tessellator` instance might not be as insane as it sounds. Though, I haven't actually tried it, meaning that could go wrong in a lot of ways.

If that doesn't work out, I'll just close this and make a separate PR which has the other improvements in it.

dan200:
Sounds good. The other change: I just removed the trailing backslash for consistency with the other folders in the file.

Lignum:
Oh, for some reason I thought you needed the slash for folders. Could've figured from the other ones. The more you know :)

apemanzilla:
Slashes aren't required for folders, but it's generally preferred as it makes it clear you're excluding a folder and not just a single file.

dan200:
I was just being pedantic about consistency. If you want to change them all to slashes, go ahead

Lignum:
Hm yes, this doesn't seem feasible. There would have to be multiple `Tessellator` instances for each cached VBO, and I'm already not confident in making our own tessellators, which means I'll make that other PR once I get around to it.

Issue #131, opened 2016-07-30
[1.80pr0] Ghost turtle peripheral when moving


negamartin:
When a turtle-moving function is called from a sub-coroutine, turtle_response events are generated as usual, but I noticed there are also some "peripheral" and "peripheral_detach" events being generated from the side the turtle is moving towards. This phantom peripheral is of type "turtle", and has all normal turtle methods, almost like the turtle can see itself as a peripheral.
However, calling the getID() method on this peripheral gives a new ID, different from any other, just like if a new computer had been created.

To recreate, use MC 1.9.4/MC 1.10.2 and CC1.80pr0. Yes, it works in both MC versions.
Then download and run the code on the 'sGfyYhbb' pastebin. ([This pastebin](http://pastebin.com/sGfyYhbb))

This code will call the turtle.up and turtle.down functions repeatedly while listening for events in another coroutine. Calling getID() before the peripheral is detached seems to be hard, so it's done immediately after the 'peripheral' event is received, and even then it works sometimes, so please be patient.

On CC1.79/MC1.8.9 this doesn't happen (only 'turtle_response' events are generated).

dan200:
Fixed. Thanks squiddev!

PR #180, opened 2017-05-03
Handle tile entity changes

boq:
Currently CC ignores updates from
```java
class Block {
//...
    public void onNeighborChange( IBlockAccess world, BlockPos pos, BlockPos neighbour );
//...
}
```

This method is called when tile entity changes (e.g. `TileEntity.markDirty`, `World.setTileEntity`). Supporting this callback fixes interactions with blocks that change TE without block update (at the cost of increased number of peripheral updates).

Main example would be vanilla furnace, that first changes block, which causes new TE to be created (and wrapped in peripheral), but then restores old one, which currently causes peripheral to keep reference to invalidated TE.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/common/BlockGeneric.java
+++ b/src/main/java/dan200/computercraft/shared/common/BlockGeneric.java
@@ -46,7 +46,7 @@ public abstract class BlockGeneric extends Block implements
     public final void dropBlockAsItemWithChance( World world, BlockPos pos, IBlockState state, float chance, int fortune )
     {
     }
-        
+
     @Override
     public final List<ItemStack> getDrops( IBlockAccess world, BlockPos pos, IBlockState state, int fortune )
     {
@@ -107,7 +107,7 @@ public abstract class BlockGeneric extends Block implements
     {
         Block.spawnAsEntity( world, pos, stack );
     }
-    
+
     @Override
     public final void breakBlock( World world, BlockPos pos, IBlockState newState )
     {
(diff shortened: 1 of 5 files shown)


SquidDev:
Ohhh, the furnace issue has always been a nuisance, so it would be nice to have it fixed. I've noticed a couple of issues however:

 - This reintroduces #131 - When moving the turtle you get a temporary turtle peripheral.
 - Testing this with Plethora you get three peripheral changes per furnace state change. It would be nice if we didn't get any.

Another potential solution would only be to update the peripherals at the end of the tick. I don't know if that would cause a whole different set of problems though - there is definitely potential for peripherals and the world to be one tick out of sync.

boq:
About second issue: you mean full updates or just attempts to update? Because [this](https://github.com/dan200/ComputerCraft/blob/master/src/main/java/dan200/computercraft/core/computer/Computer.java#L592) check is supposed to prevent full update if `IPeripheral.equals` is implemented correctly.

SquidDev:
> About second issue: you mean full updates or just attempts to update? Because this check is supposed to prevent full update if IPeripheral.equals is implemented correctly.

I mean full updates. Plethora does implement `IPeripheral.equals`, but it compares the TileEntity, which I believe temporarily changes. I'll double check though.

boq:
Ah, yes, it will get updates for both temporary and restored TEs. Though it's still better then operating on invalidated temporary TE.

About first issue, I think this may be solved by marking old turtle TE as 'moved' before new turtle block is placed (just few lines up) and then skipping neighbour updates on old TE. I will try to play with it.

boq:
@SquidDev pushed alternative solution for phantom turtles, seems to work with [your testcase](https://pastebin.com/sGfyYhbb)

Issue #138, opened 2016-08-22
GPS.locate can crash CC computer


lupus590:
Original post on forum: http://www.computercraft.info/forums2/index.php?/topic/27427-crash-any-computer-using-gpslocate/#entry258294

OP (Edited): [Code which caused crash](https://github.com/powerboat9/jam)
Gps.locate only verifies that the gps host is sending a table with 3 coords, but doesn't test to see if the coords are numbers. Sending three tables is enough to trip up the attempt to tostring the vector. 

While investigating this I also noticed that the `vector.new` doesn't verify that it has received numbers.
I've made an untested bug fix though: 

``` Lua
-- vector API
-- Make sure args are numbers
function new( x, y, z )
local v = {
  x = tonumber(x) or 0,
  y = tonumber(y) or 0,
  z = tonumber(z) or 0
}
setmetatable( v, vmetatable )
return v
end
```

``` Lua
-- GPS API
-- Make sure GPS host positions are not nil
function locate( _nTimeout, _bDebug )
    -- Let command computers use their magic fourth-wall-breaking special abilities
    if commands then
        return commands.getBlockPosition()
    end
-- Find a modem
local sModemSide = nil
for n,sSide in ipairs( rs.getSides() ) do
  if peripheral.getType( sSide ) == "modem" and peripheral.call( sSide, "isWireless" ) then
   sModemSide = sSide
   break
  end
end
if sModemSide == nil then
  if _bDebug then
   print( "No wireless modem attached" )
  end
  return nil
end

if _bDebug then
  print( "Finding position..." )
end

-- Open a channel
local modem = peripheral.wrap( sModemSide )
local bCloseChannel = false
if not modem.isOpen( os.getComputerID() ) then
  modem.open( os.getComputerID() )
  bCloseChannel = true
end

-- Send a ping to listening GPS hosts
modem.transmit( CHANNEL_GPS, os.getComputerID(), "PING" )

-- Wait for the responses
local tFixes = {}
local pos1, pos2 = nil, nil
sBadPositionError = "GPS replied with bad position, check your GPS hosts."
local timeout = os.startTimer( _nTimeout or 2 )
while true do
  local e, p1, p2, p3, p4, p5 = os.pullEvent()
  if e == "modem_message" then
   -- We received a reply from a modem
   local sSide, sChannel, sReplyChannel, tMessage, nDistance = p1, p2, p3, p4, p5
[...]

PR #181, opened 2017-05-03
Ensure GPS coordinates are numbers

SquidDev:
See #138

This can be tested by running the following code on a computer:

```lua
local modemSide = nil
for n, side in ipairs(rs.getSides()) do
    if peripheral.getType(side) == "modem" and peripheral.call(side, "isWireless") then
        modemSide = side
        break
    end
end

if not modemSide then error("Cannot find modem", 0) end

local modem = peripheral.wrap(modemSide)

modem.open(gps.CHANNEL_GPS)
while true do
    local _, side, channel, repl, message, distance = os.pullEvent("modem_message")
    if side == modemSide and channel == gps.CHANNEL_GPS and message == "PING" and distance then
        modem.transmit(repl, gps.CHANNEL_GPS, { {}, {}, {} })
    end
end
 ```

and then running `gps locate` on another. Without this patch the program will exit, with it it should work normally.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/gps
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/gps
@@ -103,7 +103,7 @@ function locate( _nTimeout, _bDebug )
 			local sSide, sChannel, sReplyChannel, tMessage, nDistance = p1, p2, p3, p4, p5
 			if sSide == sModemSide and sChannel == os.getComputerID() and sReplyChannel == CHANNEL_GPS and nDistance then
 				-- Received the correct message from the correct modem: use it to determine position
-				if type(tMessage) == "table" and #tMessage == 3 then
+				if type(tMessage) == "table" and #tMessage == 3 and type(tMessage[1]) == "number" and type(tMessage[2]) == "number" and type(tMessage[3]) == "number" then
 					local tFix = { vPosition = vector.new( tMessage[1], tMessage[2], tMessage[3] ), nDistance = nDistance }
 					if _bDebug then
 						print( tFix.nDistance.." metres from "..tostring( tFix.vPosition ) )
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dan200** [comment]
Is the vector part necessary?

File: src/main/resources/assets/computercraft/lua/rom/apis/vector [hunk 1]
```diff
@@ -76,9 +76,9 @@ local vmetatable = {
 
 function new( x, y, z )
 	local v = {
-		x = x or 0,
-		y = y or 0,
-		z = z or 0
+		x = type(x) == "number" and x or 0,
```
Vexatos: How about
```lua
x = tonumber(x) or 0
```

SquidDev:
The vector changes aren't strictly needed. It might be better to add type assertions instead, so it's easier to track this sort of error down in the future.

src/main/resources/assets/computercraft/lua/rom/apis/vector
SquidDev: Yes, that would be neater. I really should have spotted that 😦.

dan200: Make this change and i'll merge the patch. I'm generally ok with letting people misuse things if it only makes their *own* computer crash, but this is elegant.

SquidDev: Yeah, it is rather evil. I've had a prod through `rednet` and I can't see anything there which would cause similar issues.

src/main/resources/assets/computercraft/lua/rom/apis/gps
```diff
@@ -103,7 +103,7 @@ function locate( _nTimeout, _bDebug )
 			local sSide, sChannel, sReplyChannel, tMessage, nDistance = p1, p2, p3, p4, p5
 			if sSide == sModemSide and sChannel == os.getComputerID() and sReplyChannel == CHANNEL_GPS and nDistance then
 				-- Received the correct message from the correct modem: use it to determine position
-				if type(tMessage) == "table" and #tMessage == 3 then
+				if type(tMessage) == "table" and #tMessage == 3 and type(tMessage[1]) == "number" and type(tMessage[2]) == "number" and type(tMessage[3]) == "number" then
```
Vexatos: Should do it here, too.
```lua
tonumber(tMessage[1]) and tonumber(tMessage[2]) and tonumber(tMessage[3])
```

SquidDev: I'm less sure about this. Passing a string is technically correct, but it's going to be coming from gps host, which should never return strings.

Vexatos: Well the option is to either enforce a good standard, or support compatibility with breaking it.

PR #183, opened 2017-05-03
Add real world time support

timia2109:
This PR add an API to get real time and date. 

It support parsing and formatting of time and also to deal with timestamp. I use the PHP flavor without mills.

When it's ok, then I add the description in help and the API in whatsnew and changelog.


SquidDev:
I'm personally not a fan of leaking "real world" information into the game, such as the current time/day. However I guess you can do this with HTTP requests so it's not an issue which doesn't exist already.

IMO it would be nicer if it had the same methods names/functionality as [`os.date`](http://www.lua.org/manual/5.1/manual.html#pdf-os.date) and `os.time`. `os.date` acts like `format` and all the `get*` methods rolled into one. `os.time` just provides a way of converting a "date table" into a system specific number. This way you're providing something consistent with vanilla Lua behaviour, just under a different global name.

JLLeitschuh:
Time zone support?
Do you want it to be relative to the server? Do you want it to be UTC? Should it be relative to the time zone of the author of the script (probably totally impractical).

timia2109:
Yes your are right. I should implementiert time zone Support.

Restioson:
I think that it should be UTC for compatibility reasons. This could be useful for running a program every real day.

Restioson:
Also, if it's not already, it should use the ISO date formatting standard (yyyy-mm-dd).

xAnavrins:
Another option could simply be to return the unix epoch, then let the user deal with formatting and timezones.

timia2109:
So I replaced all functions with dynamic props. I havn't tested all so wait untill i finished it!

dan200:
One of the oldest design rules of CC has been "the computer should only know about the Minecraft world, not the real world", that's why the time uses the in-game time, it has a virtual filesystem, and HTTP was disabled by default. This kind of violates that.

That said, I'm a bit more lax about those things these days.

I have a suggestion to simplify this commit: instead of adding a whole new API, make it a parameter on os.time() and os.day(), like so:
os.time("local") returns the local real world time
os.time("utc") returns the UTC real world time
os.time() or os.time("ingame") return the in game time

And the same for os.day(). For the day number on "local" and "utc", you could use the number of days since January 1st, 1970, with that day being day 0.

Wojbie:
If things happen this way i really hope for simple "epoch" version next to "local" "ingame" "utc"

dan200:
The "local" and "utc" would still return values in the same range as the in-game time... a number between 0 and 24 for time, and an integer for day. Epoch doesn't fit that.

Vexatos:
Adding a string parameter to `os.time` like that would break [the Lua standard](http://www.lua.org/manual/5.1/manual.html#pdf-os.time), and we already have way too much of that anyway.

dan200:
Vexatos: The CC "os" API has nothing in common with the standard "os" API

timia2109:
Ok. I will rewrite it later and add this to the OS api. But I think the unix timestamp is important so I will add this!

Restioson:
@timia2109 btw build failed

timia2109:
I see. Testing now and push the bugfix then

Restioson:
@timia2109 before you push, please test it. This makes sure that it always works

timia2109:
So I add the following commands to the os.time( ... ) function:

|Function|Description|
|---|---|
|os.time("ingame")|Returns the hour of the game day|
|os.time("local")|Returns the hour of the day (localtime)|
|os.time("timestamp")|Returns the timestamp (Always UTC)|
|os.time("utc")|Returns the hour of the day (UTC)|

And I add the following functions to the os.day( ... ) function:

|Function|Description|
|---|---|
|os.day("ingame")|Returns the day of the game|
|os.day("local")|Returns the days since 1970-01-01 (localtime)|
|os.day("utc")|Returns the days since 1970-01-01 (UTC)|


2017-05-06 14_29_21-minecraft 1 9 4

src/main/java/dan200/computercraft/core/apis/OSAPI.java
```diff
@@ -10,10 +10,7 @@
 import dan200.computercraft.api.lua.LuaException;
 import dan200.computercraft.shared.util.StringUtil;
 
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
+import java.util.*;
```
timia2109: IntelliJ does some funny things... 👎

Restioson:
I am confused as to what `os.day` does. What is it returning? I think that you should return it in the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format. Example of date & time: `2017-05-06T10:39:41+00:00` (for gmt +00:00) - credit: Wikipedia.

timia2109:
@Restioson see Dan200s comment 😅

dan200:
Merging this in, but going to make soem changes to make it work more correctly.

Vexatos:
This apparently did not help...
grafik

Cruor:
Could we rather implement `os.time` the way it works in standard Lua for creating a better teaching environment? Most legacy program will not use this function directly, as it is not usable for delta times (`os.clock` is) and will most likely pass it in as an argument to `textutils.formatTime` as the function itself gives a so absurd result you wouldn't even know what to do about it. This is a major version change as well as a minecraft version change. Breaking a tiny bit of backwards compat should not hurt if documented properly in the "What new".

SquidDev:
To add to @Cruor's point, CC has made "breaking changes" before. For example, [`rednet.receive`](http://www.computercraft.info/wiki/Rednet.receive) used to return the distance and now returns the protocol. [Several methods](http://www.computercraft.info/wiki/Category:Deprecated_API_Functions) have even been removed. For such a minor function as `os.time` I don't think there will be a major impact of changing the implementation of `os.time` and `os.date` to mirror that of PUC Lua.

dan200:
CC has avoided breaking changes wherever possible, I can only think of one
instance (term.restore) where a widely used method was removed, and it was
done reluctantly.

The CC "os" api has absolutely nothing to do with the standard Lua "os"
API. They have the same name a couple of similar-but-distinct methods and
that's it.

oddstr13:
I seem to remember there where changes to how the turtle api works, from using events to hidden in a java black box. There is also the case of you deliberatly breaking my door lock program, which used escape to detect user leaving the terminal.

dan200:
I don't count changing undocumented implementation details. The documented behaviour of the turtle API hasn't changed. The escape key was a sandboxing oversight.

Vexatos:
I guess someone will have to do another PR to fix this eventually.

PR #184, opened 2017-05-03
Sort the result of FileSystem.list

SquidDev:
This ensures `fs.list` and `fs.find` always return the same result.

For some reason, the ComputerCraft jar was being packaged differently on some platforms, causing files to appear in a different order. As computers depend on the `colors` API being loaded before `colours`, we need to ensure that they are loaded in a consistent order.

I'm not sure this is the best solution - just sorting in `JarMount`  would be sufficient. We could also sort inside the `bios.lua` file instead. Or manually load `colors` before `colours`.


SquidDev:
Thanks!

PR #190, opened 2017-05-04
Implemented require.

oddstr13:
See https://bitbucket.org/openshell/cc-require/ for original repo and more info.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
@@ -1,0 +1,124 @@
+-- -*- coding: utf8 -*-
+-- Copyright (c) 2014 Odd Straaboe <oddstr13 at openshell dot no>
+-- License: MIT - http://opensource.org/licenses/MIT
+-- Filename: require.lua
+
+_G.package = {}
+
+_G.package.cpath = ""
+_G.package.loaded = {}
+_G.package.loadlib = function() error("NotImplemented: package.loadlib") end
+_G.package.path = table.concat({
+    "?",
+    "?.lua",
+    "?/init.lua",
+    "/lib/?",
+    "/lib/?.lua",
+    "/lib/?/init.lua",
+    "/rom/apis/?",
+    "/rom/apis/?.lua",
+    "/rom/apis/?/init.lua",
+    "/rom/apis/turtle/?",
+    "/rom/apis/turtle/?.lua",
+    "/rom/apis/turtle/?/init.lua",
+    "/rom/apis/command/?",
+    "/rom/apis/command/?.lua",
+    "/rom/apis/command/?/init.lua",
+}, ";")
+_G.package.preload = {}
+_G.package.seeall = function(module) error("NotImplemented: package.seeall") end
+_G.module = function(m) error("NotImplemented: module") end
+
+local _package_path_loader = function(name)
+    
+    local fname = name:gsub("%.", "/")
+    
+    for pattern in package.path:gmatch("[^;]+") do
+        
+        local fpath = pattern:gsub("%?", fname)
+        
+        if fs.exists(fpath) and not fs.isDir(fpath) then
+            
+            local apienv = {}
+            setmetatable(apienv, {__index = _G})
+            
+            local apifunc, err = loadfile(fpath)
+            local ok
+            
+            if apifunc then
+                setfenv(apifunc, apienv)
+                ok, err = pcall(apifunc)
+            end
+            
+            if not apifunc or not ok then
+                error("error loading module '" .. name .. "' from file '" .. fpath .. "'\n\t" .. err)
+            end
+            
+            local api = {}
+            if type(err) == "table" then
+              api = err
+            end
+            for k,v in pairs( apienv ) do
+                api[k] = v
+            end
+            
+            return api
+        end
+    end
+end
+
+_G.package.loaders = {
+    function(name)
+        if package.preload[name] then
+            return package.preload[name]
+        else
+            return "\tno field package.preload['" .. name .. "']"
+        end
+    end,
+    
+    function(name)
+        local _errors = {}
+        
+        local fname = name:gsub("%.", "/")
+        
+        for pattern in package.path:gmatch("[^;]+") do
+            
+            local fpath = pattern:gsub("%?", fname)
+            if fs.exists(fpath) and not fs.isDir(fpath) then
+                return _package_path_loader
+            else
+                table.insert(_errors, "\tno file '" .. fpath .. "'")
+            end
+        end
+        
+        return table.concat(_errors, "\n")
+    end
+}
+
+_G.require = function(name)
+    if package.loaded[name] then
+        return package.loaded[name]
+    end
+    
+    local _errors = {}
+    
+    for _, searcher in pairs(package.loaders) do
+        local loader = searcher(name)
+        if type(loader) == "function" then
+            local res = loader(name)
+            if res ~= nil then
+                package.loaded[name] = res
+            end
+            
+            if package.loaded[name] == nil then
+                package.loaded[name] = true
+            end
+            
+            return package.loaded[name]
+        elseif type(loader) == "string" then
+            table.insert(_errors, loader)
+        end
+    end
+    
+    error("module '" .. name .. "' not found:\n" .. table.concat(_errors, "\n"))
+end
--- a/src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
@@ -7,118 +7,138 @@ _G.package = {}
 
 _G.package.cpath = ""
 _G.package.loaded = {}
-_G.package.loadlib = function() error("NotImplemented: package.loadlib") end
+_G.package.loadlib = function() return nil, "not implemented: package.loadlib" end
 _G.package.path = table.concat({
-    "?",
-    "?.lua",
-    "?/init.lua",
-    "/lib/?",
-    "/lib/?.lua",
-    "/lib/?/init.lua",
-    "/rom/apis/?",
-    "/rom/apis/?.lua",
-    "/rom/apis/?/init.lua",
-    "/rom/apis/turtle/?",
-    "/rom/apis/turtle/?.lua",
-    "/rom/apis/turtle/?/init.lua",
-    "/rom/apis/command/?",
-    "/rom/apis/command/?.lua",
-    "/rom/apis/command/?/init.lua",
+  "?",
+  "?.lua",
+  "?/init.lua",
+  "/lib/?",
+  "/lib/?.lua",
+  "/lib/?/init.lua",
+  "/rom/apis/?",
+  "/rom/apis/?.lua",
+  "/rom/apis/?/init.lua",
+  "/rom/apis/turtle/?",
+  "/rom/apis/turtle/?.lua",
+  "/rom/apis/turtle/?/init.lua",
+  "/rom/apis/command/?",
+  "/rom/apis/command/?.lua",
+  "/rom/apis/command/?/init.lua",
 }, ";")
 _G.package.preload = {}
-_G.package.seeall = function(module) error("NotImplemented: package.seeall") end
-_G.module = function(m) error("NotImplemented: module") end
+_G.package.seeall = function(module)
+  if type(module) ~= "table" then
+    error("bad argument #1 to 'require' (table expected, got " .. type(module) .. ")", 2)
+  end
 
-local _package_path_loader = function(name)
-    
-    local fname = name:gsub("%.", "/")
-    
-    for pattern in package.path:gmatch("[^;]+") do
-        
-        local fpath = pattern:gsub("%?", fname)
-        
-        if fs.exists(fpath) and not fs.isDir(fpath) then
-            
-            local apienv = {}
-            setmetatable(apienv, {__index = _G})
-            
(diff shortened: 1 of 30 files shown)


dan200:
Here's my objection to this (and why require wasn't included in CC originally, though I knew a lot less abotu Lua then than I do now):
Because package.loaded is global, if you "require" something in a program, edit it, then run a program again, you won't see the changes until you reboot the computer. This breaks the "virtual computer" metaphor that CC rests on.
In CraftOS 2.0 I solved this problem by running every program in it's own isolated environment with it's own version of the package tables, but that's far too invasive a change to make in CraftOS 1.0 without breaking most user programs.

Vexatos:
That is how it works in normal Lua too. If you want to apply changes of a `require`d file, you remove it from `package.loaded`.

dan200:
I know how normal lua works. But on a real computer you're starting a new Lua VM every time you run a program, on a CC computer there's one Lua VM that persists until you turn the computer off.

oddstr13:
I didn't suggest that you should remove os.loadApi

Vexatos:
Exactly. The old way isn't gone. This is just for people who know Lua to have less of a headache coding in CC.

CrazedProgrammer:
You could implement your own require in bios.lua, so it is isolated between computers.

Vexatos:
Oh, it is isolated between computers, that is not the problem. The issue dan has is package.loaded persisting across programs (just like in standard Lua).

apemanzilla:
It would be simple enough to get rid of caching entirely if it's a problem.

Vexatos:
I'd say that we might as well implement require according to the standard since it's meant for people who are more used to it than to `os.loadAPI`.

apemanzilla:
Fair enough.

Would the license need to be changed to match the CC license?

oddstr13:
I don't think MIT is going to pose a problem, as it's very permissive.
If included as is, it would be under dual license, MIT (this file) and MMPL (the project as a whole, including this file).

apemanzilla:
Alright, that's good. What about integrating `package.path` with the `settings` API?

SquidDev:
I also feel the `turtle/` and `command/` folders should only be added to the path if their respective globals are defined. I'm not sure if it makes sense to add any of those folders at all though, as they are not compatible with `require`'s method of loading.

oddstr13:
I see no reason to not include `turtle/` and `command/` by default, as the APIs provided there nope out in any case.
```lua
if not turtle then
	error( "Cannot load turtle API on computer", 2 )
end
```

Should probably look into integrating it with the settings API, however I am not currently familiar with it.

apemanzilla:
Settings API is pretty simple, you could just do `package.path = settings.get("package.path", [default value])`

SquidDev:
Another couple of things:

 - It might be sensible for `os.loadAPI` to populate `package.loaded` (if a value isn't already there). This *could* be confusing, but also allows you to do `require "rednet"` or something.
 - On a similar vein, all of CC's other APIs (`os`, fs`, `io`, `peripheral`, `bit`, `bit32`, `rs`, etc...) should probably be registered under `package.loaded` too.
 - `package.config` isn't defined. I believe it should just  be `"/\n;\n?\n!\n-"`.
 - You're using `setfenv` - use the second argument to `loadfile` instead (it uses Lua 5.2+ conventions).
 - From what I can tell, the loader inside `package.loaders` loops over the path, and then calls `_package_path_loader` which does the same thing again. From what I can tell from reading Lua's source is that the function in `package.preload` should load the file and return it without calling it (or an error message if it failed). We should be able to do the same, but wrapping the loaded function with the global copying code.

Vexatos:
Problem is that `os.loadAPI` and `require` work on different planes of ~~sanity~~. `require`d libraries have to return the library itself while `os.loadAPI` just adds everything to a table in _G.

src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
```diff
@@ -0,0 +1,124 @@
...
+    for pattern in package.path:gmatch("[^;]+") do
+        
+        local fpath = pattern:gsub("%?", fname)
+        
+        if fs.exists(fpath) and not fs.isDir(fpath) then
+            
+            local apienv = {}
+            setmetatable(apienv, {__index = _G})
+            
+            local apifunc, err = loadfile(fpath)
```
apemanzilla: This should be updated to use either `load(code, name, "t", env)` or `loadfile(file, "t", env)` instead of `loadfile(file)` and `setfenv(...)`, for Lua 5.2 compatibility.

src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
```diff
@@ -0,0 +1,124 @@
...
+                setfenv(apifunc, apienv)
+                ok, err = pcall(apifunc)
+            end
+            
+            if not apifunc or not ok then
+                error("error loading module '" .. name .. "' from file '" .. fpath .. "'\n\t" .. err)
+            end
+            
+            local api = {}
+            if type(err) == "table" then
```
apemanzilla: While it's not the way real Lua works, if APIs return something that's not a table (e.g. they return a single function instead) this won't work. I'd change this whole section to something like:
```lua
if ok then
  if err ~= nil then
    api = err
  else
    -- os.loadAPI compatibility code
  end
else
  -- error handling
end
```

src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
```diff
@@ -0,0 +1,124 @@
...
+        end
+        
+        return table.concat(_errors, "\n")
+    end
+}
+
+_G.require = function(name)
+    if package.loaded[name] then
+        return package.loaded[name]
+    end
```
SquidDev: Ideally we could check for package loops here (a requires b, b requires a). Lua does this by storing a magic value in `package.loaded` before loading it, and checking it is equal to that here. Note that this check will also fail if a package returns `false`.

src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
```diff
@@ -0,0 +1,124 @@
...
+            
+            local apienv = {}
+            setmetatable(apienv, {__index = _G})
+            
+            local apifunc, err = loadfile(fpath)
+            local ok
+            
+            if apifunc then
+                setfenv(apifunc, apienv)
+                ok, err = pcall(apifunc)
```
SquidDev: I don't think `pcall` is needed here - Lua doesn't include an "error loading module" message when the file itself errors.

apemanzilla: This could have been done for consistency (`os.loadAPI` will catch errors are merely return `false`) but I agree, it would be better to allow the program to handle errors.

SquidDev: Well the error is re-thrown a bit later, so does still raise an error rather than returning `false`. It's just you lose the ability to use `xpcall` on required modules (and it's less consistent with vanilla Lua).

apemanzilla: Yeah, `xpcall` is what I was thinking of as well.

src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
```diff
@@ -0,0 +1,144 @@
...
+_G.package.loaders = {
+  function(name)
+    if package.preload[name] then
+      return package.preload[name]
+    else
+      return "\tno field package.preload['" .. name .. "']"
+    end
+  end,
+
+  function(name)
```
apemanzilla: I can't test the code right now, but it doesn't look like the directory of the calling program is checked. It would be nice if `shell.getRunningProgram` and `fs.getDir` were used to allow loading libraries from the same directory as the program.

Vexatos: It checks `"?"` which is the current working directory. The directory of the program should not be part of the path by default. If the FS API does not properly set the current working directory, that's a bug in there and not relevant to this PR.

apemanzilla: It's not a bug, it's how CC works. The only place the working directly is taken into account is the `shell` API. Everything else functions from the filesystem root.

oddstr13: Thou shalt not use the shell api in libraries.

apemanzilla: Unfortunately that seems to be the only way 😦

SquidDev: The issue is that the `require` library has no clue which shell the caller is using, so we've cannot determine the current directory. There are a couple of potential solutions, but all are equally ugly:

 - Use `getfenv(2).shell` to access the parent shell. We've trying to move away from Lua 5.1 functions though, so this isn't a good idea.
 - Recreate the `package` for each shell, with new loaders which work from the current directory. This makes extending `package.loaders` much harder though.
 - Recreate `require` for each shell, and pass the current directory as the second argument. This seems like the nicest of these, though it means it is like PUC Lua.

Vexatos: * Do not care and leave it as-is. Put your libraries in proper paths like you should be doing anyway, worry about fixing this part of CC some other time.

src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
```diff
@@ -0,0 +1,144 @@
...
+  "?",
+  "?.lua",
+  "?/init.lua",
+  "/lib/?",
+  "/lib/?.lua",
+  "/lib/?/init.lua",
+  "/rom/apis/?",
+  "/rom/apis/?.lua",
+  "/rom/apis/?/init.lua",
+  "/rom/apis/turtle/?",
```
dan200: Only add these on turtles

src/main/resources/assets/computercraft/lua/rom/autorun/10_require.lua
```diff
@@ -0,0 +1,144 @@
...
+  "/rom/apis/turtle/?.lua",
+  "/rom/apis/turtle/?/init.lua",
+  "/rom/apis/command/?",
```
dan200: Only add these on command computers

spike11302000:
i would leave os.loadApi for backwards compatibility

Vexatos:
As I said, it was never removed. This API only _adds_ things.

dan200:
Hey all,

I've committed a modified version of this PR as https://github.com/dan200/ComputerCraft/commit/61b2ed36a96c4c1977d39b9bc27c644f311a886c which solves my issues with requiring into the global namespace, by implementing require as part of the shell, and giving each shell-invoked program it's own sandbox.

This means that when a program requires something, it doesn't pollute the global namespace, and they only need to restart the program not the whole computer for their edits to have effect. This fits much better with the "virtual computer" model that is the base of CraftOS, and makes running lua programs from CraftOS much more like running lua programs from bash or cmd.

Thanks for the PR!

Vexatos:
1) It would still be nice if you referenced the original license.

2) `package.loaders` loaders now don't actually return functions anymore? Is there any reason to go non-standard here?

dan200:
1) I'm not sure what you're refering to here? I mostly used CraftOS 2.0 code (which follows standard lua much more slavishly than CC does) as a reference here.

2) Check again. Both the default loaders return functions.

3) Nothing in the default CC rom is intended to be required so it doesn;t make sense to include it by default. However: as each program gets it's own package.path now, you're free to add whatever you want to it in your programs before calling require. 

4) The purpose of this code is to allow users to require files written in the standard lua style, not the awful os.loadAPI style :)

Vexatos:
> the awful os.loadAPI

I am glad there is an alternative now, anyway.

Also, the environment used in `loadfile` doesn't seem to reference _G in any way... Can it still be accessed?

Vexatos:
>I've committed a modified version of this PR

>I mostly used CraftOS 2.0 code

Now I am confused. Which one did you base it on? If it is based on the PR (which it looks like, guessing from the code), you should include the license, in my opinion.

Vexatos:
>I've committed a modified version of this PR

>I mostly used CraftOS 2.0 code

Now I am confused. Which one did you base it on? If it is based on the PR (which it looks like, guessing from the code), you should include the license, in my opinion.

Furthermore, are you sure `_G` can be accessed from `require`d libraries anyway? Normally there would be a reference to `_G` in `__index` (e.g. in `os.run`), but `require` doesn't seem to do that.

dan200:
I didn't copy-paste any code from the PR, but it's obviously going to be similar as it's an implementation of the same spec.  And yep: require'd stuff runs in the same environment as the calling program, which has it's __index metavalue set to _G by os.run. Feel free to try it for yourself.

Vexatos:
It appears to run in [its own custom environment](https://github.com/dan200/ComputerCraft/commit/61b2ed36a96c4c1977d39b9bc27c644f311a886c#diff-61461ec6effd6481aaf04c7f16a4153bR44) which doesn't reference `_G` at all, which is why I am confused.

dan200:
Please actually try it out and see for yourself. The env returned by createShellEnv() is subsequently passed into os.run(), where it has it's metatable set. "require" calls will happen after that point, so by the time that method is called, _G will be in place.

oddstr13:
I would've preferred a fully standards compatible implementation, but as long as it works the same on a basic level it's fine with me.

PR #191, opened 2017-05-04
Make monitor text glow in the dark

SquidDev:
An alternative to #177, monitor text now glows in the dark. I'm putting this in a separate PR as it shares nothing with the original.

Closes #177.

### DIFF
--- a/src/main/java/dan200/computercraft/client/render/TileEntityMonitorRenderer.java
+++ b/src/main/java/dan200/computercraft/client/render/TileEntityMonitorRenderer.java
@@ -104,6 +104,7 @@ public class TileEntityMonitorRenderer extends TileEntitySpecialRenderer<TileMon
             // Draw the contents
             GlStateManager.depthMask( false );
             GlStateManager.disableLighting();
+            mc.entityRenderer.disableLightmap();
             try
             {
                 if( terminal != null )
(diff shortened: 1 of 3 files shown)


Vexatos:
Much better than #177.

dan200:
Hmm... now I wish I hadn't used such a light shade of grey for "black" on displays. This would look so much nicer with more contrast! And it's all because the Mac I used to use had a dumb gamma curve. Going to check this out locally and do some experiments with rebalancing colours.

Vexatos:
The black not actually being dark has always bothered me. Would be nice to see the colours tweaked a bit.

apemanzilla:
You could kill two birds with one stone and implement a color palette system 😛

vpontin:
Amazing addition! Thanks for everything!

dan200:
apemanzilla: If anyone wants to do that I would merge it for sure :) I'm thinking an API something like:
[term|monitor].setColor(color, r, g, b)
Would have to make sure it got reset on reboot though.

apemanzilla:
I'll take a stab at it in a bit.

dan200:
Can someone else check out this feature branch and run it? Monitors aren't rendering any differently than normal for me :s

SquidDev:
Updated screenshot using the new terminal colours. The effect is small, but definitely an improvement!



> Monitors aren't rendering any differently than normal for me :s

That's really weird. Just wondering what your set up looks like - I might have missed something really obvious.

apemanzilla:
Works for me.

Screenshot

incinirate:
Me as well
java_2017-05-04_19-47-32

PR #195, opened 2017-05-05
Fix selection boxes not being shown

SquidDev:
I really should have spotted this earlier. Sorry.

`Block.getSelectedBoundingBox` expects a bounding box relative to (0, 0, 0) but we were returning one relative to the current block. Instead we allow the default behaviour to continue, which will call `BlockGeneric.getBoundingBox` and offset it.

---

I'm looking at the implementations here and wonder if we really need to access the `TileEntity` at all here? The only implementation which requires the TE, rather than being able to use the block state is turtles. Perhaps it would be better to hoist some, if not all, of this code into the appropriate sub classes.

Also @dan200, would you be OK with me changing `TileCable`'s `getCollisionBounds` to handle each cable segment separately? This way you get nicer entity collisions with the cables.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/common/BlockGeneric.java
+++ b/src/main/java/dan200/computercraft/shared/common/BlockGeneric.java
@@ -219,6 +219,12 @@ public abstract class BlockGeneric extends Block implements
     }
 
     @Override
+    public AxisAlignedBB getSelectedBoundingBox( IBlockState state, World worldIn, BlockPos pos )
+    {
+        return getBoundingBox( state, worldIn, pos ).offset( pos );
+    }
+
+    @Override
     public final AxisAlignedBB getCollisionBoundingBox( IBlockState state, World world, BlockPos pos )
     {
         TileEntity tile = world.getTileEntity( pos );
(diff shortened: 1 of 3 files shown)


src/main/java/dan200/computercraft/shared/common/BlockGeneric.java
```diff
@@ -219,12 +219,6 @@ public final AxisAlignedBB getBoundingBox( IBlockState state, IBlockAccess world
     }
 
     @Override
-    public final AxisAlignedBB getSelectedBoundingBox( IBlockState state, World world, BlockPos pos )
```
dan200: Rather than deleting this method, could we implement it by calling getBoundingBox() and adding the blockpos?

dan200:
Yep, i'm happy for you to make getCollisionBounds more accurate :)

src/main/java/dan200/computercraft/shared/common/BlockGeneric.java
SquidDev: I'm happy to, though that is the default behaviour already. I guess it might be a tiny bit quicker as it doesn't have to go via `IBlockState` again.

dan200: My concern was small, but to do with clarity. When versions change it's good to see all the stuff you have to change in one place :)

SquidDev:
Right, all done. It might be possible to get the selection box to "fit" the cable too, but I think I'd want to do that in a separate PR - there are some other modem QOL features which could go there too.

Issue #193, opened 2017-05-04
Palettes


apemanzilla:
Given the complexity and nuances of a palette system, I wanted to open an issue for discussion on the topic.

The way I would implement it would be a system where you can change the RGB values for colors at runtime, but are still limited to 16 colors total. This gives you a lot of flexibility if you're designing something complex like a game or OS, while not increasing bandwidth or overcomplicating things.

However, there's still other issues - how would it work with the `window` API? Should there be rate limiting or similar to prevent people switching colors rapidly? Should `term.setColor` take a single value for hex (`0xff00ff`) or three separate values (`255, 0, 255`)?

dan200:
Yep. If you're adding colours it's not a palette anymore, it's just a true
colour display. This would just be a means to modify the 16 default values.

Rate limiting shouldn't be necessary: changing 16 32-bit values is far less
data than modifying a whole screen of text and we don't rate-limit that.

The details of the "window" API elude my memory to be honest, but it's
unavoidable that if different windows wanted different colours they won't
all be able to get it, so either don't expose colours through that API, or
accept that the screen colours will change whenever you switch window/make
one visible. As long as multishell correctly switches palettes when
switching tabs I don't mind the details too much

I wouldn't expose the details of RGB encoding to lua, i'd just accept 3
floats in the 0-1 range.

KingofGamesYami:
Rate limiting would be kinda weird.  I'd rather make it a relatively slow function that chews up execution time so that people avoid doing it in a loop.  Even if it's just setting the same parameters repeatedly, that's going to be redundant and cause problems with simultaneous execution.

I'd go with (R, G, B) myself, I find it very easy to understand.  That's the way code.org does it.

Messen:
A some nice color cycling effects crossed my mind after reading this.

Lignum:
I've started work on this, I'll make a WIP PR soon.

apemanzilla:
What about having black and white (maybe also grays) being unmodifiable, so they can be used for things like multishell tabs while making sure that they look the same?

dan200:
Hmm. I hadn't thought about the greyscale displays. My thinking would be:
start letting you use all the colours on them, but normalise everything to
greyscale during rendering (by averaging the 3 colour components).

BombBloke:
Re the window API, it consists of one function which returns a terminal object. If you redirect to an object and then try to call a term function that object doesn't support, you get a crash. So "completely leaving functions out" isn't really an option.

[...]

PR #197, opened 2017-05-05
Modifiable terminal colour palette

Lignum:
As discussed in #193.

This adds the following terminal functions:
- `setColour(colour, r, g, b)` - Sets the RGB of a colour. `colour` is the colour from the `colours` API, `r`, `g` and `b` are specified as numbers in the range 0-1. Will (should) error when used on a greyscale computer.
- `getColour(colour)` - Gets the RGB of a colour.

### DIFF
--- a/src/main/java/dan200/computercraft/client/gui/GuiPrintout.java
+++ b/src/main/java/dan200/computercraft/client/gui/GuiPrintout.java
@@ -205,7 +205,7 @@ public class GuiPrintout extends GuiContainer
             int lineIdx = ItemPrintout.LINES_PER_PAGE * m_page + line;
             if( lineIdx >= 0 && lineIdx < m_text.length )
             {
-                fontRenderer.drawString( m_text[lineIdx], x, y, m_colours[lineIdx], null, 0, 0, false, new Palette() );
+                fontRenderer.drawString( m_text[lineIdx], x, y, m_colours[lineIdx], null, 0, 0, false, Palette.DEFAULT );
             }
             y = y + FixedWidthFontRenderer.FONT_HEIGHT;
         }
--- a/src/main/java/dan200/computercraft/client/gui/GuiPrintout.java
+++ b/src/main/java/dan200/computercraft/client/gui/GuiPrintout.java
@@ -10,6 +10,7 @@ import dan200.computercraft.ComputerCraft;
 import dan200.computercraft.core.terminal.TextBuffer;
 import dan200.computercraft.shared.media.inventory.ContainerHeldItem;
 import dan200.computercraft.shared.media.items.ItemPrintout;
+import dan200.computercraft.shared.util.Palette;
 import net.minecraft.client.gui.inventory.GuiContainer;
 import net.minecraft.client.renderer.GlStateManager;
 import net.minecraft.util.ResourceLocation;
@@ -204,7 +205,7 @@ public class GuiPrintout extends GuiContainer
             int lineIdx = ItemPrintout.LINES_PER_PAGE * m_page + line;
             if( lineIdx >= 0 && lineIdx < m_text.length )
             {
-                fontRenderer.drawString( m_text[lineIdx], x, y, m_colours[lineIdx], null, 0, 0, false );
+                fontRenderer.drawString( m_text[lineIdx], x, y, m_colours[lineIdx], null, 0, 0, false, new Palette() );
             }
             y = y + FixedWidthFontRenderer.FONT_HEIGHT;
         }
--- a/src/main/java/dan200/computercraft/core/apis/TermAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/TermAPI.java
@@ -74,10 +74,10 @@ public class TermAPI implements ILuaAPI
             "getBackgroundColour",
             "getBackgroundColor",
             "blit",
-            "setColour",
-            "setColor",
-            "getColour",
-            "getColor"
+            "setPaletteColour",
+            "setPaletteColor",
+            "getPaletteColour",
+            "getPaletteColor"
         };
     }
     
--- a/src/main/java/dan200/computercraft/core/apis/TermAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/TermAPI.java
@@ -10,11 +10,12 @@ import dan200.computercraft.api.lua.ILuaContext;
 import dan200.computercraft.api.lua.LuaException;
 import dan200.computercraft.core.computer.IComputerEnvironment;
 import dan200.computercraft.core.terminal.Terminal;
+import org.apache.commons.lang3.ArrayUtils;
 
 public class TermAPI implements ILuaAPI
 {
-    private Terminal m_terminal;
-    private IComputerEnvironment m_environment;
+    private final Terminal m_terminal;
+    private final IComputerEnvironment m_environment;
 
     public TermAPI( IAPIEnvironment _environment )
     {
@@ -67,13 +68,19 @@ public class TermAPI implements ILuaAPI
             "getTextColor",
             "getBackgroundColour",
             "getBackgroundColor",
-            "blit"
+            "blit",
+            "setColour",
+            "setColor",
+            "getColour",
+            "getColor",
+            "resetColour",
+            "resetColor"
         };
     }
     
     public static int parseColour( Object[] args, boolean _enableColours ) throws LuaException
     {
-        if( args.length != 1 || args[0] == null || !(args[0] instanceof Double) )
+        if( args.length < 1 || args[0] == null || !(args[0] instanceof Double) )
         {
             throw new LuaException( "Expected number" );
         }            
@@ -160,7 +167,7 @@ public class TermAPI implements ILuaAPI
                 {
                     throw new LuaException( "Expected boolean" );
                 }
-                boolean b = ((Boolean)args[0]).booleanValue();
+                boolean b = (Boolean) args[ 0 ];
                 synchronized( m_terminal )
                 {
                     m_terminal.setCursorBlink( b );
@@ -267,6 +274,55 @@ public class TermAPI implements ILuaAPI
                 {
                     m_terminal.blit( text, textColour, backgroundColour );
                     m_terminal.setCursorPos( m_terminal.getCursorX() + text.length(), m_terminal.getCursorY() );
+                }
+                return null;
+            }
+            case 19:
+            case 20:
+            {
+                // setColour/setColor
+                if ( args.length < 4 || !(args[0] instanceof Double) || !(args[1] instanceof Double) || !(args[2] instanceof Double) || !(args[3] instanceof Double) ) // toil and trouble
+                {
+                    throw new LuaException( "Expected number, number, number, number" );
+                }
+
+                if ( !m_environment.isColour() )
+                {
+                    // Make sure you can't circumvent greyscale terminals with this function.
+                    throw new LuaException( "Colour not supported" );
+                }
+
+                int colour = 15 - parseColour( args, m_environment.isColour() );
+                float r = ((Double)args[1]).floatValue();
+                float g = ((Double)args[2]).floatValue();
+                float b = ((Double)args[3]).floatValue();
+
+                synchronized( m_terminal )
+                {
+                    if( m_terminal.getPalette() != null )
+                    {
+                        m_terminal.getPalette().setColour( colour, r, g, b );
+                    }
+                }
+                return null;
+            }
+            case 21:
+            case 22:
+            {
+                // getColour/getColor
+                if (args.length < 1 || !(args[0] instanceof Double))
+                {
+                    throw new LuaException( "Expected number" );
+                }
+
+                int colour = 15 - parseColour( args, m_environment.isColour() );
+
+                synchronized( m_terminal )
+                {
+                    if ( m_terminal.getPalette() != null )
+                    {
+                        return ArrayUtils.toObject( m_terminal.getPalette().getColour64( colour ) );
+                    }
                 }
                 return null;
             }
--- a/src/main/java/dan200/computercraft/core/apis/TermAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/TermAPI.java
@@ -120,50 +120,6 @@ public class TermAPI implements ILuaAPI
         }
     }
 
-    public static void setColour( Terminal terminal, HashMap<Object, Object> colours) throws LuaException
-    {
-        final double lg2 = Math.log( 2 );
-
-        for(Map.Entry<Object, Object> e : colours.entrySet())
-        {
-            if(e.getKey() instanceof Double)
-            {
-                int index = 15 - (int)( Math.log( (Double)e.getKey() ) / lg2 );
-
-                try
-                {
-                    if (e.getValue() instanceof HashMap)
-                    {
-                        @SuppressWarnings({ "unchecked" }) // There isn't really a nice way around this :(
-                        HashMap<Object, Object> colour = (HashMap<Object, Object>) e.getValue();
-
-                        setColour(
-                                terminal,
-                                index,
-                                ( (Double)colour.get( 1.0 ) ).floatValue(),
-                                ( (Double)colour.get( 2.0 ) ).floatValue(),
-                                ( (Double)colour.get( 3.0 ) ).floatValue()
-                        );
-                    }
-                    else if (e.getValue() instanceof Double)
-                    {
-                        float[] rgb = Palette.decodeRGB8( ((Double)e.getValue()).intValue() );
-
-                        setColour(
-                                terminal,
-                                index,
-                                rgb[0], rgb[1], rgb[2]
-                        );
-                    }
-                }
-                catch(ClassCastException cce)
-                {
-                    throw new LuaException( "Malformed colour table " + cce.getMessage() );
-                }
-            }
-        }
-    }
-
     @Override
     public Object[] callMethod( ILuaContext context, int method, Object[] args ) throws LuaException
     {
@@ -337,14 +293,6 @@ public class TermAPI implements ILuaAPI
             case 20:
             {
                 // setPaletteColour/setPaletteColor
-                if(args.length >= 1 && args[0] instanceof HashMap)
-                {
-                    @SuppressWarnings( { "unchecked" } ) // There isn't really a nice way around this :(
-                    HashMap<Object, Object> colourTbl = (HashMap<Object, Object>)args[0];
-                    setColour( m_terminal, colourTbl );
-                    return null;
-                }
-
                 if(args.length == 2 && args[0] instanceof Double && args[1] instanceof Double)
                 {
                     int colour = 15 - parseColour( args, true );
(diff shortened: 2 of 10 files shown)


apemanzilla:
What about having a set of unmodifiable colors, such as white and black (and maybe the grays as well)? This would allow programs like `multishell` to guarantee that the tabs will always look the same regardless of what palette the running program uses, and there's still a lot of flexibility with 14/12 custom colors.

dan200:
Nah. Either let people change them or not. If things look weird that's down
to them.

Lignum:
@apemanzilla I'm not a fan of arbitrary restrictions. If anything, this should be implemented on the Lua side of things, and I don't think CraftOS or even multishell specifically should do that. It should be left up to the user to install an OS or whatever that does this.

Lignum:
Alright, so I've fixed up the `window` API. You can now have two different palettes in two different multishell tabs and it will switch just fine. I would guess that some weird things will happen if there are multiple windows on-screen, but I don't know how feasible it would be to fix that. This also means you can now call `term.setColour` directly, instead of through `term.native()`.

Thus, the PR is feature complete, but not tested thoroughly, so I'll remove the WIP label, but I still don't advise this to be merged right away.

Lignum:
As of ac2382a86175f154a351986b7d57ec5923495e36, it is possible to set multiple palette colours at once like this:
```Lua
term.setColour({
  [colours.white] = { 0.0, 0.0, 0.0 },
  [colours.black] = { 1.0, 1.0, 1.0 },
  -- ...
})

**Comment by oddstr13** [comment]
why are colors specified as rgb float 0-1 instead of rgb int 0-255, or even a #rrggbb hex string?

**Comment by Wojbie** [comment]
> why are colors specified as rgb float 0-1 instead of rgb int 0-255, or even a #rrggbb hex string?
#193

**Comment by Lignum** [comment]
I've explained this in the issue already. It's because the 0-1 range is used in OpenGL, is easy to work with mathematically and is independent of the framebuffer colour depth. In the issue I've talked about maybe adding a `colours.rgb8`, which allows you to specify colours encoded as RGB8 in hex or as individual channels.

... and also because @dan200 said so.

**Comment by Vexatos** [comment]
> It's because the 0-1 range is used in OpenGL

It is [fairly easy to convert](https://github.com/asiekierka/Computronics/blob/af6f70fe6ac883d5fe2012639ff35d999a4bde06/src/main/java/pl/asie/computronics/oc/ColorfulUpgradeHandler.java#L65).

**Comment by Lignum** [comment]
I'm [well aware](https://github.com/dan200/ComputerCraft/pull/197/files#diff-2166e24ab232bafe9dd492f64b2e6482R56) :stuck_out_tongue:, just sticking with the flow.

> 0-255 or 0xRRGGBB colours are much more common pretty much wherever you look.

Depends on where you look. Coming from graphics programming with OpenGL, I see 0-1 more often. Either way, the subjectiveness of this issue is why I proposed adding a `colours.rgb8` function.

**Comment by Vexatos** [comment]
Furthermore, it is much more easy to work with as you can do things like using the `bit32` API to tweak your colours.

**Comment by Cruor** [comment]
Forum post [here](http://www.computercraft.info/forums2/index.php?/topic/28455-cc-175-hxcperipherals-rgb-lights/page__p__265784#entry265784) for a lamp mod had the majority of the users wanting 0-255 instead of 0-1. Though it is a small sample size I would assume they would react the same to this.

**Comment by KnightMiner** [comment]
From a user standpoint, 255 transfers over a lot better from programs like Microsoft paint, so if they see a color they like it is simply copying color values into the program rather than needing to do additional math to make it a 0-1 color

**Comment by oddstr13** [comment]
I find this unintuitive, as I practically never see colors specified as floats. But then again, I havn't done much GUI work, and I've done no 3D rendering.

What I am most used to, and my guess the main target audience of the mod (if they have any knowledge of numbers representing colors), would be #rrggbb, then 0-255 (r, g, b).
0xrrggbb makes sense, as it is easy to enter, and easy to work with (bit shifting and such).

Ease of use should be top priority, not how the underlying graphics framework works. We are abstracting away all other things of that.

**Comment by Lignum** [comment]
> Furthermore, it is much more easy to work with as you can do things like using the bit32 API to tweak your colours.

Sure, and you can work with 0-1 colours more easily because you can multiply them together among other cool stuff. We could go on like this forever, which is why I've just added `colours.rgb8`:

```Lua
term.setColour(colours.white, colours.rgb8(255, 0, 255))
-- or
term.setColour(colours.white, colours.rgb8(0xFF00FF))
```

Which should hopefully settle the debate.

Vexatos:
This does not make it easier. `setColor` still uses 0-1, so it is still the most unintuitive possible way.

oddstr13:
If you are allready doing advanced things like multiplying colors, you probably wouldn't mind converting that to 0-255 or 0xffffff. Make it simple for the noobs, and do something like `term.setColor(colors.white, 0xffffff)`.

Lignum:
> This does not make it easier. setColor still uses 0-1, so it is still the most unintuitive possible way.

Yeah... that's why there's now a utility function that you can use to work around that.

> Make it simple for the noobs, and do something like term.setColor(colors.white, 0xffffff).

Well actually, I could make an overload for `setColour` taking hex RGB8. Though a 4-parameter `term.setColour(c, r8, g8, b8)` is evidently already reserved.

Vexatos:
> to work around that
We want a good function, not a bad function with a workaround.

oddstr13:
Paint.net color picker

MaoamZedong:
I'm with Lignum on this one. 0-1 colours are much more standard in programming, and way easier to work with for anything involving graphics, e.g. blending. You guys need to get out of your image editors and learn some real graphics programming.

I think colours.rgb8 was a good way to settle it, though. I'm not sure why you guys are still arguing? Do you even read what people say?

Vexatos:
You are not the only one who has been programming for years. The only place I have seen 0-1 be used is OpenGL, and I can recall at least a dozen places ranging from graphics software to other mods to online tools using one of the other ways, simply because they are much more intuitive.

lurquhar1221:
I'm with Lignum on this one, using 0-1 is much more logical, and he is providing `colors.rgb8` in order to satisfy those who want to use the latter.

Vexatos:
You can not easily copy colour values from other software, you cannot easily use `bit32` for colour modifications, how is it supposed to be better?

MaoamZedong:
There
is
`colour.rgb8()`

justync7:
>We want a good function, not a bad function with a workaround.

I'm sorry to say this but, you're crying over spilled milk.
`colours.rgb8` satisfies **precisely** what you want.

>The only place I have seen 0-1 be used is OpenGL

Exactly, Minecraft uses OpenGL so @Lignum is following that standard.

It's also worth keeping in mind that Dan specifically asked for it to use 0-1 and he is the one that was kind enough to open source his mod for us.

Vexatos:
@MaoamZedong ...which still is nothing more than a workaround, i.e. a function to make a bad function appear less bad.

MaoamZedong:
wow its true

Vexatos:
I don't see how exposing the details of OpenGL to Lua would be different.

lurquhar1221:
Let's move this conversation to the ComputerCraft IRC (#computercraft, irc.esper.net) in order to not debate over GitHub.

oddstr13:
> It does not add unneccessary complexity for users of the mod, **and maintains the accessibility for which the mod is known.**
[README.md](/dan200/ComputerCraft/blob/master/README.md)

justync7:
@oddstr13 Exactly. It's known for accessibility that Dan wants, not what you want.

Vexatos:
dan isn't always right, you know...

Lignum:
Alright, you can now do:
```Lua
term.setColour(colours.white, 0x00FF00)
-- and also
term.setColour({
  [colours.white] = 0x00FF00,
  [colours.black] = { 1.0, 0.0, 1.0 }
})
```

MaoamZedong:
thanks for the work lignum! that seems to be the closest we can get that makes everybody happy :)

src/main/java/dan200/computercraft/client/gui/GuiPrintout.java
```diff
@@ -204,7 +205,7 @@ public void drawScreen(int mouseX, int mouseY, float f)
             int lineIdx = ItemPrintout.LINES_PER_PAGE * m_page + line;
             if( lineIdx >= 0 && lineIdx < m_text.length )
             {
-                fontRenderer.drawString( m_text[lineIdx], x, y, m_colours[lineIdx], null, 0, 0, false );
+                fontRenderer.drawString( m_text[lineIdx], x, y, m_colours[lineIdx], null, 0, 0, false, new Palette() );
```
dan200: Instead of creating this every frame, could we cache it? Just store a static Palette.DEFAULT member with the default colour values

src/main/java/dan200/computercraft/core/apis/TermAPI.java
```diff
@@ -67,13 +73,17 @@ public void shutdown( )
             "getTextColor",
             "getBackgroundColour",
             "getBackgroundColor",
-            "blit"
+            "blit",
+            "setColour",
```
dan200: I've had a rethink of these method names. I think to avoid solution with text/background colour we should rename them setPaletteColo(u)r

src/main/java/dan200/computercraft/core/apis/TermAPI.java
```diff
@@ -101,6 +111,59 @@ public static int parseColour( Object[] args, boolean _enableColours ) throws Lu
...
+    public static void setColour( Terminal terminal, int colour, float r, float g, float b )
+    {
+        if( terminal.getPalette() != null )
+        {
+            terminal.getPalette().setColour( colour, r, g, b );
+            terminal.setChanged();
+        }
+    }
+
+    public static void setColour( Terminal terminal, HashMap<Object, Object> colours) throws LuaException
```
dan200: I'm not convinced we need/want the ability to pass in a table. Calling the method 16 times for a complete palette swap should be fine.

src/main/java/dan200/computercraft/core/apis/TermAPI.java
```diff
@@ -270,6 +333,65 @@ public static int parseColour( Object[] args, boolean _enableColours ) throws Lu
                 }
                 return null;
             }
+            case 19:
+            case 20:
+            {
+                // setColour/setColor
+                if( !m_environment.isColour() )
```
dan200: I think we should handle greyscale terminals as follows:
Still allow you set the palette colour on any type of terminal, but just convert the colours to greyscale (by setting each component to (r+g+b)/3) during rendering. This also means we can allow any colour to be set during setTextColour/BackgroundColour

src/main/java/dan200/computercraft/core/apis/TermAPI.java
```diff
@@ -270,6 +333,65 @@ public static int parseColour( Object[] args, boolean _enableColours ) throws Lu
...
+                {
+                    // Make sure you can't circumvent greyscale terminals with this function.
+                    throw new LuaException( "Colour not supported" );
+                }
+
+                if(args.length >= 1 && args[0] instanceof HashMap)
```
dan200: Don't think this is needed, see above

src/main/java/dan200/computercraft/core/apis/TermAPI.java
```diff
@@ -270,6 +333,65 @@ public static int parseColour( Object[] args, boolean _enableColours ) throws Lu
...
+                {
+                    @SuppressWarnings( { "unchecked" } ) // There isn't really a nice way around this :(
+                    HashMap<Object, Object> colourTbl = (HashMap<Object, Object>)args[0];
+                    setColour( m_terminal, colourTbl );
+                    return null;
+                }
+
+                if(args.length == 2 && args[0] instanceof Double && args[1] instanceof Double)
```
dan200: I like this method of supporting both forms of colour setting here :)

src/main/java/dan200/computercraft/shared/peripheral/monitor/MonitorPeripheral.java
```diff
@@ -52,7 +56,11 @@ public String getType()
             "getTextColor",
             "getBackgroundColour",
             "getBackgroundColor",
-            "blit"
+            "blit",
+            "setColour",
```
dan200: See my comments in TermAPI, same should apply here.

src/main/resources/assets/computercraft/lua/rom/apis/colors
```diff
@@ -25,13 +25,23 @@ function combine( ... )
...
+        r = bit32.band(r, bit32.bnot(c))
+    end
+    return r
 end
 
 function test( colors, color )
     return ((bit32.band(colors, color)) == color)
 end
+
+function rgb8( r, g, b )
```
dan200: Confused what this function does. Shouldn't it be encoding three r,g,b floats to a 24 bit int, or vice versa? It seems to do neither.

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -276,6 +286,33 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
         setTextColor( color )
     end
 
+    function window.setColour( colour, r, g, b )
+        if type(colour) == "table" then
```
dan200: If table overload is removed in Term/Monitor, do the same here.

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -100,6 +106,10 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
         end
     end
 
+    local function updatePalette()
+        return parent.setColour( tPalette )
```
dan200: A 1-16 loop here is fine.

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -276,6 +286,33 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
...
+            tPalette[ colour ] = { colours.rgb8( r ) }
+        elseif type(colour) == "number" and type(r) == "number" and type(g) == "number" and type(b) == "number" then
+            local tCol = tPalette[ colour ]
+            tCol[1] = r
+            tCol[2] = g
+            tCol[3] = b
+        end
+
+        if bVisible then
+            return updatePalette()
```
dan200: Updating the whole palette is unnecessary here, just update the one colour that changed

dan200:
Just done a code review of this PR. If the requested changes are made i'll merge it in :)

src/main/java/dan200/computercraft/client/gui/GuiPrintout.java
Lignum: Hmm, the JVM is pretty good with creating lots of objects nowadays, but it can't hurt, I guess.

src/main/resources/assets/computercraft/lua/rom/apis/colors
Lignum: It does the following:
```Lua
colours.rgb8(255, 0, 255) -- -> 1.0, 0.0, 1.0
colours.rgb8(0x00FFFF) -- -> 0.0, 1.0, 1.0
```
I guess the latter overload isn't needed anymore since you can pass hex rgb into `term.setColour` directly now.

src/main/java/dan200/computercraft/core/apis/TermAPI.java
Lignum: To quote @BombBloke:

> [...] Otherwise, trying to change all 16 colours in a window (and how often will people change one colour at a time?) requires 16 pairs loops, and hence 256 calls to the parent terminal's setColour function, and if that parent has a parent, it only gets worse from there.

I agree that the overload adds clutter, so I think we should benchmark this to determine if it's really needed.

src/main/java/dan200/computercraft/core/apis/TermAPI.java
Lignum: I like it, will do!

src/main/resources/assets/computercraft/lua/rom/apis/colors
dan200: Surely the more useful things to do are:
colours.rgb8(0x00ffff) -> 0.0, 1.0, 1.0
colours.rgb8(0.0, 1.0, 1.0) -> 0x00ffff

src/main/java/dan200/computercraft/core/apis/TermAPI.java
dan200: Setting 16 colours only required 256 writes because of the redundant setting of all 16 colours on each setColour call in window. Now that it only sets one, it's back to being a linear relationship so should be fine.

Lignum: Sure, but `window` isn't the only thing in the universe that uses `setPaletteColour`. But alright, I'll get rid of it, we can always add it back if needed, after all.

src/main/resources/assets/computercraft/lua/rom/apis/colors
Vexatos: For some reason, @Lignum decided that `setPaletteColour` could take either a 24-bit integer or three numbers between 0 and 1; this function is supposed to turn three 8-bit integers into those three numbers between 0 and 1. It's really confusing by now. Why not just stick with 24-bit RGB?

Lignum: It's intended to be plugged into `term.setPaletteColour` so you can specify each channel as an 8-bit number:
```Lua
term.setPaletteColour(colours.white, colours.rgb8(255, 0, 255))
```
I added it as a compromise for the  0-1 range vs. 0-255 range debate.

Just to clarify, the current possibilities are:
```Lua
term.setPaletteColour(colours.white, 0.0, 1.0, 1.0)

-- I can't make an overload for this, hence the addition of colours.rgb8
term.setPaletteColour(colours.white, colours.rgb8(0, 255, 255)) 

term.setPaletteColour(colours.white, 0x00FFFF)
```

dan200:
That usage doesn't get you much, seeing as you can just do
term.setPaletteColour(colours.red, r/255, g/255, b/255)

Lignum:
It had its purpose before the 24-bit int overload was added, so yeah, it has kind of lost its purpose other than making that conversion a bit nicer looking (imo, at least). So, ultimately, it's your mod, keep it or kill it?

oddstr13:
I would like to suggest using `term.setPaletteColour(colours.red, r*255, g*255, b*255)` instead, if you really want/need the 0-1 range when working with colors. I imagine there will be a lot of confusion if you have to do `term.setPaletteColour(colours.red, r/255, g/255, b/255)` in order to enter colour as 0-255.

Vexatos:
Yea, a single 24-bit number or three 8-bit numbers would be less confusing than a single 24-bit number vs. a decimal number.

Lignum:
You guys are quite persistent :stuck_out_tongue:.

I've reconciled both colour formats as much as possible so that everyone's happy, meaning there's basically nothing I can do anymore without displeasing the other side. It's no longer a rational issue, it's a social issue. So at this point I'm going to hold off and go with whatever gets this merged.

asiekierka:
Please note that 0-1 range colors are not just closer to percentages, they're also more futureproof in case Minecraft ever gets 10-bit display support (ha, as if!).

I believe colours.rgb8 is a good idea, but it's *rgb8* - it should accept 0-255 values. 8-bit values. Consistently. If we want to convert away from 8-bit values back into 0-1 range values, a new method should be added.

PR #201, opened 2017-05-06
[WIP] [Peripheral] Speaker

Restioson:
# Speaker API
This PR adds a new peripheral: the speaker peripheral. This allows computers to play any sound! Theoretically, you can even play sounds from other mods, provided that you use the correct domain. However, as of writing, `speaker.play` throws a NPE due to `ItemPocketComputer`'s `ServerComputer.getPos()` returning `null`. This is fixed in #172 .

# Possible Use Cases
- Notification API! You can notify the user with a snazzy ping. Useful as currently there is no way of notifying players of something on a computer
- Music! You can play records
- Custom music! You can make music using this by combining different sounds
- Super sneaky wireless range pinpointer! You can play a tone on a wireless pocket computer every time it gets a packet according to the range of the other modem. You could use this for PVP. I created a working implementation of this. [Click here to see the video](https://streamable.com/nsms8)!

# Example
```lua
speaker.playsound("minecraft:entity.cow.ambient") -- Plays entity.cow.ambient from domain minecraft
sleep(1 / 20) -- Sleeps the required amount of time, converting from ticks to seconds
speaker.playsound("entity.cow.ambient", 0.1) -- Plays entity.cow.ambient (in default domain, minecraft) at volume 0.1
sleep(1 / 20)
speaker.playsound("entity.cow.ambient", 1, 0.1) -- Plays entity.cow.ambient at volume 1 and pitch 0.1, thereby slowing it down
sleep(1 / 20)
speaker.playnote("harp",1,2) -- Plays the noteblock harp note at volume 1 and pitch 2
```

# Functions
 - `speaker.playsound` - Takes 1 positional arguments (`string` resource name) and 2 optional positional arguments (`double` volume and `double` pitch). Returns `true` or `false` based on whether the `resourcename` exists and is a sound. Plays the sound at `resourcename` with `volume` and `pitch`, under category `RECORDS`.

# Summation of changes
 - Added `speaker_enable` config option
 - Added `speaker` peripheral & provider
 - Added `speaker.playsound(resourceLocator, volume, pitch)` function in java
- Added `speaker.playnote(noteblockInstrument, volume, pitch)` function in java

# WIP
Rewrite as peripheral.

# Thanks
Special thanks to @SquidDev for helping me understand and navigate the codebase, and @gegy1000 for helping me out with Forge.


src/main/java/dan200/computercraft/ComputerCraft.java
```diff
@@ -277,6 +281,9 @@ public void preInit( FMLPreInitializationEvent event )
         Config.turtlesCanPush = Config.config.get( Configuration.CATEGORY_GENERAL, "turtlesCanPush", turtlesCanPush );
         Config.turtlesCanPush.setComment( "If set to true, Turtles will push entities out of the way instead of stopping if there is space to do so" );
 
+        Config.minTimeBetweenSounds = Config.config.get( Configuration.CATEGORY_GENERAL, "minTimeBetweenSounds", minTimeBetweenSounds);
+        Config.minTimeBetweenSounds.setComment("The minimum time in between calls to sound.play on one computer in milliseconds" );
```
SquidDev: You'll need to add a language key to `en_US.lang`. Something like `gui.computercraft:config.min_time_between_sounds=Minimum time between sounds`. Make sure it fits on the screen though. It might also be a good idea to add a `.setMinValue(0)` so people cannot use negatives.

src/main/java/dan200/computercraft/core/apis/SoundAPI.java
```diff
@@ -0,0 +1,141 @@
...
+                else if (arguments.length == 3)
+                {
+                    if (!(arguments[2] instanceof Double)) // Arg wrong type
+                        throw new LuaException("Expected string, number (optional), number (optional)");
+                    pitch = ((Double) arguments[2]).floatValue();
+                }
+
+                ResourceLocation resourceName = new ResourceLocation((String) arguments[0]);
+
+                if (System.currentTimeMillis() - this.lastPlayTime > ComputerCraft.Config.minTimeBetweenSounds.getDouble())
```
SquidDev: It would be better to keep an instance `clock` variable, incrementing it each time `advance` is called. This way it'll keep in sync with `os.clock()` and won't cause issues in the event of server lag. See `OSAPI` for how it's done there.

src/main/java/dan200/computercraft/core/apis/SoundAPI.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public Object[] callMethod(ILuaContext context, int method, Object[] arguments) throws LuaException, InterruptedException
+    {
+        switch (method) {
+
+            // play
+            case 0:
+
```
SquidDev: Wrap each case statement in braces, so variables are not shared between them. It isn't strictly needed now, but if additional methods are added you'll want them.

src/main/java/dan200/computercraft/shared/computer/core/ServerComputer.java
```diff
@@ -56,6 +57,11 @@ public ServerComputer( World world, int computerID, String label, int instanceID
 
         m_changedLastFrame = false;
         m_ticksSincePing = 0;
+
```
SquidDev: There should probably be a config option to disable the sound API.

src/main/java/dan200/computercraft/core/apis/SoundAPI.java
```diff
@@ -0,0 +1,141 @@
...
+import net.minecraft.util.SoundCategory;
+import net.minecraft.util.SoundEvent;
+
+public class SoundAPI implements ILuaAPI
+{
+
+    /**
+     * Sound API for ComputerCraft. Provides an interface to Forge's API.
+     * Possible use cases are notification apis, alert sounds, or playing music
+     */
```
SquidDev: Javadoc should go at the beginning of the class definition, not inside it.

src/main/java/dan200/computercraft/core/apis/SoundAPI.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private String[] methods;
+    private ServerComputer computer;
+    private long lastPlayTime;
```
SquidDev: Dan generally prefixes fields with `m_`. I don't know how strict he is on this convention, but it might be consistent with the rest of the code.

src/main/java/dan200/computercraft/core/apis/SoundAPI.java
```diff
@@ -0,0 +1,141 @@
...
+        switch (method) {
+
+            // play
+            case 0:
+
+                float volume = 1f;
+                float pitch = 1f;
+
+                // Check if arguments are correct
+                if (arguments.length == 0) // Too few args
```
SquidDev: This is just me being petty now, but braces on all the if statements! Unless the condition and body are all on the same line it is, IMO, better to have explicit braces.

src/main/java/dan200/computercraft/core/apis/SoundAPI.java
Restioson: I'm happy to stop using `this.member` in favour of `m_member`.

src/main/java/dan200/computercraft/core/apis/SoundAPI.java
Restioson: Whoops, haha! In Python it is the opposite.

src/main/java/dan200/computercraft/ComputerCraft.java
Restioson: I had added the `setMinValue`. I don't understand the language key: won't this snippet do it - 
```java
for (Property property : Config.config.getCategory( Configuration.CATEGORY_GENERAL ).getOrderedValues())
        {
            property.setLanguageKey( "gui.computercraft:config." + CaseFormat.LOWER_CAMEL.to( CaseFormat.LOWER_UNDERSCORE, property.getName() ) );
        }
```

src/main/java/dan200/computercraft/core/apis/SoundAPI.java
Restioson: ah, ok. To me it just looked cleaner, but it's just peanuts to me

src/main/java/dan200/computercraft/ComputerCraft.java
SquidDev: That will register the language key, but you haven't actually provided a translation - if you open the GUI, you'll just see `minTimeBetweenSounds`. If you add the appropriate string to `en_US.lang`, then it should be translated correctly.

Restioson:
I've added your suggestions @SquidDev and I fixed the pitch issue. Do you have any other suggestions or shall I remove the WIP tag?

Lignum:
Hmm, I don't think this should be a default API, but a peripheral instead. This way server admins can control specifically who can play sound and who can't, because sounds can be incredibly useful, but also incredibly annoying, even when there's a timeout.

Restioson:
@Lignum I'll look into that.

Vexatos:
I don't think this belongs into CC itself; looks like something a peripheral in an addon should do.

Restioson:
@Vexatos why? If this were a peripheral, what would your thoughts be? It is pointedly not a peripheral for use with pocket computers.

SquidDev:
> It is pointedly not a peripheral for use with pocket computers.

Well #172 adds pocket computer peripheral support, so that is less of an issue. If we expand that to allow multiple upgrades, then you will be able to have a noteblock and modem on the same pocket computer.

Restioson:
@SquidDev would it be called a noteblock? It could be a speaker, perhaps.

Restioson:
@Lignum do you mean a physical peripheral?

Restioson:
I wonder if I could do a permission for sound.

SquidDev:
> Would it be called a noteblock? It could be a speaker, perhaps.

It doesn't really matter, that was an example. I think I've generally seen this method provided via iron noteblocks (such as Peripherals++ or Plethora). Though let's be honest [Comptronics's tape drives](http://wiki.vex.tty.sh/wiki:computronics:tape) are way cooler 😄.

Restioson:
Woah, cool! This I feel encourages even more innovation, as you need to invent your *own* format. *Theirs are still cooler though*

Vexatos:
For what it's worth, Computronics also has yet another iron noteblock...

Restioson:
This is more flexible though, allowing *any* sound to be played.

Vexatos:
Then make an addon for it.

Restioson:
@Vexatos if it gets merged, then I don't need to.

Vexatos:
It should not get merged. This is not something that belongs directly into CC.

Restioson:
So, I made a wireless range pinpointer with a turtle, pocket computer, and computer. I'll give a link to a video of it soon.

Cruor:
This is excessive for the base version of Computercraft in my opinion. A simple os.beep(hz) would be more than enough and more fitting for the mod. If you need finer control over sound you could just use a command computer, they were added for a reason.

Restioson:
@Cruor once this is merged, you can limit the sounds to be allowed to play simply by adding a class for allowed `resourceLocator`. Also, only operators have access to command computers, which is very limiting. There is also no command pocket computer - one of the reasons I made this PR was to allow for notification apis, which would be very useful on pocket computers.

dan200:
This is cool functionality, but I would rather see this as a peripheral. A "Speaker" block would fit in with CC's Printer, Disk Drive, Monitor, etc very nicely. We could also add it as a turtle upgrade and a pocket computer upgrade once SquidDev's changes for that are in, we could call it a "Noisy" turtle.

dan200:
I would also hardcode "minTimeBetweenPlays" to 1 tick, and disable the setting and API to access it. This allows you to be very noisy, but not to overload the game.

dan200:
It would also be nice if there was a way to play the standard noteblock sounds for each of the instruments without having to know the resource names. I haven't thought about what this would look like.

SquidDev:
The "standard" iron noteblock API is something like:

 - `playNote(instrument:string|number, pitch:number[, volume:number])` Plays a note block note. The instruments are defined in `BlockNote.INSTRUMENTS` - "harp", "basedrum", "snare", "hat", "bass".
 - `playSound(sound:string[, pitch:number][, volume:number])` Pretty much the existing method.

Note that pitch in the first one uses Noteblock pitches, and so should be calculated as so:

```java
final float adjPitch = (float) Math.pow(2d, (double) (pitch - 12) / 12d);
```

You can see my peripheral mod's implementation [here](https://github.com/SquidDev-CC/plethora/blob/minecraft-1.10.2/src/main/java/org/squiddev/plethora/integration/vanilla/method/MethodsNoteblock.java) if you need a hint on getting started.

Vexatos:
Should definitely be a peripheral, and I still think it should stay in other addons as it is right now. This doesn't belong into CC.

dan200:
Those methods look good to me.

Restioson:
@dan200 i'll start working on the peripheral ASAP. Hopefully by then #172 is merged so I can test it with pocket computers, which is really the reason I thought of this API. 

On a side note, would you consider integrating a speaker peripheral into the pocket computers by default? It would be very useful for notifications, like on a smartphone.

apemanzilla:
@Restioson I don't see why it should be included by default, it would be easy enough to add it and it makes more sense too.

Restioson:
@apemanzilla yeah, it was just a thought.

Restioson:
Hmm, if anyone is good at texturing and wants to try their hand at a speaker peripheral, feel free. I might otherwise just try a retexture of the Wireless Modem with a retextured front.

dan200:
I think it should be a full block, with common side textures and a distinct
front like the disk drive/printer/etc. If you commit placeholders I'm sure
someone can spruce them up

Restioson:
@dan200 ah, ok. If it were a full block, would turtles not be able to have them? Visually it would look odd. Will pocket computers be able to have them as per #172 ?

dan200:
The turtle version would still necessarily look like a flat panel

Restioson:
Could that be a retextured modem as I described earlier?

dan200:
Sure, look at how the crafty upgrade works

Vexatos:
You probably shouldn't have tried merging the current master branch into this PR...
grafik

SquidDev:
It might be neater to close this PR and re-open a new one with the relevant files copied across - a lot has changed since the original commit and it'll remove the number of merge commits. 

I'm going to describe the process I've gone to get my repo set up. There are probably nicer ways, but this works for me. You've probably done several of these steps already, so skip the ones you've already got:

 - `git clone https://github.com/Restioson/ComputerCraft.git`: Pretty obvious, clone your repo.
 - `git remote add dan200 https://github.com/dan200/ComputerCraft.git` Add `dan200/ComputerCraft` as a remote.

Then assume you create a new branch `apiSoundAPI` and add some commits. Then this repo's `master` branch gets some commits, introducing merge conflicts. Then you can do:

 - `git remote update` Download the latest commits from all all remotes.
 - `git rebase dan200/master` Rebase this branch onto this repo's master branch. This basically "recreates" all your commits, but using the new master branch as a starting point. You may need to resolve some merge conflicts, but Git guides you relatively well through that.
 - `git push -f` Push your commits, overwriting the ones already there.
 - ???
 - Profit! And a cleaner git history.

Restioson:
@Vexatos yeah, that was unintentional.

@SquidDev are you suggesting that I delete my ComputerCraft repo and then reclone it?

dan200:
Yeah.. this PR is messy now. A peripheral seems a distinct enough thing from an API to warrant a new one anyway.

Restioson:
I'm closing this PR

Wojbie:
> On a side note, would you consider integrating a speaker peripheral into the pocket computers by default? It would be very useful for notifications, like on a smartphone.

Ok just throwing a suggestion here but How about splitting the features.
Have all computers (pocket included) have access to basic os.beep(hz) (Like that beeper that is installed on most Motherboards in most computers). Basic beep only, Limited hz range and lasts exactly a second.

Then if someone needs better or more sounds they plug in peripheral Speaker that allows them to 
.playNote and .playSound like in that comment by SquidDev.

What you thing about this idea @Restioson. That way we get to have a bird (basic sounds in computers and pockets) and eat a bird (All the sounds in speaker).

Restioson:
@Wojbie sounds good. What do you think @dan200 ?

Vexatos:
OpenComputers uses a separate class for synthesizing square waves of a specific frequency manually for its beeps. Should the "motherboard beeps" in CC be square too?

Vexatos:
OpenComputers uses a separate class for synthesizing square waves of a specific frequency manually for its beeps. Should the "motherboard beeps" in CC be square too?

Also, wasn't there something about CC (by default) not exposing anything from the "outside world"? Wouldn't Minecraft sounds count as that?

Restioson:
@Vexatos I think it's more "Minecraft"ey to use noteblock sounds like @SquidDev suggested. 

About CC not exposing anything - the sounds are already *in* the game. The sounds could be recorded, lore-wise (can CC have lore? 😛)

Vexatos:
They are in the game, just not by name.

Restioson:
I don't really follow? Do you mean that as a player you cannot play them?

Vexatos:
I mean that the names do not appear in-game.

Restioson:
@Vexatos i see what you mean.

Issue #145, opened 2016-10-13
Inefficient atlas texture space usage


sfPlayer1:
CC adds several pre-stitched textures with lots of dead space to the MC texture atlas. A large atlas texture affects overall game performance severely or may even prevent the game from starting if the graphics driver's size limit is exceeded.

I have uploaded a partial atlas dump showing the problem at https://i.imgur.com/HUWv94n.png

dan200:
Much improved in current. I don't consider this an issue anymore

PR #203, opened 2017-05-06
Convert turtle rendering to use tinting

SquidDev:
This uses Minecraft's colour tinting system in order to change the colour of turtle models. This removes the need to have 16 models and textures for each colour, reducing texture atlas space and hopefully memory consumption.

This should help with #145, though it is possible that the texture size could be reduced a little more. Maybe something to play with in the future,


SquidDev:
On a side note, now that textures are no longer being used, turtles could be any RGB colour (like disks are). You wouldn't be able to use an dye to get a generic colour, it could still be set via `ITurtleAccess`. I don't know what your thoughts are on extending the API to allow this?

Restioson:
@SquidDev you could "interface" RGB colours by allowing the user to tint multiple times. I.E if you tinted it first blue, and then red, it would be purple.

SquidDev:
Note, while this does dramatically reduce the texture atlas space, there is still a bit of an issue. Using mezz's [Texture dump](https://github.com/mezz/TextureDump), you can see that turtles take up a significant proportion of the atlas.

textures_mipmap_0

While this is a probably a drop in the water for larger modpacks, it would still be nice to minimise this as much as possible. I've tried rather hard to compress it down to a 64x64 texture, but without leaving one side out, it doesn't appear possible. A couple of potential solutions:

 - Remove the normal and advanced textures and just use the tinted one, manually providing a colour.
 - Split the texture into individual files and let Minecraft's texture stitcher do the work. This is probably the nicest solution, but I want to check with you before going ahead.
 - Mirror one side of the turtle: the left and right textures are identical, and so one *could* be removed. However this may cause issues with turtle overlays and rather limits resource packs in what they can do.

Lignum:
Additionally, it's no longer necessary to stick to powers of two for texture size. Minecraft now requires OpenGL 2.1, which in turn requires the graphics driver to support non-power of two textures, meaning you should be able to safely resize the texture to fit its contents properly.

dan200:
At what version of MC did that change? When I added the ISO 8859-1 characters in CC 1.76 I had to make the texture an NPOT square because a few users couldn't run the game when it wasn't

SquidDev:
It still has to be a square - I've tested that already. I haven't tested it with non-power of two assets but I'd be surprised if the texture stitcher can cope. I think it would be better to keep it as a power of 2 of some form, simply because the model's UVs work between 0 and 16, which gets much more confusing when you're not dealing with factors of 16.

KnightMiner:
I recall seeing a Forge PR fixing some issues with non-power of two textures breaking mipmapping, no telling what other weirdness it causes hence I just avoid them

PR #205, opened 2017-05-06
Expose wireless network in the public API

SquidDev:
This PR adjusts how wireless and wired networks are represented internally, as well as exposing some methods to the public API.

## What
 - This adds `IPacketNetwork`, `IPacketSender`, `IPacketReceiver` and `Packet` to the `dan200.computercraft.api.network` package.
 - Adds `IPacketNetwork ComputerCraft.getWirelessNetwork()`.
 - Changes modem peripherals, wired modems, wireless modems and the wireless network to these interfaces instead of the original `ModemPeripheral`, `INetwork` and `IReceiver`.

## Changes made from the original
I've tried to make it a little bit more flexible so other peripheral mods can add their own interesting functionality. So here goes:

 - **`Packet` instead of individual arguments:** Instead of passing around several arguments, the packet specific information is bundled into a `Packet` class. This was done in `TileCable` already so isn't a big change, it just looks a little nicer and allows for adding additional info to the packet without too much hassle.
 - **`IPacketReceiver` is not bound to a channel:** Previously each modem would register an `IReceiver` for each channel it listened on. Now each modem registers one `IPacketReceiver` (itself) and receives all packets, only queuing those it is interested in. While this does mean you have to loop over every wireless modem instead of being able to filter them, the performance hit is negligible. In return you can implement network devices which are able to listen to all packets.
 - **The packet sender is required to be an `IPacketSender`:** This means you can get more information about the sender, such as the it's current position.

I'm happy to explain/expand on any changes made here, and give examples of how I could see them being useful in the future.

I've gone for the rather verbose `IPacketNetwork` to distinguish it from other network functionality we may wish to expose in the future, such as peripherals on the wired network.


Vexatos:
An example implementation can be found [here](https://github.com/Vexatos/OpenComputers/blob/cc-wireless/src/main/scala/li/cil/oc/integration/computercraft/ComputerCraftWirelessNetwork.scala).

SquidDev:
@dan200 I don't want to be that guy, but do you have any opinions on this PR? [Your recent tweets](https://twitter.com/DanTwoHundred/status/865228833569558528) have me slightly worried I'm going to be perpetually rebasing this :). I realise you've done a lot of work on the repo in the last couple of days, so don't want to cause any more hassle, just looking for some initial thoughts. Thanks!

dan200:
I'm generally interested in merging most of the PRs on the list, i'd let you know if I didn't like something :). Only thing holding up this one and your others is they're quite big and need a chunk of time to code review, and I've reviewed and merged 60 PRs in the last 2 weeks!

I'll take another look at this one now.

SquidDev:
Great! Sorry, I wasn't trying to push you into this, just making sure it wasn't dead in the water either. Just to say thank you for all the work you've done over the last couple of weeks - we really appreciate it :).

src/main/java/dan200/computercraft/api/network/Packet.java
```diff
@@ -0,0 +1,152 @@
...
+ * @see IPacketReceiver#receiveDifferentDimension(Packet)
+ * @see IPacketReceiver#receiveSameDimension(Packet, double)
+ */
+public class Packet
+{
+    private final int m_channel;
+    private final int m_replyChannel;
+    private final Object m_payload;
+
+    private final double m_range;
```
dan200: Should "range" and "interdimensional" be properties on the packet itself, or just additional arguments to transmit?

SquidDev: I decided to bundle it as a property as it describes something about the packet - in this case its behaviour.

dan200: Imo, range and interdimensional describe how a packet is sent, not what the packet actually is. ie: I sent packet "x" 10 metres. It's only a minor thing but would you mind making that one change?

dan200: Perhaps the following signatures would make sense, mirroring the receive functions:
transmitSameDimension(Packet p, double range);
transmiteInterdimensional(Packet p);

Which makes it clearer that packets sent interdimensionally are received by everything in every dimension regardless of range.

SquidDev: Yeah, that does seem much cleaner. Transmitting interdimesionally on a wired network is a bit of a weird concept, but it isn't something which was impossible beforehand either.

SquidDev:
Thank you! ❤️

Vexatos:
I wrote an [implementation using this new API](https://github.com/Vexatos/OpenComputers/blob/cc-wireless/src/main/scala/li/cil/oc/integration/computercraft/ComputerCraftWirelessNetwork.scala). This does exactly what I need. Thank you!

PR #206, opened 2017-05-06
Set CraftOS version to 1.8

Lignum:
Since this is release 1.80 being worked on, following the usual pattern, we should be at CraftOS 1.8. I've changed `os.version()`'s return value in `bios.lua`, if anything else needs to be adjusted for this, let me know and I'll change it.


Vexatos:
I hope the version will be increased to 1.81, actually, so that addons can properly depend on the version without having to deal with 1.80pr0 (which would compare as equal or higher in Forge's version system)

PR #208, opened 2017-05-07
Add SHA256 API

ObloxCC:
Since many Computercraft programs and GUIs use the SHA256 to store passwords, why not add it as an API that ships with computercraft? This would help limit the number of files being downloaded, and keep it from being downloaded multiple times onto the same CC computer.

*Note: GravityScore original ported this over to CC, I made a small change to ensure it would run correctly.

> In regards to copyright: both of these sources allow the free use of the code in them - they do not require you to include credits or licenses (to be honest, they aren't very clear about who exactly wrote them...) - GravityScore

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/sha256
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/sha256
@@ -1,0 +1,183 @@
+local MOD = 2^32
+local MODM = MOD-1
+
+local function memoize(f)
+	local mt = {}
+	local t = setmetatable({}, mt)
+	function mt:__index(k)
+		local v = f(k)
+		t[k] = v
+		return v
+	end
+	return t
+end
+
+local function make_bitop_uncached(t, m)
+	local function bitop(a, b)
+		local res,p = 0,1
+		while a ~= 0 and b ~= 0 do
+			local am, bm = a % m, b % m
+			res = res + t[am][bm] * p
+			a = (a - am) / m
+			b = (b - bm) / m
+			p = p*m
+		end
+		res = res + (a + b) * p
+		return res
+	end
+	return bitop
+end
+
+local function make_bitop(t)
+	local op1 = make_bitop_uncached(t,2^1)
+	local op2 = memoize(function(a) return memoize(function(b) return op1(a, b) end) end)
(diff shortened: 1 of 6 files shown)


Wojbie:
Hm. I am not sure if it's something that has to be added to CC core. It is something that can be achieved by adding the code to program itself or downloading api. It don't seem to click as something that fits to include  by default. It would make a nice treasure disk i think.

SquidDev:
This API also emulates the all bit functions, which shouldn't be needed seeing as CC provides an API to do the same.

Vexatos:
CC doesn't need any of this. If you added every "commonly used" program to CraftOS, there would be no end to the feature bloat.

ObloxCC:
Going off of adding this, I was thinking about adding a built in login (like Linux) that can be **enabled** to prohibit users from using/accessing the computer without logging in (includes blocking startup program on disk drives before password is entered)

Vexatos:
How does this belong into standard CC? This sounds like something for the forums.

ObloxCC:
> How does this belong into standard CC? This sounds like something for the forums.

Security that is included in almost all operating systems that should be included here. Also this can't be achieved without editing the ROM which doesn't go so well on servers where your information can easily be stolen, or deleted

Lignum:
> without editing the ROM which doesn't go so well on servers where your information can easily be stolen, or deleted

Servers could just edit the API anyway. And if they did, the ROM is still read-only, players would be able to see it.

Either way, I concur that this just adds unnecessary bloat to ComputerCraft, when you can just drop the API in yourself anyway. What I would be in support of, though, is to add a native `biginteger` API à la CCTweaks, that allows faster implementations of various cryptographic functions.

SquidDev:
I've always felt that CraftOS should attempt to do the bare minimum to provide a usable interface to the everyday user - a shell, some useful programs, and a couple of everyday APIs. While login scripts and hashing algorithms are fun and useful, they aren't by any means a necessity.

In fact, I think it is more interesting if people go and write their own. When I wrote my AES library, sure, I could have just used KillaVanilla's one, but it was more fun to experiment. If it had been built in, I wouldn't have even tried. Half the joy of the forums is seeing people create something new out of an old idea, and I don't think this enables that.

SquidDev:
I don't think you meant to do those last three commits on this branch :).

PR #209, opened 2017-05-07
Add a few programs

Wilma456:
I add a few small programs (nothing special)

size: Show the size of a file. Wrapper for fs.getSize()
whereis: Shows the full path of a shell programm eg.g whereis edit shows you /rom/progams/edit
webrun: Runs programs in the web. It's like pastebin run for other websites.
pwd: Prints the current directory
sleep: Wrapper for os.sleep()
lsper: Show all  attached peripherals
echo: Wrapper for print()

Nothing special, but might be useful for some user. These Programs are not complex, so I think it's the right for CraftOS.


ObloxCC:
pwd sounds like it should print a password or something

Lignum:
The name presumably comes from the unix command of the same name, which stands for "print working directory".

Wilma456:
pwd stand for **p**rint **w**orking **d**irectory and its a standard UNIX command

SquidDev:
Just a couple of thoughts:

 - `lsper` can be removed - we already have a `peripherals` command.
 - Instead of `webrun`, I think it might be nicer if `wget` took a flag or something. Like `wget -r <url>`.
 - `whereis` is often called `which` on Linux - I think they should probably be aliased.
 - I'm not really sure of the point of `pwd`, `sleep` or `echo` - you can't really write shell scripts, so they don't really have a purpose. In fact, they may be detrimental as people could use `shell.run("sleep 2")` instead of the sleep function - I've seen people do that for `clear`.
 - You're using `PascalCase` quite a lot. Dan generally uses `lowercase` or `pascalCase` (also Hungarian notation)

I'd also recommend downloading a Git client and using that instead of committing on GitHub - it's much easier and means you don't get several commits for the same changeset.

src/main/resources/assets/computercraft/lua/rom/programs/size
```diff
@@ -0,0 +1,27 @@
...
+  print("Usage: size <path>")
+  return 1
+end
+
+local filename = shell.resolve(Arg[1])
+
+if fs.exists(filename) == true then
+  if fs.isDir(filename) then
+    os.loadAPI("/usr/apis/wilmaapi")
+    local fileta = wilmaapi.listAllFiles(filename)
```
Wojbie: You are loading some wierd api that you did not include here.
Also why not use fs.list(string path) and not fs.isDir() to identify files instead of this Api..

On sidenote if size needs to calculate directory size then it probably should use recursion on all directories that are in scanned directory.

Wilma456: Thankyou for finding this. I had this program writen a few monts ago and had forgot, that this program use my API.

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,24 @@
+Arg = {...}
```
Lignum: This should be local.

src/main/resources/assets/computercraft/lua/rom/programs/sleep
```diff
@@ -0,0 +1,11 @@
+local Args = {...}
+if Args[1] == nil then
+print("Usage: sleep <time>")
```
Lignum: Please indent your code, it's more readable.

src/main/resources/assets/computercraft/lua/rom/programs/size
Restioson: Preferably load your API into the local scope. Even more preferably, don't use a non-CC api at all!

oddstr13:
`whereis` would return documentation, configuration and libraries as well. `which` returns the path of the executable that would be run.

```bash
$ whereis bash
bash: /usr/bin/bash.exe /etc/bash.bashrc /etc/bash.bash_logout /usr/include/bash /usr/share/man/man1/bash.1.gz
$ which bash
/usr/bin/bash
```

I'm not sure what the use of `pwd`, `sleep` and `echo` are, as we don't have shell scripting.

`size` could be handy, but I'd rather improve `ls` to include the `-lh` flags.

`which` is nice to have, but not absolutely essential on cc, as by default, only `.:/programs/:/programs/*/` are on the path by default.

Restioson:
I completely agree with @oddstr13 on this. Definitely add the flags to the `ls` command. However, a single standalone `size` could be useful to check the size of a single file/directory.

src/main/resources/assets/computercraft/lua/rom/help/lsper
```diff
@@ -0,0 +1 @@
+lsper list you all attached peripherals
```
Lignum: Write help files in the passive voice, that applies to the others as well. This just sounds really odd.

src/main/resources/assets/computercraft/lua/rom/help/size
```diff
@@ -0,0 +1,3 @@
+size shows you the size of a file
+
+Usage: size <file>
```
Lignum: Other help files say "ex:", not "Usage:".

oddstr13:
`size` as implemented here, reminds me a little of the `du` command, or at least the way I use it the most; `du -hcs`.

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
Restioson: Preferably indent your code with four spaces

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,24 @@
...
+
+if not Arg[1] then
+  print("Usage: webrun <url> <args>")
+  return
+end
+
+if not http then
+print("Please enable http in the config")
+return
+end
```
Restioson: Preferably indent your code with four spaces per level of indentation.

src/main/resources/assets/computercraft/lua/rom/help/size
Wilma456: edit and painzt say Usage. I will add a few comments.

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,24 @@
...
+
+local test,text = http.checkURL(Arg[1])
```
Restioson: Just stylistic, but please add whitespaces between the commas (`local test,text` -> `local test, text`)

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
Restioson: All variable names should start with a lowercase (`Arg` -> `arg`). Also, there could be more than one argument, so it could be renamed `args`.

src/main/resources/assets/computercraft/lua/rom/programs/size
```diff
@@ -0,0 +1,27 @@
...
+    local size = 0
+    for _,sizefile in ipairs(fileta) do
```
Restioson: Although you are not using the `key` return of `ipairs`, `_` is a bad name for a variable. Just name it `key` and don't use it.

src/main/resources/assets/computercraft/lua/rom/programs/size
```diff
@@ -0,0 +1,27 @@
...
+      size = size + fs.getSize(sizefile)
+    end
+    print("Size: "..size.." Bytes")
```
Restioson: Whitespace between `"`s and `..`s

src/main/resources/assets/computercraft/lua/rom/programs/size
```diff
@@ -0,0 +1,27 @@
...
+  else
+  print("Size: "..fs.getSize(filename).." Bytes")
+  end
+else
+  print("File does not exists")
```
Restioson: Should be `"File does not exist"`

src/main/resources/assets/computercraft/lua/rom/programs/sleep
```diff
@@ -0,0 +1,11 @@
+local Args = {...}
```
Restioson: Again, lowercase letter

src/main/resources/assets/computercraft/lua/rom/programs/whereis
```diff
@@ -0,0 +1,11 @@
+local Arg = {...}
```
Restioson: Lowercase!

src/main/resources/assets/computercraft/lua/rom/help/webrun
```diff
@@ -0,0 +1,3 @@
+With webrun you can run programs without download them
```
Restioson: This is structured differently to the other help pages

src/main/resources/assets/computercraft/lua/rom/programs/size
KnightMiner: Generally I've seen "_" used for a throwaway variable, generally not considered bad style

dan200:
> size: Show the size of a file. Wrapper for fs.getSize()
What's the usecase for this?

> sleep: Wrapper for os.sleep()
This was in an early version and was removed because it was pointless as we don't have shell scripts

> lsper: Show all attached peripherals
"peripherals" already does this

> echo: Wrapper for print()
See "sleep"

> whereis: Shows the full path of a shell programm eg.g whereis edit shows you /rom/progams/edit
> webrun: Runs programs in the web. It's like pastebin run for other websites.
> pwd: Prints the current directory
These are somewhat useful, maybe. I'd probably only merge pwd though.

dan200:
Edit: pwd isn't useful either: The current dir is always displayed on the command line

apemanzilla:
`size` is somewhat useful if you just want to check how much disk space is being eaten by something.

PR #210, opened 2017-05-07
Add 3D Computers*

ObloxCC:
*This changes the computer model to be semi 3D to make the computers look better and look better (ever so slight 3D effect).

_Pics:
http://imgur.com/7efDh5p
http://imgur.com/lTM0R7g
http://imgur.com/2GrYQR8_

EDIT: sorry about the added SHA256 and removed it in the commits, I am still new to GitHub.


SquidDev:
I think this could be really neat as a resource pack, but I'm not too sure about it being bundled with CC, it should at least going to be behind a config options.

I feel monitors should also have that same "embedded" look, but I'm not too sure how that would work with the monitor TESR though, perhaps something to discuss.

I'd also recommend downloading a Git client and using that instead of committing on GitHub - it's much easier and means you don't get several commits for the same changeset

ObloxCC:
> I think this could be really neat as a resource pack, but I'm not too sure about it being bundled with CC, it should at least going to be behind a config options.

I feel monitors should also have that same "embedded" look, but I'm not too sure how that would work with the monitor TESR though, perhaps something to discuss.

I'd also recommend downloading a Git client and using that instead of committing on GitHub - it's much easier and means you don't get several commits for the same changeset.

I'll attempt to give monitors the 3D look too, it might required multiple models though

Inari-Whitebear:
A commit message like "Make <file> more 3D" or something of the sort would be also more useful than simply "Update"

Vexatos:
This pull request is so messed up. Seriously.

Also, this looks exactly like something that belongs into a resource pack.

oddstr13:
I would also suggest using feature branches, instead of making pull-requests against your master branch.
`pr-3d-texture` and `pr-sha256` would be a lot better than both being in `master`. This allows for multiple, independent pull-requests at the same time.

I also second the use of a proper git client. There are many to choose from; GitHub desktop, SourceTree, TortoiseGit, or the plain git commandline client - to mention a few.

I also agree with @Vexatos that this is probably best suited as a resource pack.

ObloxCC:
This will be temporary closed while I fix everything

dan200:
I'm not a fan of these models, sorry

PR #211, opened 2017-05-07
Make shell.resolveProgram pick up on *.lua files

Lignum:
This is a fairly simple pull request. Since many people prefer their programs with a .lua extension, with this change, these files will be picked up by `shell.resolveProgram`, without specifying the extension explicitly.

For example, assuming there is a `test.lua` file in `/`, running a lua prompt in `/`:
```Lua
> shell.resolveProgram("test")
test.lua
```
It will always prefer exact matches, so if there is a `test` file as well, it will choose that one instead.

The main purpose of this is to allow running .lua files without specifying the extension:
```
> test
Hello World!
```

Before merging: Should autocomplete handle this? Currently, it behaves like before and will only autocomplete the full filename.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/shell
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/shell
@@ -111,6 +111,16 @@ function shell.resolve( _sPath )
 	end
 end
 
+local function pathWithExtension( _sPath, _sExt )
+    local nLen = #sPath
+    local sEndChar = string.sub( _sPath, nLen, nLen )
+    -- Remove any trailing slashes so we can add an extension to the path safely
+    if sEndChar == "/" or sEndChar == "\\" then
+        _sPath = string.sub( _sPath, 1, nLen - 1 )
+    end
+    return _sPath .. "." .. _sExt
+end
+
 function shell.resolveProgram( _sCommand )
 	-- Substitute aliases firsts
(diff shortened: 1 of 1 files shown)


Restioson:
We *need* this. Already it is somewhat strange that programs *don't* have the `.lua` extension. I am for going as far as making *all* CC lua files having the `.lua` extension for clarity.

Lignum:
I would also agree, but we can't feasibly do that without breaking lots of programs that refer to the default programs by their absolute paths.

Lignum:
Also, this does not work with `os.loadAPI`. I could adjust the function to resolve APIs with `shell.resolveProgram`, if a shell is present. Though I fear that that would break stuff.

Cruor:
If `require` gets merged that won't be a problem, as it resolves properly. Though it should probably also work with `os.loadAPI`.

Vexatos:
Should be easy enough to fix `os.loadAPI`, honestly.

Wilma456:
I think this should be included

Restioson:
@Lignum true, but in a perfect world. If there were only some way to recommend to use `.lua` files...

dan200:
Sounds good! I'd also be perfectly ok with renaming all builtin programs to a ".lua" extension, and doing the same for APIs, startup scripts, and anything else we call.

PR #215, opened 2017-05-08
Add which and webrun

Wilma456:
Yesterday you wrote, that you only which and webrun want to merge. So here's the version who only contains which and webrun. 

Some people has wrote, I should edit wget to add the flag -r and not made it as single programm. I think, wget is a downloader, so wget -r called like "Download the programm and run it after downloading" and not "Run the program without downloading".

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/help/webrun
+++ b/src/main/resources/assets/computercraft/lua/rom/help/webrun
@@ -1,0 +1,5 @@
+webrun run a programm from the ineternet without downloading them
+
+ex:
+"webrun http://example.com/program.lua" will run the program from this url
+"webrun http://example.com/program.lua test" will run the program from this url with the argument "test"
--- a/src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
@@ -1,37 +1,40 @@
-local Arg = {...}
+--Made by Wilma456
+local tArgs = {...}
 
-if not Arg[1] then
+if not tArgs[1] then
   print("Usage: webrun <url> <args>")
   return 1
 end
 
 if not http then
-print("Please enable http in the config")
-return 2
-end
-
-local test,text = http.checkURL(Arg[1])
-if test == false then
-  print("[Error]"..text)
+  print("Please enable http in the config")
   return 3
 end
 
-local data = http.get(Arg[1])
+local test, text = http.checkURL(tArgs[1])
+if test == false then
+  print("[Error]"..text)
+  return 4
+end
+
+local data = http.get(tArgs[1])
 if not(data) then
   print("Error while getting the program")
-  return 4
+  return 5
 end
 
 local program = data.readAll()
 data.close()
-local run,err = load(program)
+local run, err = load(program)
 if not(run) then
-  term.setTextColour(colors.red)
+  local oldcolor = term.getTextColor()
+  term.setTextColor(colors.red)
   print(err)
-  return 5
+  term.setTextColor(oldcolor)
+  return 6
 end
 
-table.remove(Arg,1)
-run(table.unpack(Arg))
+table.remove(tArgs,1)
+run(table.unpack(tArgs))
 
 return 0
--- a/src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
@@ -1,25 +1,37 @@
---Made by Wilma456
-local args = {...}
+local Arg = {...}
 
-if not args[1] then
+if not Arg[1] then
   print("Usage: webrun <url> <args>")
-  return
+  return 1
 end
 
 if not http then
-  print("Please enable http in the config")
-  return
+print("Please enable http in the config")
+return 2
 end
 
-local test,text = http.checkURL(args[1])
+local test,text = http.checkURL(Arg[1])
 if test == false then
   print("[Error]"..text)
-  return
+  return 3
 end
 
-local data = http.get(args[1])
+local data = http.get(Arg[1])
+if not(data) then
+  print("Error while getting the program")
+  return 4
+end
+
 local program = data.readAll()
 data.close()
-local run = load(program)
-table.remove(args,1)
-run(table.unpack(args))
+local run,err = load(program)
+if not(run) then
+  term.setTextColour(colors.red)
+  print(err)
+  return 5
+end
+
+table.remove(Arg,1)
+run(table.unpack(Arg))
+
+return 0
--- a/src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
@@ -1,0 +1,25 @@
+--Made by Wilma456
+local args = {...}
+
+if not args[1] then
+  print("Usage: webrun <url> <args>")
+  return
+end
+
+if not http then
+  print("Please enable http in the config")
+  return
+end
+
+local test,text = http.checkURL(args[1])
+if test == false then
+  print("[Error]"..text)
+  return
+end
+
+local data = http.get(args[1])
+local program = data.readAll()
+data.close()
+local run = load(program)
+table.remove(args,1)
+run(table.unpack(args))
--- a/src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
@@ -13,13 +13,13 @@ end
 
 local test, text = http.checkURL(tArgs[1])
 if test == false then
-  print("[Error]"..text)
+  printError("[Error]"..text)
   return 4
 end
 
 local data = http.get(tArgs[1])
 if not(data) then
-  print("Error while getting the program")
+  printError("Error while getting the program")
   return 5
 end
 
@@ -27,10 +27,7 @@ local program = data.readAll()
 data.close()
 local run, err = load(program)
 if not(run) then
-  local oldcolor = term.getTextColor()
-  term.setTextColor(colors.red)
-  print(err)
-  term.setTextColor(oldcolor)
+  printError(err)
   return 6
 end
 
--- a/src/main/resources/assets/computercraft/lua/rom/programs/which
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/which
@@ -1,0 +1,12 @@
+--Made by Wilma456
+local Arg = {...}
+if Arg[1] == nil then
+	print("Usage: which <command>")
+	return
+end
+local path = shell.resolveProgram(Arg[1])
+if path == nil then
+	print("Can't find command "..Arg[1])
+	return
+end
+print(Arg[1].." is /"..path)
--- a/src/main/resources/assets/computercraft/lua/rom/programs/which
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/which
@@ -1,12 +1,13 @@
 --Made by Wilma456
-local Arg = {...}
-if Arg[1] == nil then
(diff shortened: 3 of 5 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,25 @@
...
+  return
+end
+
+if not http then
+  print("Please enable http in the config")
+  return
+end
+
+local test,text = http.checkURL(args[1])
+if test == false then
```
apemanzilla: Not super important, but you can just do `if not test then...` instead of `if test == false then...`

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,25 @@
...
+  print("[Error]"..text)
+  return
+end
+
+local data = http.get(args[1])
+local program = data.readAll()
```
apemanzilla: If `http.get` fails, `data` will be `nil` and the program will crash. It should check whether `http.get` succeeded and fail gracefully if not.

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,25 @@
...
+data.close()
+local run = load(program)
+table.remove(args,1)
+run(table.unpack(args))
```
apemanzilla: These two lines (24 and 25) can be replaced with simply `run(unpack(args, 2)`

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,25 @@
...
+local test,text = http.checkURL(args[1])
+if test == false then
+  print("[Error]"..text)
+  return
+end
+
+local data = http.get(args[1])
+local program = data.readAll()
+data.close()
+local run = load(program)
```
apemanzilla: The code source should be specified, either as the URL or the filename extracted from the URL. `load(program, source)`

Additionally, syntax errors will cause `run` to be `nil` and a second value containing the error message to be returned. This should be handled gracefully via `local run, err = load(...) if not run ... else ... end`

Wilma456:
The Bugs are now fixed

Restioson:
I think that it would be more beneficial if `wget` took a flag, such as `-r`/`--run` which would basically execute `webrun`'s functionality. I believe @SquidDev also mentioned this.

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -1,25 +1,37 @@
---Made by Wilma456
-local args = {...}
+local Arg = {...}
```
Restioson: `args` is a standard name for a variable like this. It should be changed back to `args`

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -1,25 +1,37 @@
...
 if not http then
-  print("Please enable http in the config")
-  return
+print("Please enable http in the config")
+return 2
 end
 
-local test,text = http.checkURL(args[1])
+local test,text = http.checkURL(Arg[1])
 if test == false then
```
Restioson: You indented this `if` for seemingly no reason

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -1,25 +1,37 @@
...
-  return
+  return 1
 end
 
 if not http then
-  print("Please enable http in the config")
-  return
+print("Please enable http in the config")
+return 2
 end
```
Restioson: Please indent the code in the `if` block with four spaces.

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -1,25 +1,37 @@
...
   print("Usage: webrun <url> <args>")
-  return
+  return 1
 end
 
 if not http then
-  print("Please enable http in the config")
-  return
+print("Please enable http in the config")
+return 2
```
Restioson: Why are you arbitrarily returning `2`? There is a *nix specification for return codes. `0` for success, and `1` for generic error. `2` means misuse of shell builtins, which is not the problem here. Please read [this](http://tldp.org/LDP/abs/html/exitcodes.html)

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -1,25 +1,37 @@
...
-local test,text = http.checkURL(args[1])
+local test,text = http.checkURL(Arg[1])
 if test == false then
   print("[Error]"..text)
-  return
+  return 3
+end
+
+local data = http.get(Arg[1])
+if not(data) then
```
Restioson: You do not need to put `data` in brackets

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -1,25 +1,37 @@
...
+  return 4
 end
 
-local data = http.get(args[1])
 local program = data.readAll()
 data.close()
-local run = load(program)
-table.remove(args,1)
-run(table.unpack(args))
+local run,err = load(program)
```
Restioson: Please use whitespace between the commas

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -1,25 +1,37 @@
...
+if not(run) then
+  term.setTextColour(colors.red)
+  print(err)
```
Restioson: ComputerCraft still prints errors in white, and what's more, you don't return textColour to what it was. Also, when coding for international use cases, please use *American English* (I am not American - it's just a standard)

Wilma456: ComputerCraft prints the syntax or other coding error in red, but I will change this to American English.

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
Wilma456: You say args is bad and I should change it to args?

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
Wilma456: Done

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,40 @@
...
+end
+
+if not http then
+  print("Please enable http in the config")
+  return 3
+end
+
+local test, text = http.checkURL(tArgs[1])
+if test == false then
+  print("[Error]"..text)
```
apemanzilla: Use `printError(text)` instead.

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,40 @@
...
+  return 4
+end
+
+local data = http.get(tArgs[1])
+if not(data) then
+  print("Error while getting the program")
```
apemanzilla: `printError(text)`

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,40 @@
...
+  return 5
+end
+
+local program = data.readAll()
+data.close()
+local run, err = load(program)
+if not(run) then
```
apemanzilla: `if not run then printError("Error running program: " .. err) return 6 end`

src/main/resources/assets/computercraft/lua/rom/programs/which
```diff
@@ -0,0 +1,12 @@
+--Made by Wilma456
+local Arg = {...}
+if Arg[1] == nil then
+	print("Usage: which <command>")
+	return
+end
+local path = shell.resolveProgram(Arg[1])
+if path == nil then
+	print("Can't find command "..Arg[1])
```
apemanzilla: `printError(text)`

asiekierka:
I don't think we should be teaching future generations of programmers to run untrusted code directly for the Internet.

apemanzilla:
@asiekierka while I would usually agree, `pastebin run` has already become the de facto way of installing anything more than a single file.

SquidDev:
There is also a bit of me which goes, "they've gotta learn, and ComputerCraft is the most harmless place for it to go wrong". This may not be the most appropriate attitude to security though.

CrazedProgrammer:
ComputerCraft has 0 security features and should have none. The openness and vulnerability of its systems will teach people to back up frequently.

asiekierka:
@CrazedProgrammer Having no security features is one thing, promoting features which go against common concepts of security - not just allowing them, but encouraging them -  is another. 

@SquidDev That is a fine viewpoint to have. And I did forget about "pastebin run" being a thing, I suppose webrun is not a far stretch from it.

justync7:
`pastebin run` in CC is essentially the same as `curl [url] | sh` in a real scenario. It's used quite commonly and may not be super secure, but it makes installing things a lot easier.

lupus590:
@justync7 presumably anyone doing that would know what the URL contains, I guess we need to assume CC users have the same common sense to check their downloads.

ObloxCC:
`pastebin run [code]`
`[url] [size]`
`are you sure you want to download [Y/n]`

src/main/resources/assets/computercraft/lua/rom/help/webrun
```diff
@@ -0,0 +1,5 @@
+webrun run a programm from the ineternet without downloading them
```
ObloxCC: program and internet are spelled incorrectly

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
```diff
@@ -0,0 +1,37 @@
+--Made by Wilma456
```
ObloxCC: This line would probably have to be removed before merging

src/main/resources/assets/computercraft/lua/rom/programs/which
```diff
@@ -0,0 +1,13 @@
+--Made by Wilma456
```
ObloxCC: This line should probably be removed too

src/main/resources/assets/computercraft/lua/rom/programs/http/webrun
Restioson: Change it to `args`

dan200:
Not interested in these. Sorry

src/main/resources/assets/computercraft/lua/rom/programs/which
lupus590: @dan200  What's your view on credit like this? Now that CC is open source should we be adding script author information? I know that treasure disks give their credit on the disk.

Restioson: @lupus590 IMO git already tracks your contributions via commit authors

dan200:
I would be open to adding a list of contributors to the credits help file
with everyone who's contributed to a merged PR. Feel free to start writing
one.

src/main/resources/assets/computercraft/lua/rom/programs/which
lupus590: @Restioson yes, but that doesn't get included in the compiled version of the mod. Although @dan200  [idea](https://github.com/dan200/ComputerCraft/pull/215#issuecomment-302019291) works too.

dan200:
Lua scripts aren't compiled dude

PR #217, opened 2017-05-08
Add a basic logging system

SquidDev:
This PR adds a basic logging system to ComputerCraft, with the aim of making it easier to track errors with peripherals and other registry objects.

 - Add `ILogger` interface and implementations for stdout and Log4J (which is what Minecraft uses). The reason I'm doing it this way is to avoid a hard dependency on Log4J, thus not breaking emulators.
 - Log error and warning messages in several cases where exceptions are either swallowed or rethrown with a loss of information.


SquidDev:
@apemanzilla Minecraft uses Log4j internally - I'm pretty sure slf4j isn't included and I don't want to add a dependency. already. It should be pretty trivial for emulators to register their own logger should they wish to.

lupus590:
Wouldn't `stderr` make more sense than `stdout`?

src/main/java/dan200/computercraft/core/logger/Logger.java
```diff
@@ -0,0 +1,65 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.logging.Level;
+
+public class Logger
+{
+    private static ILogger instance;
+
+    @Nonnull
+    public static ILogger getInstance()
```
JLLeitschuh: Any worries about this being called from two threads at the same time and causing two `ILogger` instances to be created?

SquidDev: Not really. It could happen, but I'm not especially worried if there are two `StdoutLogger`s created - it isn't required to be a singleton and their constructor hasn't any side effects.

dan200:
What's the purpose of this? Shouldn't we just log the same way vanilla does?

SquidDev:
Vanilla uses Log4J for logging, which this does use behind the scenes. The only reason for the extra level of indirection is not to add additional dependencies for ComputerCraft emulators. I'm happy to remove that layer though and just use Log4J directly.

dan200:
That does seem simpler.

src/main/java/dan200/computercraft/core/computer/ComputerThread.java
```diff
@@ -139,7 +140,7 @@ public void run() {
                                     // Step 3: abandon
                                     if( worker.isAlive() )
                                     {
-                                        //System.out.println( "computercraft: Warning! Failed to abort Computer " + computercraft.getDescription() + ". Dangling lua thread could cause errors." );
+                                        ComputerCraft.log.warn( "Failed to abort Computer " + computer.getID() + ". Dangling lua thread could cause errors." );
```
dan200: Until we've fixed cases where this can happen, could we leave this line commented out? Otherwise you're introducing a vector for a malicious user to flood the server's log.

src/main/java/dan200/computercraft/core/lua/LuaJLuaMachine.java
```diff
@@ -412,6 +415,7 @@ public void execute()
                                             }
                                             catch( Throwable t )
                                             {
+                                                ComputerCraft.log.error( "Error running task", t);
```
dan200: While no well behaved peripheral should ever throw an exception other than LuaException, some so, so this introduces another mechanism for a malicious flood the server log.  Seeing as the error is returned to the lua user anyway, could we remove this line?

src/main/java/dan200/computercraft/core/lua/LuaJLuaMachine.java
```diff
@@ -478,6 +482,7 @@ public void execute()
                         }
                         catch( Throwable t )
                         {
+                            ComputerCraft.log.error( "Error calling " + methodName + " on " + apiObject, t );
```
dan200: See above.

src/main/java/dan200/computercraft/shared/computer/apis/CommandAPI.java
```diff
@@ -98,6 +99,7 @@ public void shutdown()
             }
             catch( Throwable t )
             {
+                ComputerCraft.log.error( "Error running command.", t );
```
dan200: See above.

src/main/java/dan200/computercraft/shared/computer/apis/CommandAPI.java
```diff
@@ -208,6 +210,7 @@ private Object getBlockInfo( World world, BlockPos pos )
                                 catch( Throwable t )
                                 {
                                     // Ignore buggy command
+                                    ComputerCraft.log.error( "Error running command.", t );
```
dan200: See above.

src/main/java/dan200/computercraft/core/lua/LuaJLuaMachine.java
SquidDev: The issue here is *well behaved peripheral* - every so often people write buggy code which ends up throwing NPEs or other runtime exceptions. Whilst the exception is returned to the user, it doesn't include useful debugging information such as line numbers or stack traces - this PR is designed to make it easier for users/mod developers to track down and so solve issues in peripherals.

src/main/java/dan200/computercraft/shared/computer/apis/CommandAPI.java
SquidDev: I'm happy to remove these as this is less important. I'm still generally against swallowing exceptions without _some_ form of logging though.

src/main/java/dan200/computercraft/core/lua/LuaJLuaMachine.java
lupus590: Suggestion: Make this extra debug log info a config option which defaults to off? Perhaps mention in the config that it will be useful to turn it on when debugging peripheral addons.

SquidDev: I was thinking of doing that, but I'd honestly prefer it was on - these bugs appear in the wild too, and having logging enabled at all times makes it easier to process user bug reports too. I'm not convinced that flooding a log via peripherals is going to be the most efficient way to bring down a server. That being said, I obviously don't want to provide tools which can be abused.

dan200: Debug API which is only of use to developers should be passed with .trace() and silenced by default, no?

src/main/java/dan200/computercraft/shared/computer/apis/CommandAPI.java
dan200: In most cases, they're logged by the the Lua side which receives the error.

src/main/java/dan200/computercraft/core/lua/LuaJLuaMachine.java
SquidDev: I'm happy to switch to trace or debug (neither of which are shown IIRC). I chose error as this being hit indicates a bug with someone's code, which trace does not. I'd argue it should be noisy though - even when an error won't crash Minecraft (such as in network code), it'll still log it with stack trace and everything.

src/main/java/dan200/computercraft/shared/computer/apis/CommandAPI.java
SquidDev: Yes, but again you lose the stack trace which is the most valuable part of the error.

dan200: At the end of the day: I don't want to give players a means to programmatically spam the console. CC has been removed from modpacks in the past for violating this.

lupus590: @dan200  surely a config option to disable debug output would fix that though, while still allowing developers access to debug info when they need it.

granted @SquidDev  has a point that logs are useful in "wild encounters" but ultimatly it's a traidoff. Could a cooldown per peripheral/computer be useful? so no one problem can spam the log?

SquidDev: Ahhh, I wasn't aware of that. I'm not sure what the suitable compromise between "provide as much useful information to the developer as possible" and "don't spam the log" is. I'd personally err on more useful information with the ability to turn it off. 

From what I can tell from previous reports of log spam issue, it has been an incessant thing - this will only be a problem if someone can find a method which errors which is *generally* are rare thing.

---

A logging cooldown would be possible, though I'm not entirely sure if the implementation/benefit trade off is worth it.

lupus590: A modification to the idea, each computer/peripheral has a max rate of logging per minute.

SquidDev:
I've added a config option which will disable logging on the previously mentioned lines. I don't know if you want me to add the same checks for the various provider logs I've added too (peripheral, bundled cable, etc...)

src/main/java/dan200/computercraft/shared/computer/apis/CommandAPI.java
```diff
@@ -208,6 +213,10 @@ private Object getBlockInfo( World world, BlockPos pos )
                                 catch( Throwable t )
                                 {
                                     // Ignore buggy command
+                                    if( ComputerCraft.logPeripheralErrors )
+                                    {
+                                        ComputerCraft.log.error( "Error running command.", t );
```
dan200: Please change to "Error checking permissions on command"

src/main/java/dan200/computercraft/ComputerCraft.java
```diff
@@ -106,6 +107,7 @@
     public static String http_whitelist = "*";
     public static boolean disable_lua51_features = false;
     public static String default_computer_settings = "";
+    public static boolean logPeripheralErrors = true;
```
dan200: Please set the default to false :)

dan200:
Perfect. Merged!

SquidDev:
Thanks!

PR #218, opened 2017-05-08
Rewrite the computer thread system

SquidDev:
This makes a couple of significant changes to the original system, to reduce the number of threads created and allow for multiple threads in the future. There are several notable changes from the original implementation:

 - A blocking queue is used for the main task queue queue. This removes    the need for the "monitor" variable and allows for multiple threads polling this queue in the future.
 - The thread used to execute tasks is "cached" between tasks, significantly reducing the number of threads which need to be created. If a task needs to be stopped then the thread is then terminated and a new one constructed, though this rarely happens.

I'd really like someone with more multi-threading/concurrency knowledge to go over this and ensure it is sound. I'm not entirely sure if I need the semaphore class, or if `.wait()` and `.notify()` are sufficient. Nor am I sure if I've got my locks in entirely the correct place.


src/main/java/dan200/computercraft/core/computer/ComputerThread.java
```diff
@@ -6,227 +6,304 @@
...
+                    if( !done )
+                    {
+                        thread.interrupt();
+                        thread = null;
+                        runner = null;
+                    }
+                }
+            }
+            catch( InterruptedException ignored )
+            {
```
JLLeitschuh: Reset the interrupt flag??
```java
Thread.currentThread().interrupt();
```
http://www.yegor256.com/2015/10/20/interrupted-exception.html

src/main/java/dan200/computercraft/core/computer/ComputerThread.java
```diff
@@ -6,227 +6,304 @@
...
-                    m_computerTasksPending.add( queue );
+                    input.await();
+                    try
+                    {
+                        task.execute();
+                    }
+                    catch( Throwable e )
+                    {
+                        System.out.println( "ComputerCraft: Error running thread." );
+                        e.printStackTrace( System.out );
```
JLLeitschuh: What if the interrupted exception gets thrown?
Do you want to handle that differently?
What about something like an `OutOfMemoryError`?

JLLeitschuh:
Your whole starting/stopping infrastructure seems to be covered by guava's service helper methods.

https://github.com/google/guava/wiki/ServiceExplained

Perhaps consider using those instead?

src/main/java/dan200/computercraft/core/computer/ComputerThread.java
```diff
@@ -6,227 +6,304 @@
...
+                    {
+                        if( s_stopped ) return;
+                    }
+
+                    execute( queue );
+                }
+            }
+            catch( InterruptedException ignored )
+            {
+            }
```
JLLeitschuh: See my comment about the `InterruptedException` below.

src/main/java/dan200/computercraft/core/computer/ComputerThread.java
```diff
@@ -6,227 +6,304 @@
...
             {
-                //System.out.println( "Event queue overflow" );
+                System.out.println( "ComputerCraft: Error running thread." );
+                e.printStackTrace( System.out );
             }
         }
-        
-        synchronized (m_monitor)
+
+        void submit( ITask task )
```
JLLeitschuh: Can submit be called by multiple threads concurrently?

src/main/java/dan200/computercraft/core/computer/ComputerThread.java
SquidDev: This part was just a direct copy of the original implementation :). I think it would probably be safe just to catch `Exception`, though we may also want to catch `OrphanedThread`.

src/main/java/dan200/computercraft/core/computer/ComputerThread.java
SquidDev: No. Each `TaskRunner` is handle exactly one `TaskExecutor`.

src/main/java/dan200/computercraft/core/computer/ComputerThread.java
SquidDev: Yeah, I saw your comments on one of the older issues about this. I meaning to do it and forgetting :).

src/main/java/dan200/computercraft/core/computer/ComputerThread.java
SquidDev: Is this really needed here? The thread is going to terminate at this point anyway it doesn't seem worth it. I might be wrong though.

SquidDev:
> Your whole starting/stopping infrastructure seems to be covered by guava's service helper methods.

It sort of is. However, using that does seem rather overkill for a very simple problem. I'd also rather avoid introducing a dependency on Guava, so emulators don't have to bundle it (though I accidentally did that in #159).

SquidDev:
One thing worth noting is that if the computer cannot be aborted (such as in #174 or in string pattern matching) then the computer will spin forever. From what I can tell, [`LuaThread.State.lua_resume`](https://github.com/dan200/ComputerCraft/blob/master/luaj-2.0.3/src/core/org/luaj/vm2/LuaThread.java#L313-L339) (namely line 326) does not appear to respond to interruptions. 

Also worth noting that the `synchronized` blocks on `m_machine` cannot be interrupted, causing problems when a new thread is created. It might be better to switch to a `ReentrantLock` and use `lockInterruptibly`.

JLLeitschuh:
>  I'd also rather avoid introducing a dependency on Guava, so emulators don't have to bundle it (though I accidentally did that in #159).

Doesn't Minecraft/forge have a dependency on Guava already?

PR #221, opened 2017-05-08
Fix several crashes caused by incorrect NonNull annotations.

SquidDev:
Awfully sorry about this. It appears that Minecraft's annotations are occasionally wrong. IntelliJ will automatically add "not-null" checks on these annotations, resulting in crashes when they are actually null.

Whilst this can be fixed by disabling IntelliJ's checks, incorrect annotations are misleading for the developer too, so I have changed them where appropriate. Sorry about this.


SquidDev:
Thanks!

PR #223, opened 2017-05-09
Add /bin to Shellpath

Wilma456:
This Script checks, if /bin exists and a folder. If true, it will add it to the path. Program authors can just say to her installer "Place this file in /bin" and her program can be run from every folder on the computer. You don't need to edit startup.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/autorun/shellpath
+++ b/src/main/resources/assets/computercraft/lua/rom/autorun/shellpath
@@ -1,0 +1,3 @@
+if fs.isDir("/bin") then
+	shell.setPath(shell.path()..":/bin")
+end


xAnavrins:
Not sure what this is useful for, there's no /bin folder in the default /rom, if you need it for a custom OS, you can easily add it yourself with `shell.setPath(shell.path()..":/bin")`

apemanzilla:
I don't see the point of doing this by default, it's a single line to be added to the startup for people who want it.

KnightMiner:
The point is so you do not have to specifically define it, you have a default place that you have write access to (especially useful if a utility program wants to work across multiple paths without the need to create/modify startup every time)

apemanzilla:
@KnightMiner sure, but basically nothing uses `/bin` in CC right now and anything that does makes sure it's added to the path already (e.g. packman). If we're going to start making small changes like this it would be better to fix up the rest of the ROM as well.

Also, following CC conventions, `/programs` would make more sense (e.g. `/rom/programs`)

Wilma456:
What's so bad with a folder for downloaded programs?

apemanzilla:
@Wilma456 there's nothing inherently bad, the issue is that 1) it's not standardized and nothing else built-in uses or references `/bin`, and 2) it's literally a one line addition to the startup for anyone who does want to use it.

Wilma456:
Why not makiing using easy for the user? Why not make a standard with /bin? If /bin automatic added to path, the many installers will downland their programs to /bin. Some people download a program (maybe a login) as startup and don't want to edit this programs. There's no standardize in Computercraft now. And Computercraft is 5 Years old. Almost every program use his own folder. If you wait for a standardize, you can wait 100 years and nothing will happen.

ObloxCC:
> Why not makiing using easy for the user? Why not make a standard with /bin? If /bin automatic added to path, the many installers will downland their programs to /bin. Some people download a program (maybe a login) as startup and don't want to edit this programs. There's no standardize in Computercraft now. And Computercraft is 5 Years old. Almost every program use his own folder. If you wait for a standardize, you can wait 100 years and nothing will happen.

If anything like this would be added it would probably be either a second programs folder (out of rom) or a downloads folder (out of rom). Other than that I doubt anything else would be added because it would just clutter the users 'home/root' folder up.

Lignum:
I feel like there's a bigger issue here, being the modification of the path. It's difficult to write programs that modify the path since they would need to do it by adding code to startup, which is hard because there might already be a startup file. Thus, perhaps we should permit a `startup` folder to be made, which runs all files inside of it when the computer starts. That way, to automate it, you would simply have to drop a file inside `startup/`, not interfering with any other startup code. This would of course allow lots of other cool things other than changing the path to be done.

I'm not sure how exactly I would feel about having a default programs directory in the path, but I am certain that it shouldn't be called `/bin` since we're not dealing with binary files, but as @apemanzilla suggests, maybe something like `/programs` instead.

KnightMiner:
Only problem with a startup folder is there can be load order issues, as a program might depend on another program which is alphabetically after it. Not sure the simplest way to work with dependencies.

Lignum:
I would expect most startup programs to be self-contained. Either way, you'd have control over the loading order by simply prefixing the filenames with numbers.

apemanzilla:
I'd be fine with just running programs in `/startup/` in alphabetical order. If the program is sensitive to ordering just add a bunch of 0s to the start of the name of the startup script.

Wilma456:
The only problem is, that the folder is called bin and not programs? I can change this, if the most of you want this.

apemanzilla:
If #228 gets merged it will be easy for programs to add themselves to the path if necessary, I don't think we'd need any default path changes with it.

dan200:
Not a fan of this. CraftOS is not unix

PR #224, opened 2017-05-09
Lower case most resource names

SquidDev:
In Minecraft 1.11, resource names are required to be lower case. This PR lowercases all texture and model names, hopefully making the 1.11.2 port slightly easier. It also moves model registration code into `preInit`to ensure we don't get texture/model errors in the `preInit` stage.

Note that block states are not lower cased, as that isn't really needed. It might be worth converting all block names and the mod ID to be lower case now, to make migrating easier later on.


dan200:
This looks reasonable, can you confirm everything still works as it should? The block names and mod id's will need some thought though: we need to be careful not to break old save games.

SquidDev:
I don't think anything breaks - I've checked that all models (item and blocks) are loaded/displayed correctly, as well as ensuring all computer/turtle/book GUIs work.

KnightMiner:
If it does not transfer cleanly, Forge has an ID remapping feature that I have used in one of my mods before. Subscribe to the on missing mappings event in the main mod class.

PR #225, opened 2017-05-09
Update rednet.open()

Wilma456:
If you call rednet.open() without a argument, it will automatic choose a modem. So you have not to write the code to search the modem.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/rednet
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/rednet
@@ -8,8 +8,16 @@ local tHostnames = {}
 
 function open( sModem )
 	if type( sModem ) ~= "string" then
-		error( "expected string", 2 )
+	    for key,value in ipairs({"left","right","up","bottom","back","front"}) do
+            	if peripheral.getType(value) == "modem" then
+                	sModem = value
+                	break
+            	end
+            end
 	end
+	if sModem == nil then
+        	error("No modem found")
+   	end
 	if peripheral.getType( sModem ) ~= "modem" then	
 		error( "No such modem: "..sModem, 2 )
 	end


src/main/resources/assets/computercraft/lua/rom/apis/rednet
```diff
@@ -8,8 +8,16 @@ local tHostnames = {}
 
 function open( sModem )
 	if type( sModem ) ~= "string" then
-		error( "expected string", 2 )
+	    for key,value in ipairs({"left","right","up","bottom","back","front"}) do
```
SquidDev: You seem to be using a weird mix of spaces and tabs here. The rest of this function seems to be using tabs, so you might want to change this block to use tabs.

I'd also recommend using `rs.getSides()` instead of manually enumerating them. Perhaps even use `peripheral.getName()`.

You might also want to use more descriptive variable names than `key` and `value`: something like `n, sSide`  would be OK.

KingofGamesYami:
I prefer the current behavior - this seems rather "black box"y in a way that could introduce some problems.

ObloxCC:
I agree with @KingofGamesYami , I know many people that use both wired and wireless modems (on the same machine) and this would probably create some problems for them.

src/main/resources/assets/computercraft/lua/rom/apis/rednet
ObloxCC: Try to add an if statement here that ensures that there is only one modem attached (when auto selecting modem) so people don't get problems with multiple modems turning on and sending duplicate messages, etc.

SquidDev:
An alternative solution here would be to open every connected modem. I don't know what other people think about this.

SquidDev:
@apemanzilla The one issue with `peripheral.find` is that it returns the wrapped peripheral, while a string is accepted here. An alternative enhancement to this would be to accept multiple values to `rednet.open`. Then, for each argument:

 - If it is a string, use the current logic: check it's a string referencing a modem and if so open the appropriate channels, otherwise error.
 - If it is a table, try to check it's a modem otherwise error. Checking that `open`, `transmit`, `isOpen` and `close` _should_ be enough, though it's rather ugly. Then open the appropriate channels as above.
 - Else error, saying a string or table was expected.

This allows you to do `rednet.open("top")`, `rednet.open(peripheral.wrap("top"))` and `rednet.open(peripheral.find("modem"))`.

dan200:
I believe an old version of CC had this behaviour: at the time it was fine to do, because we only had wireless modems and we didn't have the ability to send non-rednet modem messages so opening one modem was much the same as the other.

Today, however, this is not the case: with wired modems, it's possible to have a completely different network of computers accessible from each side of the computer, so there's no sensible default to pick. It's better that the user choose, much as they do with other peripheral interactions.

PR #227, opened 2017-05-09
[WIP] Improving cable/wired modem interactions

SquidDev:
This is an attempt to give networking cables a more "multipart" feel, without actually implementing any multipart API. There are still some improvements which can be made, I'm just putting this up to gauge interest and get some thoughts on the current implementation.

## Current changes
 - Cable and modem can be broken individually, without destroying the tile entity.
 - Ray tracing will go through "holes" in the cable.
 - Pick block will determine which part you are looking at, using that for the item.
 - Selection box will only highlight the element you are looking at (the modem or cable part).
 - Selection box "hugs" the cable more closely.

## Still to do
 - Improve the selection box rendering - it is currently rather complicated and it probably can be tidied up a little.
 - Correctly clear the peripheral ID in the event of breaking modem part. Otherwise you can wrap a different peripheral type and preserve the ID.
 - Fix "breaking" overlay. This probably requires a TESR, which is a pain.


Wojbie:
I just realized this would merge conflict with #312. Would you mind if I redid #312 on your code and PR that changes into this?

SquidDev:
@Wojbie The two PRs solve very different problems, so it's probably better to keep them separate, and just rebase whichever one isn't merged.

Speaking of which, @dan200 do you have any idea when you'd next be able to look at PRs. I realise you've been busy with Redirection (and had some serious wrist injuries), so no rush :).

PR #228, opened 2017-05-10
Startup directories

Lignum:
This PR is an attempt to solve the problem of having only a single startup file. You can now create a `startup` directory, which will have all of its files executed on startup. These will be run sorted alphabetically by the filename.

For example, the directory tree
```
/
|---- startup
       |---- hello.lua
       |---- world.lua
```
assuming each file prints its filename would print "hello.lua" and then "world.lua". These files don't have to have a .lua extension, for reference.

You can still create a  `startup` (and `startup.lua` as of #211) file as usual. However, the startup directory may not have a .lua extension.

This was inspired by a discussion in #223. With this new system, programs will have a much easier time modifying startup to include their own behaviour.


src/main/resources/assets/computercraft/lua/rom/startup
```diff
@@ -209,22 +209,45 @@ if fs.exists( "/rom/autorun" ) and fs.isDir( "/rom/autorun" ) then
...
+                tUserStartups = startups
                 break
             end
         end
     end
 end
-if sUserStartup then
-    shell.run( sUserStartup )
+if type( tUserStartups ) == "table" then
+    table.sort( tUserStartups )
```
SquidDev: `fs.list` will sort the output now, so this isn't strictly needed. Though I guess this is clearer.

Lignum: Oh I wasn't aware. Might be clearer, but I don't wanna sort unnecessarily either.

src/main/resources/assets/computercraft/lua/rom/startup
```diff
@@ -209,22 +209,45 @@ if fs.exists( "/rom/autorun" ) and fs.isDir( "/rom/autorun" ) then
...
 
+local function findStartups( sBaseDir )
+    local tStartups = nil
+    local sBasePath = "/" .. fs.combine( sBaseDir, "startup" )
+    local sStartupNode = shell.resolveProgram( sBasePath )
+    if sStartupNode then
+        tStartups = { sStartupNode }
+    end
+    -- It's possible that there is a startup directory and a startup.lua file, so this has to be
+    -- executed even if a file has already been found.
```
ObloxCC: use `--[[` and `]]--` for multi line comments instead of `--` every line

SquidDev: I think this is just a matter of preference. I'd agree with you for a multi-paragraph comment but here it doesn't really matter - several other `rom` files use `--` for multi-line comments.

src/main/resources/assets/computercraft/lua/rom/startup
```diff
@@ -209,22 +209,44 @@ if fs.exists( "/rom/autorun" ) and fs.isDir( "/rom/autorun" ) then
...
+    if fs.isDir( sBasePath ) then
+        if tStartups == nil then
+            tStartups = {}
+        end
+        for _,v in pairs( fs.list( sBasePath ) ) do
+            tStartups[ #tStartups + 1 ] = shell.resolveProgram( "/" .. fs.combine( sBasePath, v ) )
```
apemanzilla: I don't really see the point of using `shell.resolveProgram` here, you already have the directory path and filename so you may as well just use `fs.combine(sBasePath, v)`

SquidDev: I'd guess so it catches `startup.lua` and `startup`. Though it does seem rather overkill to use `shell.resolveProgram` just for that.

apemanzilla: It makes sense for the `startup` script, but this is for listing scripts in `/startup/`, where we already have the exact filename.

SquidDev: Good catch. I hadn't thought about that.

src/main/resources/assets/computercraft/lua/rom/startup
Lignum: Yeah, I generally only use `--[[`/`]]` if I have more than 4 lines of comment, which is rarely the case.

src/main/resources/assets/computercraft/lua/rom/startup
```diff
@@ -209,22 +209,44 @@ if fs.exists( "/rom/autorun" ) and fs.isDir( "/rom/autorun" ) then
...
+            if startups then
+                tUserStartups = startups
                 break
             end
         end
     end
 end
-if sUserStartup then
-    shell.run( sUserStartup )
+if type( tUserStartups ) == "table" then
```
dan200: No need for a type check here when a nil check will do

src/main/resources/assets/computercraft/lua/rom/startup
```diff
@@ -209,22 +209,44 @@ if fs.exists( "/rom/autorun" ) and fs.isDir( "/rom/autorun" ) then
...
+        tStartups = { sStartupNode }
+    end
+    -- It's possible that there is a startup directory and a startup.lua file, so this has to be
+    -- executed even if a file has already been found.
+    if fs.isDir( sBasePath ) then
+        if tStartups == nil then
+            tStartups = {}
+        end
+        for _,v in pairs( fs.list( sBasePath ) ) do
+            tStartups[ #tStartups + 1 ] = "/" .. fs.combine( sBasePath, v )
```
dan200: You need to check the things in the "startup" folder aren't folders themselves.

Lignum: Oops, will fix

src/main/resources/assets/computercraft/lua/rom/startup
Lignum: Hm, I did this to be 100% sure that it's a table, but I guess that's a bit paranoid since there's no external code modifying `tUserStartups`.

dan200:
Great! Document this change in changelog and whatsnew and i'll merge it in.

Lignum:
Cheers!

src/main/resources/assets/computercraft/lua/rom/help/whatsnew
```diff
@@ -1,34 +1,35 @@
...
+* Turtle upgrades now act as tools and peripherals
+* The Filesystem.list result is now sorted
+* The number of values to unpack can now be manually specified
+* Small terminal & monitor rendering improvements
+* General improvements to the documentation
+* Redstone inputs are no longer reset when adding peripherals
+* Turtles now use tinting
+* shell.resolveProgram now picks up on *.lua files
+* Fixed a handful of bugs in ComputerCraft
+* Startup can now be a directory containing multiple startup files
```
apemanzilla: Might be worth mentioning that they will be executed in alphabetical order.

KnightMiner: And that `startup.lua` always runs before files in the `startup` folder.

Lignum: Guys this is a changelog, not an accurate description of the feature.

apemanzilla: Fair enough. Is there a `startup` help entry?

PR #230, opened 2017-05-11
Dye improvements

SquidDev:
This PR makes it possible to dye turtles arbitrary colours. Turtles can now be coloured using the same recipe as disks and leather armour.

## Changes
 - Add more general item colouring crafting recipe, which allows using ore dictionary dyes.
 - Migrate disks to this recipe.
 - Allow dying turtles via `ITurtleAccess` or crafting with dyes.

## Further improvements
A couple of people have been asking for the ability to dye computers and peripherals too. This seems slightly excessive, but worth considering. However, dyable cables could be useful - they only connect to cables without a colour or of the same colour.

## Screenshots
A silly little upgrade to test this. This isn't included in the PR.

Crafting a turtle:

Pocket computers:


Wilma456:
Looks really cool

src/main/java/dan200/computercraft/api/turtle/ITurtleAccess.java
```diff
@@ -122,19 +122,37 @@
      * @param dyeColour 0-15 to dye the turtle one of the 16 standard Minecraft <em>dye</em> colours, or -1 to remove
      *                  the dye from the turtle.
      * @see #getDyeColour()
+     * @see #setColour(int)
      */
     void setDyeColour( int dyeColour );
```
dan200: I don't mind just removing these two methods. I'm not aware of any mods that used them and small API changes are expected between minecraft versions.

dan200: "These two methods" being setDyeColour and getDyeColour

src/main/java/dan200/computercraft/client/render/TileEntityTurtleRenderer.java
```diff
@@ -51,24 +50,7 @@
...
-        new ModelResourceLocation( "computercraftedu:turtleJunior_grey", "inventory" ),
-        new ModelResourceLocation( "computercraftedu:turtleJunior_pink", "inventory" ),
-        new ModelResourceLocation( "computercraftedu:turtleJunior_lime", "inventory" ),
-        new ModelResourceLocation( "computercraftedu:turtleJunior_yellow", "inventory" ),
-        new ModelResourceLocation( "computercraftedu:turtleJunior_lightBlue", "inventory" ),
-        new ModelResourceLocation( "computercraftedu:turtleJunior_magenta", "inventory" ),
-        new ModelResourceLocation( "computercraftedu:turtleJunior_orange", "inventory" ),
-        new ModelResourceLocation( "computercraftedu:turtleJunior_white", "inventory" ),
-    };
+    private static ModelResourceLocation BEGINNER_TURTLE_COLOUR_MODEL = new ModelResourceLocation( "computercraftedu:turtleJunior_white", "inventory" );
```
dan200: Note to self: Make a decision what to do about this legacy ccedu stuff!

src/main/java/dan200/computercraft/shared/util/ColourTracker.java
```diff
@@ -0,0 +1,48 @@
...
+        int avgR = totalR / count;
+        int avgG = totalG / count;
+        int avgB = totalB / count;
+
+        float avgTotal = (float) total / (float) count;
+        float avgMax = (float) Math.max( avgR, Math.max( avgG, avgB ) );
+        avgR = (int) (avgR * avgTotal / avgMax);
+        avgG = (int) (avgG * avgTotal / avgMax);
+        avgB = (int) (avgB * avgTotal / avgMax);
+
```
dan200: Can't figure out what this second block of code does, why don't we just do a straight average?

SquidDev: I'm not entirely sure myself. This code is taken from the Disk recipe, which is taken from Minecraft's leather dying recipe. I stuck with this rather than the simpler code to remain consistent with the rest of Minecraft.

src/main/java/dan200/computercraft/api/turtle/ITurtleAccess.java
SquidDev: Great, that makes the implementation a little simpler :).

src/main/java/dan200/computercraft/client/render/TileEntityTurtleRenderer.java
SquidDev: Ideally the API could be powerful enough that there didn't have to be any special support. I'm not sure how feasible that is though.

SquidDev:
@dan200 Just wondering what your thoughts are on changing `IPocketAccess` to use RGB instead of dye colours? This would allow modems to use their original colour which would be nice.

dan200:
For the light? Sounds good to me. Dyeable pocket computers in general
sounds good too :)

SquidDev:
Beautiful:

java_2017-05-14_15-01-52

There is a bit of me which would like to add a Water Bucket + Item recipe to clear the dye, but dunking your electronics in water probably isn't the wisest decision. I don't know what other people's thoughts are on the matter?

Vexatos:
Well, the "electronics" are made of redstone.

SquidDev:
> Well, the "electronics" are made of redstone.

Which also washes off with a water bucket! I wonder if a wet sponge would make more sense, though it is less accessible than a bucket.

KnightMiner:
I think we should support dunking it in a water filled cauldron to clear the dye, but give a broken computer from that. Totally a good idea.

For a serious solution, i was about to suggest a sponge or wool to clear the dye though. Maybe the undyeing recipe dyes the wool to signify it's removal?

Lignum:
> Which also washes off with a water bucket!

> I'm​ think we should support dunking it in a water filled cauldron to clear the dye, but give a broken computer from that.

Honestly, I think we'd get away with it, since you can already drop a pocket computer in water and nothing will happen.

apemanzilla:
> Honestly, I think we'd get away with it, since you can already drop a pocket computer in water and nothing will happen.

Hmm, maybe that should change? 😈

dan200:
We don't allow disks to have their colour "cleared", so it doesn't seem necessary here either. Especially as you can combine dyes to get the rgb colour you want if you really like!

SquidDev:
Great. Thank you!

PR #231, opened 2017-05-11
Fix Bug with setting Colour

Wilma456:
If you calls term.setBackgroundColor or term.setTextColor with the wrong argument, your computer will crash. With my Bugfix only your program crash and your computer keep runing.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/window
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/window
@@ -277,11 +277,23 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     end
 
     function window.setTextColor( color )
-        setTextColor( color )
+		for key,value in pairs(colors) do
+			if value == color then
+        		 setTextColor( color )
+				return
+			end
+		end
+		error("Color doesn't exists")
     end
 
     function window.setTextColour( color )
-        setTextColor( color )
+		for key,value in pairs(colors) do
+			if value == color then
+        		 setTextColor( color )
+				return
+			end
+		end
+		error("Colour doesn't exists")
     end
 
     function window.setPaletteColour( colour, r, g, b )
@@ -316,11 +328,23 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     end
 
     function window.setBackgroundColor( color )
-        setBackgroundColor( color )
+		for key,value in pairs(colors) do
+			if value == color then
+        		setBackgroundColor( color )
+				return
+			end
+		end
+		error("Color doesn't exists")
     end
 
     function window.setBackgroundColour( color )
-        setBackgroundColor( color )
+		for key,value in pairs(colors) do
+			if value == color then
+        		setBackgroundColor( color )
+				return
+			end
+		end
+		error("Colour doesn't exists")
     end
 
     function window.getSize()
--- a/src/main/resources/assets/computercraft/lua/rom/apis/window
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/window
@@ -268,18 +268,19 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     function window.isColour()
         return isColor()
     end
- 
-    local floor, log, log2 = math.floor, math.log, math.log(2)
+
     local function setTextColor( color )
-	if not(type(color) == "number") then
-        error("Expected number", 2)
-    else
-        -- Type assertions, whatever, etc...
-        if not tHex[color] then
-           if color < 0 or color > 0xFFFF then error("Colour out of range", 2) end
-              color = 2^floor(log(color) / log2)
-           end
+	if type(color) == "number" then
+       if type(tHex[color]) == "string" then
         nTextColor = color
+        if bVisible then
+            updateCursorColor()
+        end
+       else
+        error("Colour out of range", 2)
+       end
+    else
+       error("Color must be a number",2)
     end
     end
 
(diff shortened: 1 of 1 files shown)


Wilma456:
Try term.setBackgroundColor(2000). This will crashed your Computer and it's a number.  The Code from SquidDev doesen't stop this.

SquidDev:
@Wilma456 That seems a bug with multishell: `term.native().setBackgroundColour(2000)` works as expected.

Wilma456:
The Code now checks just, if color is a number. The most errors are caused by spell errors, so it will solve the most errors. If it a bug in multishell, it must be fixed in multishell I think.

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -270,19 +270,19 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
...
-        nTextColor = color
-        if bVisible then
-            updateCursorColor()
+        if type(color) == "number" then
+	    nTextColor = color
+            if bVisible then
+               updateCursorColor()
+            end
+        else
+           error("Color must be a number")
```
SquidDev: Would probably be better to use `error("Expected number", 2)` here, as that is what the `term` API uses.

SquidDev:
> If it a bug in multishell, it must be fixed in multishell I think.

Sorry, it is a bug in the window API. I always conflate the two for whatever reason. My suggested change involving `math.log` *should* solve the problem, but is rather inefficient.

Wilma456:
nTextColor = math.log(color) and nTextColor = math.floor(math.log(color) / math.log(2)) + 1 does not work. The computer crashed on boot.

SquidDev:
@Wilma456 Did you change the `tHex` table too? I'll have a look - I'd have thought it would work.

Wilma456:
No. What I should change in the table?

SquidDev:
Right. Having another look at the code, I think the "best" solution would be:

You basically normalise the colour to be between 1 and 16, so it needs to change to a simple array:

```lua
local tHex = { "0", "1", "2", -- [[ etc to "f" ]] }
```
An alternative solution would be to have both. So do:

```lua
local floor, log, log2 = math.floor, math.log, math.log(2)
local function setBackgroundColor( color )
    -- Type assertions, whatever, etc...
    if not tHex[color] then
        if color < 0 or color > 0xFFFF then error(""Colour out of range", 2) end
         color = 2^floor(log(color) / log2)
    end
    nBackgroundColor = color
end
```

Wilma456:
@SquidDev With this code the colours are wrong

SquidDev:
@Wilma456 Can you elaborate? It appears to work correctly for me.

Wilma456:
@SquidDev 

Here's my actual code: https://pastebin.com/qRmyUevL

Ans here's a picture of CraftOS using this code: http://www.bilder-upload.eu/show.php?file=67ed09-1494688304.png

SquidDev:
Your implementation of `setTextColor` sets `nBackgroundColor` instead of `nTextColor`.

Wilma456:
@SquidDev Copy and paste Error. Sorry. :man_facepalming: 

@dan200 The bug is now fixed and the PR is ready to merge.

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -268,16 +268,19 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
...
     local function setTextColor( color )
-        if type(color) == "number" then
-	    nTextColor = color
-            if bVisible then
-               updateCursorColor()
-            end
-        else
-           error("Color must be a number")
-        end
+	if not(type(color) == "number") then
```
SquidDev: It'd probably be nicer to use `~=` here - `if type(color) ~= "number" then`.

SquidDev: Also, not entirely sure what is going on with the indentation here - it seems like a really weird mix of spaces and tabs.

dan200:
This code is very messy. Does someone want to just submit a cleaner version of this same code that just checks if the number is a key in tHex? We don't need to support people passing in non-colours.

Wilma456:
@dan200 Done!

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -270,19 +270,23 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
...
+       if type(tHex[color]) == "string" then
         nTextColor = color
         if bVisible then
             updateCursorColor()
         end
+       else
+        error("Colour out of range", 2)
+       end
+    else
+       error("Color must be a number",2)
```
KnightMiner: Minor issue, but this line spells it "color" and the above is "colour", being consistent with whatever old errors do would be better.

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -312,17 +316,21 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
...
 
     local function setBackgroundColor( color )
+    if type(color) == "number" then
+       if type(tHex[color]) == "string" then
         nBackgroundColor = color
+       else
+        error("Colour out of range",2)
+       end
+    else
+       error("Color must be a number",2)
```
KnightMiner: Same here

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -312,17 +316,21 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
...
-    function window.setBackgroundColor( color )
-        setBackgroundColor( color )
     end
 
-    function window.setBackgroundColour( color )
-        setBackgroundColor( color )
-    end
+    window.setBackgroundColor = setBackgroundColor
 
+    window.setBackgroundColour = setBackgroundColo
```
KnightMiner: Should be "setBackgroundColor", not "setBackgroundColo"

src/main/resources/assets/computercraft/lua/rom/apis/window
```diff
@@ -270,19 +270,23 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     end
 
     local function setTextColor( color )
+	if type(color) == "number" then
+       if type(tHex[color]) == "string" then
```
KnightMiner: Indentation seems a little odd here

dan200:
This code is still super messy. Rather than having a long conversation spelling out how to fix, i've just committed a cleaner implementation: https://github.com/dan200/ComputerCraft/commit/8c4331d15aaf07b87c1033f36799813bf633a188
Closing as this merge is no longer necessary

PR #232, opened 2017-05-11
Do not insert empty or duplicate entries into shell history

SquidDev:
If a string is empty or the same as the previous command then it will not be inserted into history. This makes history slightly easier to navigate as you don't have to go over several duplicate entries or empty lines.


dan200:
sure

SquidDev:
Thank you! 😃.

PR #234, opened 2017-05-12
Remap all blocks and items to underscore_case

SquidDev:
This makes block/model names a little mode consistent and should help with porting to 1.11.2.

 - Prefix all tile entities with `computercraft:`.
 - Change all `pascalCase` and `CC-*` items to use `underscore_case`.
 - Listen to the missing mappings event and gracefully convert blocks/items (as suggested by @KnightMiner in #224).

Note, all the resources have changed because git mapped CRLF to LF. I'm happy to have a go at reverting that, but I'd rather convert the remaining resources to use LF.


dan200:
Have you tested this with a world saved in 1.79? Hopefully the missing mappings event will help with the block names, but tile entity names get saved out too

SquidDev:
Yep, everything converts as expected. There was a long cycle of load backup, start game and repeat while testing this 😄. It uses `GameRegistry.registerTileEntityWithAlternatives` which will save everything in the prefixed form, but will still handle the non-prefixed version fine.

dan200:
Excellent :) This looks good then

SquidDev:
Great. Thanks!

PR #235, opened 2017-05-12
Add .lua extension to all programs and startup

Wilma456:
The title says it all


SquidDev:
> Add .lua *exteansio* to all *progarsm* and startup

I fear the title doesn't say what you think it says 😛.

Wilma456:
#211 
dan200 wrote:

> Sounds good! I'd also be perfectly ok with renaming all builtin programs to a ".lua" extension, and doing the same for APIs, startup scripts, and anything else we call.
I have done that now for all files in programs and the startup file. The APIs doesen't work with .lua extension at the moment.

CrazedProgrammer:
Have some respect for Dan and the community.

lupus590:
I've had this in my code pack for a while and haven't had any issues with it, that said though I haven't played for a while and rarely used APIs with an extension so I can't confirm if it does work

```lua
--allows loading of api's with .lua on the end of the file name
--author: theoriginalbit
--post link: http://www.computercraft.info/forums2/index.php?/topic/19383-loading-api-help/page__view__findpost__p__185496

local function loadAPI(path)
  --# check the file exists
  if not fs.exists(path) then
        error("File does not exist", 2)
  end
  --# check the path isn't a directory
  if fs.isDir(path) then
        error("Cannot load directory", 2)
  end
  --# get the filename and extension (if exists)
  local name = fs.getName(path)
  --# remove the extension (if exists)
  name = name:match("(%a+)%.?.-")
  --# check there isn't an API loaded under this name
  if _G[name] then
        error("API "..name.." already loaded", 2)
  end
  --# os.loadAPI logic
  local env = setmetatable({}, { __index = _G })
  local func, err = loadfile(path)
  if not func then
        error(err, 0)
  end
  setfenv(func, env)
  func()
  local api = {}
  for k,v in pairs(env) do
        api[k] =  v
  end
  _G[name] = api
  return true
end
```

SquidDev:
I think my one concern about this is that auto-complete display the file extensions, which is rather ugly. For instance: `cle` will complete to `clear.lua` rather than `clear`. If we're going to do this, I think `shell.programs()` should remove the `.lua` (as long as it doesn't introduce ambiguities).

Wilma456:
@SquidDev This has been done in #239 now.

@lupus590 I think this should be down in os.loadAPI and not in it's own function.

@CrazedProgrammer Sorry, I'm not a native speaker.

CrazedProgrammer:
Not being a native speaker is not an excuse :P

dan200:
Fix the conflicts and i'll merge this. We'll need to make a few other changes to make everything play nicely with extensions, but this is the first step.

Wilma456:
@dan200 I have made with #250 a  new PR. You can close this PR.

Issue #212, opened 2017-05-07
[Enhancement] Idea: man pages


Restioson:
It would be useful if we could have a command line utility to print out a lengthy help page for a utility, essentially the `man` command from *nix. We could store the man page in the same folder as the program, with the same name as the program, and with the extension `.man.txt`, or we could store it at `/man/program_name/man.txt`. This is just a thought.

Lignum:
There's the `help` program.

Restioson:
Can you do `help pastebin`? I'll try it out in a sec.

Lignum:
You can, it even works with most APIs. Try `help term`.

SquidDev:
Yeah, `help` exists. However, it would be nice if it `less`-like pagination. For instance: step forwards/back with arrow keys and page-up/down\*, quit early with `q`, etc...

\*: Sure, you can also use C-f to move forwards, but ctrl keys are a faff to add and probably not needed.

apemanzilla:
@SquidDev I've been (sort of) working on some changes to the `textutils` library and that's one of the things I plan to implement.

CrazedProgrammer:
This would be very nice, I love manpages!

ObloxCC:
I could try to fix up the help system if you want?

dan200:
We already have "help".

SquidDev:
@dan200 There have been some comments that the built-in help files aren't always that useful. Part of these files' usefulness is limited by the method by which they are read (`help`). What would your stance be on someone contributing a more powerful help browser? What form this takes is probably up for discussion.

Restioson:
Cool, was just an idea :smile:

PR #236, opened 2017-05-12
Add readline keybindings to read function

SquidDev:
This adds a selection of keybindings to the `read` function which mirror those provided by readline (shell command prompts, several REPLs, etc...). I'm not entirely sure whether this should actually be included - it does bloat the already large `bios.lua`. However it does make inputting long text (especially in the `lua` program) slightly easier. I guess you can make up your mind for yourself 😄.

## Keybindings
This provides the most basic keybindings as given [here](https://tiswww.case.edu/php/chet/readline/rluserman.html#SEC3).

 - <kbd>C-a</kbd> / <kbd>C-e</kbd> to jump to beginning/end of line
 - <kbd>C-d</kbd> to exit read without inputting anything. This could cause problems as `read` can now return `nil`. However, it does mean you can exit `lua` and `shell` without having to type an explicit "exit".
 - <kbd>C-b</kbd> / <kbd>C-f</kbd> to move left/right one character
 - <kbd>M-b</kbd> / <kbd>M-f</kbd> to move left/right one word.
 - <kbd>C-p</kbd> / <kbd>C-n</kbd> to move backwards/forwards in history or completions.
 - <kbd>C-u</kbd> / <kbd>C-k</kbd> to delete to the beginning/end of line.
 - <kbd>C-w</kbd> / <kbd>M-d</kbd> to delete to the beginning of the previous word/end of next    word. (Note, the docs say it should be <kbd>M-DEL</kbd> rather than <kbd>C-w</kbd>. However, most of the things I've experimented with use the latter).

This doesn't define any of the more complex command like undo, transpose, yanking, etc... I'm happy to add them if people *really* want them, but it feels rather overkill.

## Implementation
I'm currently listening to `key_down`/`key_up` events to track the status of control/alt keys. This is fine most of the time, though could cause problems with multiple users or people exiting the GUI with keys down. A couple of changes which could be made elsewhere to fix this:

 - Trigger `key_up` events when the terminal is closed.
 - Include modifier keys in the `key` event. I'm not a fan of this as it seems too "friendly" for the user, but it is the most fool-proof way.

## Possible extensions
It might be nice to add similar shortcuts to `edit`. A similar system could also be added to the pager described in #212. I'd be happy to add them if people want.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -282,6 +282,8 @@ function read( _sReplaceChar, _tHistory, _fnComplete )
     local sLine = ""
     local nHistoryPos
     local nPos = 0
+    local tDown = {}
+    local nMod = 0
     if _sReplaceChar then
         _sReplaceChar = string.sub( _sReplaceChar, 1, 1 )
     end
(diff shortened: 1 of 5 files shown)


dan200:
Not a fan of these, sorry. CraftOS is not unix!

SquidDev:
That's fine, I wasn't too sure about it myself 😄.. I don't there is an equivalent set of key bindings you'd consider using instead? For instance Ctrl+Left/Right to jump forwards/backwards words?

PR #237, opened 2017-05-12
[Peripheral] Speaker

Restioson:
# Speaker Peripheral
This PR adds the venerable speaker peripheral. This allows computers to play any sound! Theoretically, you can even play sounds from other mods, provided that you use the correct domain.

# Possible Use Cases
- Notification API! You can notify the user with a snazzy ping. Useful as currently there is no way of notifying players of something on a computer
- Music! You can play records
- Custom music! You can make music using this by combining different sounds
- Super sneaky wireless range pinpointer! You can play a ping on a turtle every time it gets a packet, and relay the ping back to home base. You could use this for PVP. (example coming soon) 

# Example
```lua
speaker.playSound("minecraft:entity.cow.ambient") -- Plays entity.cow.ambient from domain minecraft
sleep(1 / 20) -- Sleeps the required amount of time, converting from ticks to seconds
speaker.playSound("entity.cow.ambient", 0.1) -- Plays entity.cow.ambient (in default domain, minecraft) at volume 0.1
sleep(1 / 20)
speaker.playSound("entity.cow.ambient", 1, 0.1) -- Plays entity.cow.ambient at volume 1 and pitch 0.1, thereby slowing it down
sleep(1 / 20)
speaker.playNote("harp", 1, 2) -- Plays the noteblock harp note at volume 1 and pitch 2
```

# Functions
 - `speaker.playSound` - Takes 1 positional arguments (`string` resource name) and 2 optional positional arguments (`double` volume and `double` pitch). Returns `true` or `false` based on whether the `resourcename` exists and is a sound. Plays the sound at `resourcename` with `volume` and `pitch`, under category `RECORDS`.
- `speaker.playNote` - Takes 1 positional arguments (`string` instrument) and 2 optional positional arguments (`double` volume and `double` pitch). Returns `true` or `false` based on whether the `instrument` exists. Plays the sound at `blocks.note.instrument` with `volume` and `pitch`, under category `RECORDS`.

# Rate limiting 
To prevent sound pollution, sounds are limited to once per tick. However, up to 8 notes can be played per tick to allow for chords. If a note is played, then this will not allow a sound to be played in the same tick as well. 

# Summation of changes
 - Added speaker block
 - Added `speaker` peripheral
 - Added `speaker.playSound(resourceLocator, volume, pitch)` function in java
 - Added `speaker.playNote(noteblockInstrument, volume, pitch)` function in java
 - Added turtle speaker
 - Added pocket speaker

# Thanks
Special thanks to @SquidDev for helping me understand and navigate the codebase and with git ~~too many times than I care to admit~~, and @gegy1000 for helping me out with Forge.


Wojbie:
Just a question. If its attached to pocket in your example how are you connecting wireless to turtle with only speaker peripheral attached to pocket? Or is it not attached to pocket in this example?

Restioson:
My video used an outdated version of this PR -- I'll update it. Thanks for catching it!

Vexatos:
I still think this belongs into its own addon instead of CC itself.

Cruor:
As you cannot have two peripherals (currently) on a pocket computer, how is the user not able to currently be notified of anything on a computer? You can just hook up a monitor, redstone lamp or noteblock to the computer, which will just as easily notify any players nearby depending on how its set up. 

I am not against this being a peripheral, but I don't agree with it being in the main mod, and as @Vexatos says, this functionality is already in many of the existing CC peripheral mods. I believe that a simple beep function, without the need of any peripherals is the best solution, as it would also work with the current peripheral limit on pocket computers.

Wojbie:
Soo first thing first. This is a great peripheral don't get me wrong. But i will agree with the people before me it does feel like it shouldn't be in CC by default. 

I also fully agree with @Cruor that getting something like os.beep(Hz) for pockets would be a greatest thing ever since toasted sliced bread.

Restioson:
I'll add os.beep, but I don't know whether this should be in the same PR. Thoughts?

Wojbie:
I would suggest separate PR. That way Dan can choose to add both, one, or none.

src/main/java/dan200/computercraft/shared/peripheral/common/BlockPeripheral.java
```diff
@@ -323,6 +333,10 @@ public IBlockState getActualState( @Nonnull IBlockState state, IBlockAccess worl
                 }
                 break;
             }
+            case Speaker: {
```
gegy1000: Bracket should be on a newline to fit with the rest of the code style

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,242 @@
...
+        };
+    }
+
+    @Override
+    public Object[] callMethod(IComputerAccess computerAccess, ILuaContext context, int methodIndex, Object[] args) throws LuaException
+    {
+        switch (methodIndex)
+        {
+            // playsound
+            case 0: {
```
gegy1000: This `{` should also be on a newline.

src/main/java/dan200/computercraft/shared/peripheral/common/BlockPeripheral.java
Restioson: k

Restioson:
I will submit a separate PR in time.

dan200:
Great work! This has been implemented really well, the API is nice, and even the block graphics look nice! Bravo. I only have one thing I need changed, which I will add in a comment in a minute

Just a note for anyone adding new peripherals in the future though: You don't have to make your peripheral a subtype of a "peripheral" block if you don't want to: it's ok to register a new block ID. Combining them all into one block is a legacy (and a backwards compatibility constraint) of the days when we only had 255 block IDs to play with for the whole game!

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,243 @@
...
+    public String[] getMethodNames()
+    {
+        return new String[] {
+                "playSound", // Plays sound at resourceLocator
+                "playNote" // Plays note
+        };
+    }
+
+    @Override
+    public Object[] callMethod(IComputerAccess computerAccess, ILuaContext context, int methodIndex, Object[] args) throws LuaException
```
dan200: So here's my issue with this PR: All the code called from this method is completely non-thread safe. Methods called from Lua always happen on the Lua thread, so it's not safe to call methods on anything managed by the main world directly unless it's your own data that you're correctly managing access to with syncronized blocks.

My recommendation is to use ILuaContext.issueMainThreadTask to ensure your sounds are played on the main thread. It's still safe and recommended to check all your parameters are correct before making this call so you can earlyout and give appropriate errors if they're incorrect.

Restioson: Thanks! I'll do this today, and hopefully resolve the conflicts with upstream.

Restioson:
Credit @gegy1000 for model though. (I can't model)

Restioson:
As far as I am concerned, this PR is ready for merge.

src/main/java/dan200/computercraft/ComputerCraft.java
```diff
@@ -325,6 +335,8 @@ public static void syncConfig() {
         turtlesObeyBlockProtection = Config.turtlesObeyBlockProtection.getBoolean();
         turtlesCanPush = Config.turtlesCanPush.getBoolean();
 
+        maximumFilesOpen = Math.max(1, Config.maximumFilesOpen.getInt());
```
dan200: I presume this was meant to be maxNotesPerTick?

src/main/java/dan200/computercraft/shared/peripheral/speaker/SoundTask.java
```diff
@@ -0,0 +1,45 @@
+/*
```
dan200: It wasn't necessary to create a whole new class for this. Use an anonymous class.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,250 @@
...
+        {
+            throw new LuaException("Expected string, number (optional), number (optional)");
+        }
+
+        if (!SoundEvent.REGISTRY.containsKey(new ResourceLocation("block.note." + arguments[0])))
+        {
+            throw new LuaException("Invalid instrument, \"" + arguments[0] + "\"!");
+        }
+
+        if (arguments.length > 1)
```
dan200: This will fail if you pass (string, nil, nil), when it probably shouldn't (this commonly happens when you wrap a method in Lua). You should do null checks

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,250 @@
...
+        {
+            if (!(arguments[1] instanceof Double))   // Arg wrong type
+            {
+                throw new LuaException("Expected string, number (optional), number (optional)");
+            }
+            volume = ((Double) arguments[1]).floatValue();
+
+        }
+
+        if (arguments.length > 2)
```
dan200: Same as above

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,250 @@
...
+        {
+            if (!(arguments[1] instanceof Double))  // Arg wrong type
+            {
+                throw new LuaException("Expected string, number (optional), number (optional)");
+            }
+            pitch = ((Double) arguments[2]).floatValue();
+        }
+
+        if (arguments.length > 3)
```
dan200: I would just remove this check. it's perfectly okay in Lua style to pass extra stuff on the end that isn't used

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,250 @@
...
+        {
+            throw new LuaException("Expected string, number (optional), number (optional)");
+        }
+
+        if (!(arguments[0] instanceof String)) // Arg wrong type
+        {
+            throw new LuaException("Expected string, number (optional), number (optional)");
+        }
+
+        if (arguments.length > 1)
```
dan200: See comments in playNote

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        if (arguments.length > 3)
+        {
+            throw new LuaException("Expected string, number (optional), number (optional)");
+        }
+
+        ResourceLocation resourceName = new ResourceLocation((String) arguments[0]);
+
+        if (m_clock - m_lastPlayTime >= TileSpeaker.MIN_TICKS_BETWEEN_SOUNDS || ((m_clock - m_lastPlayTime == 0) && (m_notesThisTick < ComputerCraft.maxNotesPerTick) && isNote))
```
dan200: You're still accessing a lot of variables in a thread  unsafe way here.. it might be fine, but it's not guaranteed

src/main/java/dan200/computercraft/shared/peripheral/speaker/TileSpeaker.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * This file is part of ComputerCraft - http://www.computercraft.info
+ * Copyright Daniel Ratcliffe, 2011-2016. Do not distribute without permission.
+ * Send enquiries to dratcliffe@gmail.com
+ */
+
+package dan200.computercraft.shared.peripheral.speaker;
+
+        import dan200.computercraft.api.peripheral.IPeripheral;
```
dan200: Bad whitespace

src/main/java/dan200/computercraft/ComputerCraft.java
Restioson: Yeah. I meant to change that. Sorry!

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
SquidDev: Slightly off topic, but what would your thoughts be on abstracting all argument validation out into a helper class? One thing which might be nice would be including the provided type, like standard Lua functions do (so `bad argument #1 (string expected, got nil)` instead of `Expected string, string`).

src/main/java/dan200/computercraft/shared/peripheral/speaker/SoundTask.java
Restioson: The issue with this is that I can't use a constructor in an anonymous class, which is currently how I pass the args initially.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
Restioson: Do you have any suggestion as to how I could improve this? I do not have much experience with threads in Java, but do have some in other languages. I generally steer clear of them

src/main/java/dan200/computercraft/shared/peripheral/speaker/TileSpeaker.java
Restioson: Odd, must've accidentally hit tab a few times

Restioson:
I'll try push fixes to @dan200 suggestions tomorrow, when I have time.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
Restioson: Should I used synchronized?

SquidDev: Yes. You can probably mark the entire method as `synchronized`.  I don't know if `TilePeirpheral.update` also needs to lock though.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SoundTask.java
dan200: I've never found a need to put an anonymous class on the constructor. Just "construct" things in the enclosing scope.

src/main/java/dan200/computercraft/shared/proxy/ComputerCraftProxyCommon.java
```diff
@@ -541,14 +555,15 @@ private void registerTileEntity( Class<? extends TileEntity> klass, String name
...
         // Tile Entities
-        registerTileEntity( TileComputer.class, "computer" );
-        registerTileEntity( TileDiskDrive.class, "diskdrive" );
-        registerTileEntity( TileWirelessModem.class, "wirelessmodem" );
-        registerTileEntity( TileMonitor.class, "monitor" );
-        registerTileEntity( TilePrinter.class, "ccprinter" );
-        registerTileEntity( TileCable.class, "wiredmodem" );
-        registerTileEntity( TileCommandComputer.class, "command_computer" );
-        registerTileEntity( TileAdvancedModem.class, "advanced_modem" );
+        GameRegistry.registerTileEntity( TileComputer.class, "computer" );
```
Restioson: I must have accidentally overwritten this in the merge.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
Restioson: I'll do it just to be safe.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
Restioson: To me, or to Dan? I would be happy to do this in my PR

dan200:
If you like, feel free to fix everything except the multithreading concerns, and i'll push a fix to them myself, might be easier than a long back and forth.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SoundTask.java
Restioson: Would I not need to make them final then?

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
SquidDev: This was directed at Dan. I realise I'm hijacking another thread (sorry), I just wanted to put out some feelers before creating a full blown issue/PR.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SoundTask.java
Restioson: Alternatively, I could use an inner class.

SquidDev: You'd have to make them `final`. For `volume` and `pitch` you can always create `final` copies of the original variable.

Restioson: That is true. I'll probably do that.

Restioson:
@dan200 I've ~~hopefully~~ fixed everything including the multithreading using `synchronized` methods. If this doesn't cut it, merge and push your own fixes 😄

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        if (!SoundEvent.REGISTRY.containsKey(new ResourceLocation("block.note." + arguments[0])))
+        {
+            throw new LuaException("Invalid instrument, \"" + arguments[0] + "\"!");
+        }
+
+        if (arguments.length > 1)
+        {
+            if (!(arguments[1] instanceof Double) && !(arguments[1] == null))   // Arg wrong type
```
SquidDev: To be really fussy, it should probably be `arguments[1] != null`. Similarly below. However, you could probably shift this up above so it's `if(argument.length > 1 && arguments[1] != null) { /* Do double check/volume assignment here */ }`.

Restioson: @SquidDev I agree with the first part, but to make it optional, I'd need to have another block for `arguments.length > 1 && arguments == null { /* Do default here */ }` whereas previously I could just use a ternary.

SquidDev: Isn't this OK? I might be misunderstanding you. It's pretty much what you have now but without the ternary.
```java
if(arguments.length > 1 && arguments[1] != null) {
    if(!(arguments[1] instanceof Number) {
        throw new LuaException("Expected string, number (optional), number (optional)");
    }
    volume = ((Number)arguments[1]).floatValue();
}
```

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -0,0 +1,264 @@
...
+            }
+            pitch = arguments[2] != null ? ((Double) arguments[2]).floatValue() : 1f;
+        }
+
+        // If the resource location for note block notes changes, this method call will need to be updated
+        Object[] returnValue = playSound(new Object[] {"block.note." + arguments[0], volume, Math.pow(2d, (pitch - 12) / 12d)}, context, true);
+
+        if (returnValue[0] instanceof Boolean && (Boolean) returnValue[0])
+        {
+            m_notesThisTick++;
```
dan200: This seems to only be incremented in playNote, not playSound. Surely this means playSound() isn't rate limited?

dan200:
Getting there! Just left a comment on one possible bug.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
Restioson: This is intentional. It is rate limited, once per tick. However, to allow for chords, playNote has 8 times per tick. Could be switched to both being 8 times per tick, but could have sound spam, whereas notes are short.

dan200: Ahh, I didn't notice the "isNote" check. Fair enough. Merging!

dan200:
This is ready to merge :) Before I do though, could you please add the change to the help system. "whatsnew" and "changelog" are the files to modify.

Restioson:
On an unrelated note, in `changelog`, it says "Pocket upgrde api", and not "upgrade".

SquidDev:
Oooooh, sorry about that. Apparently it doesn't matter how many times I check spelling, something will always get through. Sorry. On an equally unrelated note [there are a couple of other](https://github.com/dan200/ComputerCraft/commit/922818dfa532441f9dfa17eb872aa764653c0f32#commitcomment-22048615) changelog items which probably need clearing out.

dan200:
feel free to fix it

Restioson:
Yay!

apemanzilla:
Now that speakers are in, the `dj` program could probably use a revamp.

Restioson:
@apemanzilla 😛

Restioson:
Hey, it'd be cool if there was a sound file format, like MIDI but for CC. `ccmidi`? `ccmid`?

SquidDev:
There are several players for the [Note Block Studio](http://www.stuffbydavid.com/mcnbs) format, such as BombBloke's [Note](https://github.com/theoriginalbit/MoarPeripherals/wiki/Note) program\*. This is pretty much the defacto standard,

\* Yes, there are lots of others. This was the first one which sprung to mind so don't yell at me for not mentioning yours.

Restioson:
Hate to necro here, but we appear to have an issue: the lack of documentation on the wiki, or anywhere for that matter. I'll edit the PR with more info about the pitching and volume, etc, and then someone (or me, but I'd need to gain wiki editing privileges) can edit them.

SquidDev:
I've posted this on Discord, but just adding this here for consistency too: RE documentation, it might be worth having a read through https://github.com/dan200/ComputerCraft/issues/271 and the documentation related discussion in https://github.com/dan200/ComputerCraft/issues/343#issuecomment-326817125. Basically, no one is entirely sure whether we should put documentation on the wiki until we have a formal release or not.

PR #238, opened 2017-05-12
Refactor the filesystem and HTTP code

SquidDev:
This started as "add some methods" and kinda expanded into "shuffle lots of code around". The main reason for this refactor is to remove the duplicate file handle code between the `fs` and `http` API. I'm happy to revert some (or all) of those changes if desired.

## Refactoring changes
 - Move the encoding/decoding from the `Filesystem` implementation to the individual handles.
 - Move each handle into a `core.apis.handles` package from the main `fs` API.
 - Change the HTTP response to inherit from these handles.

## Functionality changes
 - Allow binary handles' `read` function to accept a number, specifying how many characters to read - these will be returned as a Lua string.
 - Add `readAll` to binary handles
 - Allow binary handles' `write` function to accept a string which is decoded into the individual bytes.
 - Add `binary` argument to `http.request` and friends in order to return a binary handle.

## Potential improvements/considerations
 - For binary HTTP requests, write the request body as raw bytes.
 - Remove the `accept-charset` for binary HTTP requests.
 - Change the "binary" flag for `http.request` to be a separate message. `http.reqeust` now takes 4 arguments, which is beginning to get a little ugly.
 - Shift everything to using nio `Channel`s. This would be really nice as it allows seeking. However, it would require changing the interface of `IMount`/`IWritableMount` so I'm not sure.


src/main/java/dan200/computercraft/core/lua/LuaJLuaMachine.java
```diff
@@ -509,6 +510,11 @@ else if( object instanceof String )
             String s = object.toString();
             return LuaValue.valueOf( s );
         }
+        else if( object instanceof byte[] )
+        {
+            byte[] b = (byte[]) object;
+            return LuaValue.valueOf( Arrays.copyOf( b, b.length ) );
```
dan200: Why the copy here?

SquidDev: In the rare case that the returner decides to mutate the value. `LuaValue.valueOf` doesn't copy it, so it seemed safer to do it here.

dan200: Fair enough, I'd assumed it would do it's own copy

SquidDev:
Great. Thank you very much!

dan200:
Just trying out the game now, and every time I try save a file (with "edit" or manually typing "fs" commands into lua), it creates the file but it ends up empty. This is the only code that touched file I/O, can you look into this?

PR #239, opened 2017-05-13
Make shell.programs filter .lua extension

Wilma456:
shell.programs doesn't show the .lua extension now,

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/shell
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/shell
@@ -180,7 +180,11 @@ function shell.programs( _bIncludeHidden )
 	-- Sort and return
 	local tItemList = {}
 	for sItem, b in pairs( tItems ) do
-		table.insert( tItemList, sItem )
+	    if sItem:find(".lua") == sItem:len()-3 then
+	        table.insert( tItemList, sItem:sub(1,-5) )
+            else
+	        table.insert( tItemList, sItem )
+	    end
 	end
 	table.sort( tItemList )
 	return tItemList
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/shell
```diff
@@ -180,7 +180,11 @@ function shell.programs( _bIncludeHidden )
 	-- Sort and return
 	local tItemList = {}
 	for sItem, b in pairs( tItems ) do
-		table.insert( tItemList, sItem )
+	    if sItem:find(".lua") == sItem:len()-3 then
```
SquidDev: This won't work for files with multiple `.lua`s in them. Instead, do `if sItem:find("%.lua$) then` or `if sItem:sub(-4, -1) == ".lua" then`.

I'd also add a check that the file without the `.lua` extension isn't also in the `tItems` set.

This bit of code uses tabs, so you might want to use them here too. I'd really recommend turning on "show invisibles/whitespace" in your editor so you can see this better.

Wilma456: What's the best code, to stop inserting items twice? I would use this code:
local tItemList = {}
local checkItem = {}
	for sItem, b in pairs( tItems ) do
	    if sItem:find(".lua") == sItem:len()-3 then
               if checkItem[sItem] ~= true then
	        table.insert( tItemList, sItem:sub(1,-5) )
                checkItem[sItem:sub(1,-5)] = true
             end
            else
                if checkItem[sItem] ~= true then
	          table.insert( tItemList, sItem )
                   checkItem[sItem] = true
                end
	    end
	end
	table.sort( tItemList )
return tItemList
Just for show and not tested yet. Is there a better solution than mine?

SquidDev: The above section already ensures items are only collected once, so you just need to check that the file without `.lua` isn't also in `tItems`.

PR #241, opened 2017-05-13
Update to 1.11.2

SquidDev:
I believe Llamas are the traditional sign of "look 1.11.2!":

A Llama with the new palettes

I haven't really got much to say to this - it's a pretty self-explanatory commit. It seems pretty stable, but I'm sure there are some issues.  I'd personally like to see this as a separate branch and have 1.10.2 stay as the primary version, merging changes into the 1.11.2 branch when appropriate. I realise this does introduce a little bit more work for everyone though...


KnightMiner:
I just start a 1.11 pack, so I'll be more than happy to test things out for that.

I do agree that separate branches makes more sense even if the 1.10.2 branch just freezes, just to make things easier for those developing addons for 1.10.2 or for potential maintenance.

Vexatos:
I have successfully been maintaining mods on five different minecraft versions at once. Merging things upwards is usually trivial.

SquidDev:
Ouch, 5 sounds painful. I think I've done 3 versions at a time, and it wasn't bad, it's just a bit of a faff when you come to release time :). Seeing as ComputerCraft doesn't do much item handling, I don't think juggling 1.10 and 1.11 would be very difficult (much easier than 1.7.10 and 1.8.9 say), but it also isn't a zero cost exercise.

Restioson:
+1

apemanzilla:
@Restioson there's a 👍 reaction for a reason

src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -155,21 +156,21 @@ public void shutdown()
         }
     }
 
-    private static IPocketUpgrade findUpgrade( ItemStack[] inv, int start, IPocketUpgrade previous )
+    private static IPocketUpgrade findUpgrade( NonNullList<ItemStack> inv, int start, IPocketUpgrade previous )
     {
-        for (int i = 0; i < inv.length; i++)
+        for( int i = 0; i < inv.size(); i++ )
```
JLLeitschuh: Change this to a list iterator? `for (ItemStack invStack : inv)`??

SquidDev: No. If you look further down I'm doing `ItemStack invStack = inv.get( (i + start) % inv.size() );` in order to start from the current slot and loop through the inventory.

JLLeitschuh:
The changes seem pretty consistent across the codebase. Can you throw a quick explanation of WHY these changes were necessary the way that they are? Did something internally change around how null was handled (thats what seems to have happened).

Really nice job scraping through the code and finding all of this stuff to update.

SquidDev:
@JLLeitschuh In Minecraft 1.11, Mojang switched from using `null` to represent empty `ItemStack`s to a special constant `ItemStack.EMPTY` instead. As `null` is no longer used, every use of it as an `ItemStack` needs to be replaced with `ItemStack.EMPTY`. This is slightly more than a find and replace job, as any stack with a count of 0 is considered "empty", and so `.isEmpty()` needs to be used instead of `== ItemStack.EMPTY`.

Most of the changes made in this PR are changing these uses of `null`, as well as adding additional annotations to aid analysis of nullness (nullility?, nullity?).

Wilma456:
Why this don't get merged?

SquidDev:
@Wilma456 To quote Dan from another issue:

>  Only thing holding up this one and your others is they're quite big and need a chunk of time to code review, and I've reviewed and merged 60 PRs in the last 2 weeks!

Dan has put a phenomenal amount of work into this repo in the last few weeks, but we must acknowledge that he has a job and a real life too - he can't look at everything instantly, especially large PRs like this one.

Personally I'd rather see this merged later rather than sooner - it would be nice to get a 1.10.2 release out before focusing on 1.11.2.

rdalkire:
Mr Squid Dev:

When I upgrade a computercraft sponge server from 1.8.9 to 1.11.2, using this new jar plus the one for juaj-jse, computercraft still seems to work in the sense that I can place a new computer, launch it and get a normal looking prompt.  

However, my preexisting computercraft items disappear, and in the forge log I see many lines like this:

`Replacing id <id> named as computercraft:<itemname> with air block. If the mod becomes available again later, it can reload here.`

Do you know of any tool or step-by-step technique that can find/replace those items? If not, do you have any quick off-hand conceptual design for how such a migration tool would work?  

Implementation or even the concept design itself might help remove a barrier for pre-established server admins to adopt the new version.

In any case thanks for all of your contribution sir!

SquidDev:
@rdalkire It should have handled the conversion automatically, but there was a little bug in the system. I've updated the PR with a fix - so blocks and items should no longer be destroyed.

If you're using Crazed's build server, it should show up in the next couple of hours. If you're building it yourself, then you obviously don't need to wait 😄.

rdalkire:
Very cool... however, and sorry if this is a dumb question but, when you say you've updated the PR, what do you mean?  I don't see any new commits.  Is there some other way to view these updates you speak of?

SquidDev:
@rdalkire I've rebased and modified the existing commits and pushed them, rather than creating a new commit. IMO it makes the PR look cleaner, though does make it harder to track changes through time. In this case, the only lines which changed [were these ones][1] - I'd forgotten that item names would also be lower cased by 1.11.2.

[1]: https://github.com/dan200/ComputerCraft/pull/241/commits/fb794461a488ff20596dc83f3ec369d7b22b47b2#diff-35bc80775629f3be7f30e01d8654a499R504

rdalkire:
Excellent @SquidDev I appreciate the detail you're providing.  

See the transparency, people? This shows-off the great potential advantage of open source projects. Yeah I know I'm preaching to the choir.

Anyway I plan to give this a try late tonight.  Thanks again!

Wojbie:
Just a bug report: redstone set functions are not causing redstone updates correctly in this PR. When called redstone/repeater/torch don't get updated and stay in old state. Any block update next to them will fix that. It works as expected when you are powering block - redstone connected to said block reacts correctly.

SquidDev:
@Wojbie Thanks for the report - should be fixed now.

Issue #51, opened 2015-12-13
The advanced drawing characters confilct with the border (just looks weird)


dirk1763:
The characters 128 to 159 split each character spot in a 2*3 grid that can be switched off and on individually. Except the bottom right pixel, but that can be archived by inverting the colors. But at the border of the computerc screen the background is bigger than the text area. So the inverted character looks strange.
![border](https://cloud.githubusercontent.com/assets/8234006/11768414/8e3837b6-a1cc-11e5-91e1-047d0b716865.PNG)

test script: http://pastebin.com/xT3U7PSp

dan200:
Not a lot I can do about this. The 2px margin between the border and the text was there from before character backgrounds could be changed, so to maintain backwards compatibility the decision was made to extend the background colour of the edge characters into the gap (otherwise, you'd just get a fixed black border, even if you'd fille the screen blue, for example). This behaviour is the same for the teletext characters as it is for any other.

BombBloke:
It strikes me that that a fixed black border may look better than this. Worth a try?

dan200:
It would look better for this use case, but not most others.

KnightMiner:
Maybe make it toggleable? That will allow you to have both advantages based on what you need

dan200:
toggleable where by whom?

KnightMiner:
As in a function which enables/disabled the automatic border (and maybe allows setting the color as well). Something like `term.forceScreenBorder( bool force, int color )`

dirk1763:
my real monitor has a (tiny) black border, too :smile: 
An other approach would be to strech the characters from 128 to 159 when they are printed on the border. This way there would not backwards compatibility problem.

dirk1763:
A third way to do it, would be that only the term.clear method sets the border colors and printing characters does not.

oddstr13:
A solution to this might be to remove the border completely, changing the terminal border texture slightly.

If not, I think I would prefer the border being solid black, rather than having it magically extend the background color. This would allow texture packs to either remove the black border, or leave it as is.

dan200:
I thought up a nice solution to this:

Allow the background colours of the border positions to be independently set by writing to positions 0 or width+1/height+1. term.clear() would still set all of these, and term.clearLine() would set the left and right border chars of that line, so the behaviour of many current programs will be unaffected. But the control is available to those who need it

It will take a while to implement this (especially making the window API play nice with it), so might not make it into the next beta, but we'll see.

[...]

PR #242, opened 2017-05-14
[WIP] Replace terminal margin with a black border

Lignum:
This replaces the current terminal margin with a black border, like this (as discussed in #51):

### DIFF
--- a/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
+++ b/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
@@ -430,6 +430,12 @@ public class WidgetTerminal extends Widget
                             palette
                     );
                 }
+
+                mc.getTextureManager().bindTexture( background );
+                Colour black = Colour.Black;
+                GlStateManager.color( black.getR(), black.getG(), black.getB(), 1.0f );
+                drawTexturedModalRect( startX, startY, 0, 0, m_leftMargin, getHeight() );
+                drawTexturedModalRect( startX + getWidth() - m_rightMargin, startY, 0, 0, m_rightMargin, getHeight() );
             } else
             {
                 // Draw a black background
--- a/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
+++ b/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
@@ -395,13 +395,14 @@ public class WidgetTerminal extends Widget
 
                 // Draw margins
                 TextBuffer emptyLine = new TextBuffer( ' ', tw );
+                TextBuffer emptyBgLine = new TextBuffer( 'f', tw );
                 if( m_topMargin > 0 )
                 {
-                    fontRenderer.drawString( emptyLine, x, startY, terminal.getTextColourLine( 0 ), terminal.getBackgroundColourLine( 0 ), m_leftMargin, m_rightMargin, greyscale, palette );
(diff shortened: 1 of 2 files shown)


Lignum:
Just realised that I forgot about monitors... and those will be problematic since there are no left/right margins, leaving blank space if I remove the pixel stretching. So I guess I'll mark this as a WIP.

Additionally, what do you guys think of a `term.setBorderColour` function?

apemanzilla:
I think I'd prefer to just have black borders rather than a method to change it.

lupus590:
does this mean that border pixel/chars are no longer slightly bigger than their non-border counterparts?

Lignum:
@lupus590 Yes

oeed:
I'm not sure I like this to be honest. I'd say leave it as is or reduce the size of the frame.

CrazedProgrammer:
I have no problems with the current way. It makes for some creative border design.

apemanzilla:
The original issue was that the special teletext characters don't work in a predictable fashion with the current way the border works, in #51.

Restioson:
> Kinda did the monitor margin

Kinda? 😛

Wojbie:
> Kinda? 😛

Top and bottom seem to be fixed, left and right not so much.

Restioson:
Hmm...

Lignum:
> Top and bottom seem to be fixed, left and right not so much.

Yeah, top and bottom are easy, since they're just background chars rendered with a black colour. Can't do that neatly on the left and right, I'd need to do it manually with the Tessellator, but the coordinates are really difficult to work with.

Wojbie:
OK its a total Bodge but couldn't you just edit textures for monitors and have the sides have extra 5 pixels of black?

Lignum:
@Wojbie Yeah, but resource packs would break.

Lignum:
Though, before I even bother fixing it, @dan200 What do you think of this? Would you merge it?

Wojbie:
> I'd need to do it manually with the Tessellator, but the coordinates are really difficult to work with.

Another bodge idea.. instead of doing it manually couldn't you just add 2 fake rows on right and left that are empty and squished to 2 pixels. Like that pixel stretching that was on 2 side rows but with extra rows and squished instead of stretched . Would that be easier?

KnightMiner:
Dan did mention on the old issue the idea of allowing the user to set the background color of those pixels individually using 0/WIDTH+1 as the coordinates if we wanted to do that for the monitors only make them real pixels (just small)

BombBloke:
[As he mentioned,](https://github.com/dan200/ComputerCraft/issues/51#issuecomment-164477970) the problem with allowing user control over borders is that it doesn't mesh well with "fake terminals" such as those produced by the window API. And I honestly can't think of a way to *make* it work in any satisfactory manner.

"All black, all the time" isn't perfect either, but I just can't see a solution that doesn't involve some amount of compromise, and that one seems to have the least to me.

lupus590:
What about changing the texture of the casing (gold in the [this](https://camo.githubusercontent.com/76a5a72f56e07e00957d4eafa833a288b0823de3/68747470733a2f2f7368697474792e646f776e6c6f61642f677358792e706e67) screenshot) so that it covers the black boarder on the terminal?

Restioson:
Again, resource packs.

dan200:
Not interested in this change, as I don't like how it effects the look of built in programs like multishell and paint. Especially as this PR admits it isn't functional for monitors yet.

PR #243, opened 2017-05-14
Replace reflection with access transformers

SquidDev:
This makes the code slightly neater and allows us to catch places where the variable/function has been renamed between Minecraft versions (such as `createStackedBlock`).

Worth noting that people may have to run `./setup.sh` again, as some IDEs won't pick up on the changes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,6 +63,12 @@ dependencies {
 
 }
 
+jar {
+    manifest {
+        attributes('FMLAT': 'computercraft_at.cfg')
+    }
+}
+
 processResources
 {
     // this will ensure that this task is redone when the versions change.
--- a/src/main/java/dan200/computercraft/shared/common/BlockGeneric.java
+++ b/src/main/java/dan200/computercraft/shared/common/BlockGeneric.java
@@ -134,7 +134,7 @@ public abstract class BlockGeneric extends Block implements
     }
 
     @Override
-    protected final ItemStack createStackedBlock( @Nonnull IBlockState state )
+    public final ItemStack createStackedBlock( @Nonnull IBlockState state )
     {
         return null;
     }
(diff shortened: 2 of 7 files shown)


KnightMiner:
I think ATs are also somewhat more efficient/faster as reflection is known to be I efficient.

SquidDev:
Whilst ATs are generally more efficient than reflection, I'm pretty sure on modern JVMs the cost isn't _that_ bad. Either way, none of this code is called in a performance critical region so it doesn't really matter.

SquidDev:
Thank you!

dan200:
Hmm. I'm not getting decompile error running setup.sh :/

dan200:
Lot's of error of this form:
20:07:02.863 [WARN] [org.gradle.api.internal.project.ant.AntLoggingAdapter] [ant:javac] /Users/Dan/Documents/code/Minecraft/ComputerCraft/build/tmp/recompileMc/sources/net/minecraft/block/BlockSilverfish.java:47: error: createStackedBlock(IBlockState) in BlockSilverfish cannot override createStackedBlock(IBlockState) in Block
20:07:02.864 [WARN] [org.gradle.api.internal.project.ant.AntLoggingAdapter] [ant:javac]     protected ItemStack createStackedBlock(IBlockState state)
20:07:02.864 [WARN] [org.gradle.api.internal.project.ant.AntLoggingAdapter] [ant:javac]                         ^
20:07:02.865 [WARN] [org.gradle.api.internal.project.ant.AntLoggingAdapter] [ant:javac]   attempting to assign weaker access privileges; was public

SquidDev:
Right, I'm really sorry about this. I was under the impression that ATs propagated to child classes (that is what my testing had indicated). However, after digging through the AT implementation, it appears they do not. I'd tested the compiling the jar, and running it an obfuscated and obfuscated environment but hadn't actually decompiled it, and therefore didn't notice this. I'm really, really sorry.

Whilst the two patches for records may be kept, I will need to revert the changes to (`Block.createStackedBlock`, as used by `TurtleCompareCommand`). I don't know if you want me to create another PR to do that or just do it yourself. If so, it's worth noting that the function name has also changed from `func_149644_j` to `func_180643_i`.

dan200:
PR please!

PR #250, opened 2017-05-17
Add .lua extension to all programs and startup (updated)

Wilma456:
@dan200 wants to merge #235, but there are conflicts and I don't know how to solve them. Here's the Version without conflicts.


dan200:
Alright! Let's see what this breaks.

PR #252, opened 2017-05-17
Remove sleep from shutdown and reboot

Wilma456:
Shutdwon now directly shutdown and reboot is noW directly rebooting without waiting.


dan200:
No thanks

PR #253, opened 2017-05-17
Fix files not writing data when closed

SquidDev:
As the raw stream was being provided to the parent class, buffered data was not written, resulting in empty files. This ensures the buffered reader/writer is the one which is closed.

In response to [this comment](https://github.com/dan200/ComputerCraft/pull/238#issuecomment-302175615)


dan200:
cheers!

PR #254, opened 2017-05-17
[WIP] Add credits file

ObloxCC:
This list should be updated weekly/biweekly and is meant to be merged before release


Vexatos:
No need. [GitHub already does it for you](https://github.com/dan200/ComputerCraft/graphs/contributors).

SquidDev:
@Vexatos I believe this was in response to [this comment](https://github.com/dan200/ComputerCraft/pull/215#issuecomment-302019291)

However, I think Dan wanted it added to [`rom/help/credits`](https://github.com/dan200/ComputerCraft/blob/master/src/main/resources/assets/computercraft/lua/rom/help/credits) rather than a whole new file. This way it's actually bundled with the mod.

dan200:
Yes, this is pointless. I meant adding a contributors section to the credits file (which we'll just copy from github)

PR #257, opened 2017-05-18
Add shell.parse()

Wilma456:
I added the function shell.parse.  This function allows you to parse arguments. Here's a few examples:
```
args,options = shell.parse("-abc,"example")
args is {"example"}
options is {a=true,b=true,c=true}
shell.parse("-gh","--hello=World","example","123")
args is {"example","123"}
options is {g=true,h=true,hello="World"}
shell.parse("--foo","bar")
args is {"bar"}
options is {foo=true}
```

Why I want to add this function to Computercarft?
I think many people would use this function and I'm planing to make a PR for list, that use this function.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
@@ -421,6 +421,31 @@ if multishell then
     end
 end
 
+function shell.parse(...)
+local params = {...}
+local args = {}
+local options = {}
+for key,data in ipairs(params) do
+  if data:find("--",1,true) == 1 then
+    data = data:sub(3,-1)
+    local head,body = data:match("([^=]+)=([^=]+)")
+    if body ~= nil then
+      options[head] = body
+    else
+      options[data] = true
+    end
+  elseif data:find("-") == 1 then
+    data = data:sub(2,-1)
+    for i=1,#data do
+      options[data:sub(i,i)] = true
+    end
+  else
+    table.insert(args,data)
+  end
+end
+return args,options
+end
+
 local tArgs = { ... }
 if #tArgs > 0 then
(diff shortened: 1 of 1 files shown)


SquidDev:
If we're going for this, I'd much rather have something which automatically checks usage, generates `--help`, etc... Maybe something like Python's [`argparse`](https://docs.python.org/3/library/argparse.html). I've got a pretty simple [Lisp implementation](https://github.com/SquidDev/urn/blob/master/lib/extra/argparse.lisp) of it which could be used as a reference implementation. There is also this [Lua argument library](https://github.com/mpeterv/argparse/blob/master/src/argparse.lua) but it is a bit overkill for CC.

I'm not sure whether this is too "Unixy" for CC though - most built-in CC programs don't use flags, preferring sub-commands instead.

Wilma456:
Maybe more people will use flags, if they can parse it easy with this function. I think my code is enough for CC. CC don't need more parsing.

SquidDev:
It isn't necessarily about providing more "parsing" (though all the above libraries do support an extended syntax), it's about providing better feedback for the user. For instance, consider this simple `ls` implementation:

```lua
local args, options = shell.parse(...)
local showHidden = options["a"]

-- TODO: Actually implement this
```

Obviously this works great for `ls -a` and `ls foo`, etc... However, what happens if someone tries `ls -b`? It'l l get swallowed up into `options`, instead of producing an error or including it in `args`. If you change your argument parser to require a list of possible flags and arguments, then you don't hit this issue.

dan200:
Not really a big fan of merging in APIs that none of the default programs actually use, and for simplicities sake I doubt I'd want any of the default programs to use this, so I think this is a no.
It sounds like an easy thing to distribute as a lib now we have require though :)

PR #258, opened 2017-05-18
Speaker max hearable range limit

Wojbie:
Limits max volume of speaker to volume of normal note block.

Due to way volume in playsound is implemented values that are multiple of 1 are extending range source can be heard from. Without this limit with value of 10000 speaker can be heard from 160000 blocks away. This would allow someone to spam whole server with noises.

This limits it to 48 blocks max (3) that is standard for note block.

I am unsure if this should be hard coded or a setting but i think hard coded in this case.


dan200:
Makes sense. We should also error on passing in a negative volume too. Could you look into if we need to do anything similar on pitch?

Wojbie:
No need to worry about pitch. Minecraft clamps it to 0.5 - 2.0.
Didn't think about negative volume.. oh well something to test later.

PR #260, opened 2017-05-19
Add fileextension for Paint

Wilma456:
I added a file extension for Paint: mcpi. It stands for "**Minecraft** **Computercraft** **Paint** **Image**". I have tried to find a extension that no other program use. If the user call paint, with a filename, who has no extension and this file does not exists, paint will add .mcpi to the filename. If you know better extension, who no other program use, let me know.


SquidDev:
From what I remember, the community has generally used `.nfp` as a file extension, in homage of [NPaintPro](http://www.computercraft.info/forums2/index.php?/topic/5029-145-npaintpro/). I feel it would probably be worth sticking with it.

You might also want to just use the same check as `edit.lua` does, which looks a little neater:

```lua
if not fs.exists( sPath ) and not string.find( sPath, "%." ) then
	sPath = sPath .. ".nfp"
end
```

Wilma456:
Don't know, that a extension already exits. I have change it.

dan200:
Does anybody know if "paint" images are 100% cross-compatible with NPaintPro images?

SquidDev:
I had a prod at the code and it looks like it. NPaintPro also supports another format (nft) which isn't compatible, but we don't need to worry about that.

I don't know if it would be worth adding a setting, like in #278?

dan200:
Doesn't seem necessary, but if one is added, i would like it to match the design I proposed in my latest comment over there!

Wilma456:
I will add a extension for it, when you merge my other PR, because I don't want conflict.

dan200:
The other PR is now merged. Please ensure your changes here match the style of that one *exactly* so we don't have to waste time going back and forth again.

Wilma456:
@dan200  Now done!

dan200:
Perfect

Issue #247, opened 2017-05-16
Change speaker playNote to take rightclick ticks instead of pitch


Cruor:
Using pitch for the `playNote` function is not intuitive unless you know music theory or by googling a chart. Counting the amount of times you have right clicked the block however is trivial, and is way more intuitive to use without having to look up details online. 

For example, if you want to play the note currently you would do `pitch = 2^(n / 12)` where `n` is the amount of steps relative to the base tune value (F#4). For example G#4 is  `pitch = 2^(2 / 12)`. Even with knowledge about music theory, this isn't completely foolproof, as the base tune value would be expected to be A4 and not F#4.

Here is a snippet in Lua to convert the amount of "ticks" on the note block to the correct pitch, however i believe this should be the default behavior of the function. `2^((tick % 25 - 12) / 12)` where % is the modulo operator.

dan200:
I'd be happy to merge this change

Cruor:
This is apparently the current behavior of the `playNote` function. I was going by the examples posted in the PR so my bad on that one. Going to PR a help file atleast.

PR #263, opened 2017-05-19
Speaker-related cleanup

Lignum:
This doesn't really change anything in terms of functionality. I've mainly cleaned up the speaker-related code I could find to be consistent with the project's formatting.


src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -127,38 +125,37 @@ public String getType()
...
+                throw new LuaException( "Expected string, number (optional), number (optional)" );
             }
             volume = arguments[1] != null ? ((Double) arguments[1]).floatValue() : 1f;
-
         }
 
-        if (arguments.length > 2)
+        if( arguments.length > 2 )
         {
             if (!(arguments[1] instanceof Double) && arguments[2] != null)  // Arg wrong type
```
Wojbie: I think this should say `(!(arguments[1] instanceof Double) && arguments[2] != null)`

Lignum: Yes, and the operands should be the other way around, too.

src/main/resources/assets/computercraft/lua/rom/help/speakers
```diff
@@ -2,7 +2,7 @@ The Speaker is a peripheral device available for CraftOS. Type "help peripheral"
 
 Methods exposed by the Speaker:
 playSound( sResourceName, nVolume, nPitch )
-playNote( sInstrumentName, nVolume, nTicks )
+playNote( sInstrumentName, nVolume, nPitch )
```
Wojbie: I think it was reffering to line 9 : 

>  Ticks is the amount of times a noteblock has been tuned (right clicked).

KnightMiner: As I recall, internally it uses the term pitch for the function in the code thus what lead to that change. According to #247 it goes by right clicks, so I assume this is correct and the function in the code just needs it's variable name changed to prevent confusion.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -55,32 +55,30 @@ public BlockPos getPos()
         return m_speaker.getPos();
     }
 
-    /* IPeripheral implementations */
+    /* IPeripheral implementation */
```
Restioson: Huh? Isn't there more than one implementation?

Restioson:
Looks good 👍

dan200:
Resolve the conflicts and I'll merge this.
Just for the record though: I'm not too bothered by tiny syntactic things like whether there are spaces around arguments or not, as long as the indentation is right it's all good.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
Lignum: No, you implement the entirety of `IPeripheral`.

Lignum:
Huh, I was able to merge that locally without any conflicts. I hope nothing exploded.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
Restioson: Oh, I get it

PR #264, opened 2017-05-20
Pocket.beep() implementation

Wojbie:
This PR adds limited basic functionality of speaker into pocket computers themselves as part of pocket API.

It allows pocket to play one harp (or in MC 1.12 or later bell) note per tick.

This allows for basic functionality of speaker for purpose of notifying player that they should look at pocket. Could be used for many useful applications like chat notices or job complete messages.

Based off Restioson's speaker code.


src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -150,6 +162,78 @@ public void shutdown()
...
+                    throw new LuaException( "Cannot find player" );
+                }
+				
+                ResourceLocation resourceName = new ResourceLocation( "block.note.bell" );
+                
+                //Fallback to harp if this version of minecraft don't have bell.
+                if ( !( SoundEvent.REGISTRY.containsKey( resourceName ) ) )
+                {
+                    resourceName = new ResourceLocation( "block.note.harp" );
+                }
```
Restioson: Note that the above block should be removed on port to 1.12

src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -150,6 +162,78 @@ public void shutdown()
...
+                if ( m_clock - m_lastBeepTime >= 1 ) //Once on tic only
```
Restioson: Just a note, but should be `tick`

src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -150,6 +162,78 @@ public void shutdown()
...
+                    {
+                        final EntityPlayer player = (EntityPlayer) m_computer.getEntity();
+                        final World world = m_computer.getWorld();
+                        final BlockPos pos = player.getPosition().up();
+                        final ResourceLocation resource = resourceName;
+                        final float vol = Math.min( volume, 1f );
+                        final float soundPitch = (float) Math.pow( 2d, (pitch - 12) / 12d );
+                        
+                        context.issueMainThreadTask( new ILuaTask() 
+						{
```
Restioson: Indentation here looks a bit... odd

src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -150,6 +162,78 @@ public void shutdown()
...
+							@Nullable
```
Restioson: And here

src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -150,6 +162,78 @@ public void shutdown()
...
+                
+                //Fallback to harp if this version of minecraft don't have bell.
+                if ( !( SoundEvent.REGISTRY.containsKey( resourceName ) ) )
+                {
+                    resourceName = new ResourceLocation( "block.note.harp" );
+                }
+                if ( m_clock - m_lastBeepTime >= 1 ) //Once on tic only
+                {
+                    if ( SoundEvent.REGISTRY.containsKey(resourceName) )
+                    {
```
Restioson: Legacy code from #237 -- you can safely remove this `if` as you are referencing a known to exist `ResourceLocation`

Restioson:
I've added some comments, but otherwise this PR looks good!

SquidDev:
I'd personally rather have a more synthetic sound, rather than using Minecraft's built-in noteblock sounds. Or maybe something like [the Windows ding](https://www.youtube.com/watch?v=LbSIE0kg_Aw) 😛. There is a bit of me which would quite like this to be accessible for all computers (`os.beep` or whatever), but that's rather hard without leaking Minecraft-specific code into the "core".

Restioson:
I think it's only really necessary to have `pocket.beep`, as normal computers can just have speakers. It is an option though

Wilma456:
Why not os.beep() for all Computers?

Wojbie:
> but that's rather hard without leaking Minecraft-specific code into the "core".

dan200:
This defeats the point of the noisy upgrade, no?

Wojbie:
Not really. Noisy update has a much bigger range and can play any possible sound. This adds just a watered down version of speaker to pocket for basic notification stuff. I would compare it like build in beeper in motherboard vs speakers. One is for notification stuff and other is to play music.

dan200:
Do you want this just because it's only possible to have one upgrade on pocket computers so you can't have a speaker and a modem? If so I'd just fix that instead

Wojbie:
While i do believe in multiple upgrades idea in this case its just me thinking about fact that we carry around multiple devices that are able to capture our attention with sound if needed. Like glucose meters or old PDA devices. I strongly believe that ability to make a basic sound for user to hear.
It would be like pocket having its own basic noteblock (not peripheral) like you you can place next to computer. In case of pocket its a functionality that you simply loose.

Restioson:
@dan200 it's more of a design choice to have only one peripheral. This is a lightweight and very useful feature imo.

Wojbie:
Update to limit usefulness of pocket.beep in playing music. This makes it more notification tool then sound tool.
Now can only play pitches 0,5,10,15,20,24 using integers from 0-5.

src/main/resources/assets/computercraft/lua/rom/help/pocket.txt
```diff
@@ -2,5 +2,7 @@ pocket is an API available on pocket computers, which allows modifying its upgra
 Functions in the pocket API:
 pocket.equipBack()
 pocket.unequipBack()
+pocket.beep( [nVolume] , [nPitch] )
 
-When equipping upgrades, it will search your inventory for a suitable upgrade, starting in the selected slot. If one cannot be found then it will check your offhand.
\ No newline at end of file
+When equipping upgrades, it will search your inventory for a suitable upgrade, starting in the selected slot. If one cannot be found then it will check your offhand.
+When using beep, build in speaker is limited to only one beep per ingame tick. Works similiary to Speaker peripheral exept nPitch accepts only numbers from 0-5 and plays only 6 tones instead of 25 allowed by speaker.
```
Lignum: except

SquidDev: Also "the built in speaker" rather than "build in speaker".

src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -209,7 +209,7 @@ public void shutdown()
                     final BlockPos pos = player.getPosition().up();
                     final ResourceLocation resource = resourceName;
                     final float vol = Math.min( volume, 1f );
-                    final float soundPitch = (float) Math.pow( 2d, (pitch - 12) / 12d );
+                    final float soundPitch = (float) Math.pow( 2d, (Math.floor(pitch)*5 - 12) / 12d );
```
Restioson: I don't understand what you've done here. The deleted code was right, surely?

dan200:
Sorry, not interested in this feature. I'd much rather people use Noisy Pocket Computers, and allow people to craft Wireless Noisy Pocket Computers if neceassary.

Wojbie:
Ok. Thanks for looking at it anyways. Cheers!

Wojbie:
Ok. Thanks for looking at it anyways. Cheers.

Oh well now is time to figure out more java to make Wireless Noisy Pocket Computers that possible.

PR #267, opened 2017-05-20
Shovel upgrade converts grass to path blocks

### LINKED ISSUE #127: [1.80pr0][Feature Request]Mining Turtle & Grass Path block.
This is not a bug, but feature request.
### Expected

Digging Turtle changes Grass Block into Grass Path Block with turtle.dig() function.
### Actual

The Turtle returns false in the above situation.
### Detail

From 1.9, "Grass Path" block is added. It can be created from Grass block by right clicking with Spade.
So, naturally, Digging Turtle should create Grass Path block from Grass block by dig() function as Farming Turtle can create Farmland block.

(And now, a player make Grass Path block change into Farmland block by right clicking with Hoe...)

SquidDev:
Using `turtle.dig()` with a shovel upgrade on a grass block will create a path block.

Closes #127

### DIFF
--- a/src/main/java/dan200/computercraft/shared/turtle/upgrades/TurtleShovel.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/upgrades/TurtleShovel.java
@@ -5,12 +5,22 @@
  */
 
 package dan200.computercraft.shared.turtle.upgrades;
+
+import dan200.computercraft.api.turtle.ITurtleAccess;
+import dan200.computercraft.api.turtle.TurtleCommandResult;
+import dan200.computercraft.api.turtle.TurtleSide;
+import dan200.computercraft.api.turtle.TurtleVerb;
+import dan200.computercraft.shared.turtle.core.TurtlePlaceCommand;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.item.Item;
-import net.minecraft.util.math.BlockPos;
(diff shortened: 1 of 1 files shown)


src/main/java/dan200/computercraft/shared/turtle/upgrades/TurtleShovel.java
```diff
@@ -41,4 +51,20 @@ protected boolean canBreakBlock( World world, BlockPos pos )
...
+
+    @Nonnull
+    @Override
+    public TurtleCommandResult useTool( @Nonnull ITurtleAccess turtle, @Nonnull TurtleSide side, @Nonnull TurtleVerb verb, @Nonnull EnumFacing direction )
+    {
+        if( verb == TurtleVerb.Dig )
+        {
+            ItemStack shovel = m_item.copy();
+            ItemStack remainder = TurtlePlaceCommand.deploy( shovel, turtle, direction, null, null );
+            if( remainder != shovel )
```
dan200: Is reference equality sufficient here? Would .equals be better?

SquidDev: Hrrmrm. It _should_ be, as the right click action for hoes/shovels should return the original stack. However it is possible that could change in the future. Would you like me to change `TurtleHoe`'s implementation as well?

dan200:
Ah, I didn't realise hoe did that too. I'm happy with this then.

PR #268, opened 2017-05-20
Add Contributors to credits.txt

Wilma456:


dan200:
Pick one username for yourself please :)

Wilma456:
@dan200 The problem is, that  I use "Wilma456" as name for the Internet. But Wilma456 was already in use in Minecraft, so I had to choose another name.

SquidDev:
I don't really think Minecraft usernames matter - I'm sure several people on that list don't use the same name in-game as on GitHub.

src/main/resources/assets/computercraft/lua/rom/help/credits.txt
```diff
@@ -8,3 +8,16 @@ Visit http://www.minecraftu.org and http://www.deepspace.me/space-cadets to find
 
 Join the ComputerCraft community online at http://www.computercraft.info
 Follow @DanTwoHundred on Twitter!
+
+GitHub Contributors:
```
Lignum: I think this should just say "Contributors". That'd leave room for people who didn't contribute directly via Git.

SquidDev:
On a slightly different note, should the "program contributors" section be update to include the authors of the other treasure disk programs? I don't know if they are credited elsewhere too.

Lignum:
Though, thinking about this now, this file should probably be created from the GitHub statistics every release. It would make organisation much easier, otherwise we'd have people PRing their names into the file, which already sounds like it'll go wrong.

Cruor:
Shouldn't technically Oddstr13 also be added to the file? As the PR was closed and @dan200 merged a slightly modified version that made it handle multiple shells better?

timia2109:
Why not add a pre-build program or script which load the contributors from the GitHub API and create this file?

jordansinn1994:
Who determines the order of the list? The list doesn't seem like that great of an idea, there is no representation of the amount of work someone has done.

src/main/resources/assets/computercraft/lua/rom/help/credits.txt
```diff
@@ -8,3 +8,16 @@ Visit http://www.minecraftu.org and http://www.deepspace.me/space-cadets to find
...
+SquidDev
+Lignum
+Restioson
+timia2109
+Wilma456 (Ingame:Jakob0815)
```
dan200: Please remove your in game name

src/main/resources/assets/computercraft/lua/rom/help/credits.txt
dan200: Please alphabetise the order of the names on this list

Wilma456:
@dan200 Done

dan200:
Cheers. I will merge this when we do a release, so that we remember to add anyone else who contributes

timia2109:
I create a small Java program which extract the contributors from the GitHub API. You can found it [here](https://github.com/timia2109/Contributor_CreditFile_Creator/releases/tag/v0.1). Feel free to use. Might add this jar as pre-build step.

Small using example ```Contributor_CreditFile_Creator.jar dan200/ComputerCraft ../ComputerCraft/src/main/resources/assets/computercraft/lua/rom/help/credits.txt .../path/to/pretext```

apemanzilla:
@timia2109 I feel that would be better implemented as a Gradle task, so that it doesn't need yet another binary.

dan200:
Would be happy to accept a PR that automatically updated this list during deployment (i'm imagining a constant in credits.txt that gets replaced like $MCVERSION et al), but i'll merge this for now.

PR #269, opened 2017-05-20
Update os.version()

Wilma456:
If you now run os.version(true), you get a number and not a string. Example Code:
```
if os.version(true) < 1.8  then
    print("This Program need features that has been added in 1.8")
    return
end
```


SquidDev:
I'd argue people shouldn't rely on `os.version` to get the current mod version, seeing as many "operating systems" override it. People should ideally parse the `_HOST` variable or just check for the presence of the functions they require.

CrazedProgrammer:
I use `_HOST` for checking the version of CC.

dan200:
Yeah, _HOST is a better thing to test if you care about cc version. But you
should really just check for the feature you need directly.

PR #272, opened 2017-05-21
[RFC] [Help Wanted] [Not Ready to Merge] BIOS changes to allow for easier custom OS making

lupus590:
Not sure where to best ask for help, so I'm making an early pull request.

A complete rewrite of the BIOS to allow for a custom OS to be specified per computer.

I want some advice on how to best do this. The new BIOS needs to be able to read the `settings` file at the root of the FS, which would best be done with the settings API. Loading APIs becomes a problem when that's in a new craftOS file which hasn't been loaded yet. And I would like to keep the BIOS as bare bones as possible. Advice?

The end result of this pull request should have no changes to the mods behaviour in its default setting.


SquidDev:
You definitely need to move the string metatable/environment protection back into `bios.lua`. I feel the Lua 5.3 stubs should probably also be part of `bios.lua`, though that is probably more up for debate.

I'd personally prefer to make this change on the Java side, as it means you can avoid mounting CraftOS's `rom` folder when using a custom OS - it gives it more of a "fresh start" feel.

With regards to loading the settings, you could always copy the bare minimal across (so just `textutils.unserialize`). You don't really need to load the whole `settings` API.

lupus590:
Learning Java is still on my todo list, so I can't do anything about the ROM mounting.

`textutils` is an API so that still has the same issue (although I could just copy the needed function)

lupus590:
Thoughts on functions like loadfile, dofile, print, write, etc.?

dan200:
Closing this as it isn't close to a usable PR. Feel free to continue the discussion in the comments, or make a new thread in suggestions.

dan200:
It seems to me a simpler way to achieve these aims would just be to allow bios.lua to be overridden in resource packs like all the other lua files. I would accept a PR which did this

lupus590:
One advantage over a resource pack bios that this will have is being able to have a different OS on different machines. Where as a new bios.lua makes changes server wide.

PR #277, opened 2017-05-22
Add Multishell Events

Wilma456:
I added 3 Events that fired, when something with multishell happen.


Wojbie:
I don't think those events are truly needed. They seem like something with rather small use case. Could you provide some context?

Wilma456:
I think this could be useful for a OS

Wojbie:
Then OS can add it by itself. Cause any good is will have multishell implemented internally as part of OS.

lupus590:
Don't all the tab ids change when one closes? I would think that would be a useful event, being able to have the updated IDs

A possible template for the event:
* event name: `multishell_tab_close`
* p1:  a table with keys as the old ID of a tab and the values as the new ID of that tab

apemanzilla:
What would be a use case where a program needs to know about multishell events? Ideally programs should run identically whether run via `shell` or `multishell`, and I can't think of any uses where knowing specific changes about the tabs would be useful.

KnightMiner:
Only case I ever need to know about multishell was because I hardcoded the screen size and needed to cut a row off the bottom if multishell was enabled.

dan200:
This doesn't seem useful, and also poses the question: *should* programs be informed about changes to other ones?

PR #278, opened 2017-05-22
Add setting for .lua extension

Wilma456:
edit will only add .lua extension, when edit.extension is true.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -817,8 +817,8 @@ end
 settings.set( "shell.allow_startup", true )
 settings.set( "shell.allow_disk_startup", (commands == nil) )
 settings.set( "shell.autocomplete", true )
-settings.set( "edit.default_extension", ".lua" )
-settings.set( "edit.extension", true )
+settings.set( "edit.autocomplete", true ) 
+settings.set( "edit.default_extension", "lua" )
 settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
 if term.isColour() then
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -818,6 +818,7 @@ settings.set( "shell.allow_startup", true )
 settings.set( "shell.allow_disk_startup", (commands == nil) )
 settings.set( "shell.autocomplete", true )
 settings.set( "edit.autocomplete", true )
+settings.set( "edit.extension", true )
 settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
 if term.isColour() then
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -817,7 +817,7 @@ end
 settings.set( "shell.allow_startup", true )
 settings.set( "shell.allow_disk_startup", (commands == nil) )
 settings.set( "shell.autocomplete", true )
-settings.set( "edit.autocomplete", true )
+settings.set( "edit.default_extension", ".lua" )
 settings.set( "edit.extension", true )
 settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -14,8 +14,10 @@ if fs.exists( sPath ) and fs.isDir( sPath ) then
 end
 
 -- Create .lua files by default
-if not fs.exists( sPath ) and not string.find( sPath, "%." ) then
-	sPath = sPath .. ".lua"
+if settings.get( "edit.extension" ) then
+   if not fs.exists( sPath ) and not string.find( sPath, "%." ) then
```
SquidDev: Just combine these into one if statement.

SquidDev: You're also using 3 spaces here and a tab character then 3 spaces on the next line. Seeing as the rest of the file uses 4 spaces, you should probably change this. 

This isn't the first time I've said this on your PRs, so can you please turn whitespace highlighting on in your editor - having a consistent whitespace makes it easier to read and edit the code in the future.

src/main/resources/assets/computercraft/lua/rom/help/settings.txt
```diff
@@ -11,6 +11,7 @@ Default Settings:
 shell.autocomplete - enables auto-completion in the Shell.
 lua.autocomplete - enables auto-completion in the Lua program.
 edit.autocomplete - enables auto-completion in the Edit program.
+edit.extension - edit will add .lua extension
```
SquidDev: Probably should read `edit will add .lua when no extension is specified` or something.

dan200:
May I propose a different design for this:

Call the setting "edit.default_extension"
Then, when attempting to edit a file which doesn't exist, and whose name does not already include an extension:
If the setting is nil, or the empty string, do nothing.
If the setting is any other string, append "." plus the string to the filename.
Obviously, "lua" would be the default value.

Wilma456:
@dan200  I have done this now

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -817,7 +817,8 @@ end
 settings.set( "shell.allow_startup", true )
 settings.set( "shell.allow_disk_startup", (commands == nil) )
 settings.set( "shell.autocomplete", true )
-settings.set( "edit.autocomplete", true )
+settings.set( "edit.default_extension", ".lua" )
```
dan200: This should be "lua" not ".lua"

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -817,7 +817,8 @@ end
...
+settings.set( "edit.extension", true )
```
dan200: Remove this setting. Extensionless behaviour should be indicated by setting default_extension to ""

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -15,7 +15,7 @@ end
 
 -- Create .lua files by default
 if not fs.exists( sPath ) and not string.find( sPath, "%." ) then
-	sPath = sPath .. ".lua"
+    sPath = sPath..settings.get("edit.default_extension")
```
dan200: Only do this line if the setting is non-empty, and add the dot automatically

Restioson: Also, whitespace between `h`, `..`, and `s`

dan200:
Will merge once above changes are made @Wilma456

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -817,7 +817,8 @@ end
 settings.set( "shell.allow_startup", true )
 settings.set( "shell.allow_disk_startup", (commands == nil) )
 settings.set( "shell.autocomplete", true )
-settings.set( "edit.autocomplete", true )
```
Wojbie: Wait a second why are you removing edit.autocomplete in this line? This is still a valid setting and should not be removed.

dan200: Can't believe I didn't notice this. What a mess

Wilma456:
@dan200  Now done!

dan200:
Once again we had to change every single line in what should of been a trivial commit... butt nevertheless this is ready to merge

PR #279, opened 2017-05-22
.lua extentions for treasure disks

lupus590:


dan200:
Merged. Though Treasure Disks have been deprecated and disabled by default for a few versions now and I may end up just removing them altogether.

PR #280, opened 2017-05-23
Add virtual peripherals

Wilma456:
I had add virtual peripherals. Examples for Usage:
-Use peripherals over remote
-Make a virtual speaker for a command computer, that use the command API
-Read the screen with a virtual monitor, without overwrite the term API
That just 3 examples. Maybe you have more.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
@@ -1,4 +1,6 @@
 local native = peripheral
+local internalcou = 1
+local perita = {}
 
 function getNames()
 	local tResults = {}
@@ -13,6 +15,9 @@ function getNames()
 			end
 		end
 	end
+    for k,v in pairs(perita) do
+        table.insert( tResults, k )
+    end
 	return tResults
 end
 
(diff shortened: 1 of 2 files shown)


Wojbie:
Honestly this sounds like something people put in API part of CC forums.
As this whole thing is pure Lua code and can be easily implemented using end user code i don't really see point of filling the environment with extra functions with rare use cases.

Wilma456:
But you have to overwrite the API to do this and I think overwriting the API is not a good coding style, but you have to do this too much in CC actually.

Wojbie:
Every single program that adds functionality to existing system has to overwrite it.
I just simply don't see usability of this feature for a new player and advanced player can easily implement his own variation of this on its own.

Wilma456:
I just want, that overwriting APIs is no longer needed in many usecases.

dan200:
As people have said, this doesn't seem a good fit for inclusion in CC: it complicates the peripheral API for newcomers, and provides nothing an advanced user couldn't add themselves.

This is better off as a user-distributed library, which luckily are very easy to distribute now we have require :)

CoolMan119:
If you really want that you can make the jar with your PR but it won't be on the CC repo.

dan200:
or just make an API and release it as a resource pack or a pastebin, as this is only a lua change

PR #282, opened 2017-05-24
Dynamically generate contributors list

apemanzilla:
The credits file will be automatically generated with names from the git history.


build.gradle
```diff
@@ -69,8 +71,42 @@ jar {
...
-{
+import org.ajoberstar.grgit.Grgit
+
+task contributors {
+    group = 'other'
+    description = 'Creates a contributors file based on the git commit history'
+
+
+    doLast {
+	def header = '''
```
dan200: Instead of deleting credits.txt and putting this text here, could we not just have credits.txt contain a ${CONTRIBUTORS} variable that gets replaced? Think of how version numbers are populated in mcmod.info

SquidDev:
Just wondering what effect this has on build time? I doubt it's going to be an issue, but I'm pretty sure this'll be run every build as you can't really do this incrementally.

apemanzilla:
@SquidDev the task completes instantaneously on my computer. I could set it up to track the `HEAD` location and only run when it's changed, but I'm not sure that's worth the effort.

SquidDev:
@apemanzilla That's fine. I just like to check as I've done some fancy stuff in Gradle files which have had resulted in noticeably longer build times.

apemanzilla:
@dan200 @SquidDev I've implemented both your suggestions. The `processResources` task now uses the template expansion similar to `mcmod.info`, negating the need for an extra task. Additionally, the task will now be cached appropriately and will be rerun if commit hash changes.

simon816:
What I've seen on other projects is just the inclusion of "contributors" under credits. Seems a bit unnecessary to list all people who've submitted code with git.

apemanzilla:
@simon816 People who've committed code via git are contributors. [And beside, why spend 10 minutes doing something manually when you can spend an hour automating it?](https://xkcd.com/1319/)

Restioson:
+1 for Gradle task

apemanzilla:
Thanks!

PR #284, opened 2017-05-24
Update os.run() and shell.run()

Wilma456:
If a Program, that you run with os.run() or shell.run() return anything, this functions will return it too.


SquidDev:
I'm really not sure it makes sense to return the result or error from `os.run` (and so `shell.run`). If people need it they should be using `loadfile` and `pcall`, after all `os.run` does all sorts of other stuff like printing the error to the console.

MineRobber9000:
I agree with SquidDev on this one. loadfile and pcall should be used to return a result or error.

dan200:
I'd be up with it returning true or false (or maybe even false, "error
message"), but no more than that. programs aren't functions

dan200:
A quick look at the code reveals it already returns success/fail. We don't need more than this.

PR #288, opened 2017-05-28
Fix to getPalletteColour

### LINKED ISSUE #287: Pallete getPaletteColour return number missmatch with orginals.
I noticed when creating a different version of multishell that returns from term.native().getPaletteColour() does not match numbers you provided it with.

Here is code that shows it in action. It should print 3 trues but it does not.

```lua
	c1,c2,c3 = math.random(),math.random(),math.random()
	term.native().setPaletteColor(colors.lime,c1,c2,c3)
	c4,c5,c6 = term.native().getPaletteColor(colors.lime)
	print( c1 == c4, c1 ,c4 , c1-c4)
	print( c2 == c5, c2 ,c5 , c1-c4)
	print( c3 == c6, c3 ,c6 , c1-c4)
```

It seems there is problem when number is translated from lua side to java side and back. Sadly i lack java skills and understanding how minecraft modding works to make it the fix myself.

Edit: I do realize this is related to floating point numbers accuracy.

Wojbie:
Done as @dan200 specified in #287. Changes way colors are stored from float to double. Closes #287.

### DIFF
--- a/src/main/java/dan200/computercraft/client/gui/FixedWidthFontRenderer.java
+++ b/src/main/java/dan200/computercraft/client/gui/FixedWidthFontRenderer.java
@@ -33,7 +33,7 @@ public class FixedWidthFontRenderer
         m_textureManager = textureManager;
     }
 
-    private static void greyscaleify( float[] rgb )
+    private static void greyscaleify( double[] rgb )
     {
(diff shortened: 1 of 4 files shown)


dan200:
Can you confirm this fixes the bug?

Wojbie:
Yes it fixes a bug.
Here is screenshot confirming it.

dan200:
great :)

PR #289, opened 2017-05-28
Add pocket computer light support to Speaker.

Wojbie:
This makes use of new pocket computer light access peripherals have and adds said functionality to speaker. If noisy pocket has made sound the pocket computer light will turn blue for a second. 

This showcases use of this feature to future peripheral makes and allows end user to quickly find which pocket in their inventory was making sound.


Restioson:
+1, this would be useful.

Wojbie:
Just adding visual aid what this adds to game: 
Pocket that made sound in last second:
on
Normal state of pocket:
off
Its a little thing but oh so nice. (And would be helpful for people with many pockets.

Restioson:
I think it could perhaps be dark blue, as this would   perhaps contrast more. Play around with the colours!

Wojbie:
@Restioson Here is my little test for all colors. Sadly main CC colors don't look nice for this use.

As for real dark blue here is a test using hex color 0x3320fc (not CC color).

What do you think?

Restioson:
@Wojbie I think the green or dark blue is nice

KnightMiner:
I agree that dark green is the most visible, but dark blue is pretty close and its a color that makes more sense for "is playing sound", as green makes me think "is turned on"

Wojbie:
Ok dark blue it is then. Did a squish while at it to not have 3 commits in 5 line PR.

hugeblank:
Slightly unrelated: 
Since this PR, has anyone thought of adding a method that allows the user to directly change the color of the light for the pocket computer? Would be useful for chat programs and the like.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
```diff
@@ -55,6 +55,11 @@ public BlockPos getPos()
         return m_speaker.getPos();
     }
 
+    public synchronized boolean madeSound()
+    {
+        return (m_clock - m_lastPlayTime <= 20) ;
```
Lignum: This doesn't need to be `synchronized` because we're exclusively reading from atomic fields. Additionally, I feel it'd be nicer to make that `madeSound(long ticks)` to get rid of that magical `20` constant.

src/main/java/dan200/computercraft/shared/pocket/peripherals/PocketSpeaker.java
```diff
@@ -74,6 +75,7 @@ else if ( entity != null )
                 speaker.setLocation( entity.getEntityWorld(), entity.posX, entity.posY, entity.posZ );
             }
             speaker.update();
+            access.setLight( speaker.madeSound() ? 0x3320fc : -1 );
```
Lignum: This is even more magical :P.

Wojbie: Well this is colors that this one is set to use. Dunno what is your problem with that.

src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java
Wojbie: You really love your functional programing don't you?
But i agree it would make sense to allow for some interesting interaction at later changes.

Lignum: > You really love your functional programing don't you?

Yes

src/main/java/dan200/computercraft/shared/pocket/peripherals/PocketSpeaker.java
Lignum: `0x3320fc` is a random number just floating around in here, usually you'd add a comment or make a constant like `DEFAULT_LIGHT_COLOUR`, but I suppose it's already quite self-evident in this case.

Wojbie: Plus this is exactly same way its coded in modem,

PR #292, opened 2017-05-29
Remove rightAlt from edit

Wilma456:
This is a fix for German user. If you want to write {,[,] or }, which is often needed in LUA, on a German Keyboard, you have to hold rightAlt, which open the menu. So it's not easy to write these Chars. I think, the two Ctrl Keys are enough. Why we need a 3rd Key?

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -679,7 +679,7 @@ while bRunning do
 
             end
 
-        elseif param == keys.leftCtrl or param == keys.rightCtrl or param == keys.rightAlt then
+        elseif param == keys.leftCtrl or param == keys.rightCtrl then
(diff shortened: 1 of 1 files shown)


SquidDev:
I think the `rightAlt` handling was originally added [to avoid this exact problem](http://www.computercraft.info/forums2/index.php?/topic/22289-). On my keyboard the rightmost alt key fires both `leftCtrl` and `rightAlt`, meaning that pressing it has no effect.

Cruor:
Just tested this on a Scandinavian keyboard layout (which had the problem previously) and it works just fine.

Lignum:
As dan says in the thread that SquidDev linked, this has been fixed and it has worked fine for me ever since the release that introduced the patch.

dan200:
yeah. a better fix would be to somehow solve CC getting the duplicate key events in the first place. But i suspect this is inherent to how MC works with LWJGL

PR #293, opened 2017-05-29
Fix generic-related compiler warning

Lignum:
Fixes this warning:
```
/home/lig/Projects/ComputerCraft/build/sources/main/java/dan200/computercraft/client/render/TurtleMultiModel.java:40: warning: [rawtypes] found raw type: List
        m_faceQuads = new List[6];
                          ^
  missing type arguments for generic class List<E>
  where E is a type-variable:
    E extends Object declared in interface List
/home/lig/Projects/ComputerCraft/build/sources/main/java/dan200/computercraft/client/render/TurtleMultiModel.java:40: warning: [unchecked] unchecked conversion
        m_faceQuads = new List[6];
                      ^
  required: List<BakedQuad>[]
  found:    List[]
```
This is because you seemingly can't create an array of objects with generic params, leading to an unchecked assignment. I've simply replaced this with a map, which IMO is much saner anyway.

### DIFF
--- a/src/main/java/dan200/computercraft/client/render/TurtleMultiModel.java
+++ b/src/main/java/dan200/computercraft/client/render/TurtleMultiModel.java
@@ -14,7 +14,9 @@ import javax.annotation.Nonnull;
 import javax.vecmath.Matrix4f;
 import javax.vecmath.Point3f;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


dan200:
Maps are slower to access than arrays, but I'll take that hit for a cleaner compile.

PR #295, opened 2017-05-30
Add Variables to Shell

Wilma456:
I add Variables to Shell. You can set them with shell.setenv(varname,content). To use Vars in the shell, run $varname. This is only the basic. There's no default Vars and you can only set Vars with a function and not in the shell. If this get merge, I will keep working.


src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -19,6 +19,7 @@ local function createShellEnv( sDir )
     local tEnv = {}
     tEnv[ "shell" ] = shell
     tEnv[ "multishell" ] = multishell
+    shellenv = {}
```
SquidDev: This should be declared outside this function, and should be local. You should probably have it be inherited from the parent shell too - like the the path and aliases.

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -114,7 +115,22 @@ else
...
-    local sPath = shell.resolveProgram( _sCommand )
+    local checkvar = { ... }
+    table.insert( checkvar, 1, _sCommand )
+    for k, v in ipairs( checkvar ) do
+        if v:find( "$", 1, true ) == 1 then
+            v = v:sub( 2, -1)
+            if type( shellenv[v] ) == "string" or type( shellenv[v] ) == "number" or type( shellenv[v] ) == "boolean" then
+                v = tostring( shellenv[v] )
+            elseif type( shellenv[v] ) == "function" then
+                v = tostring( shellenv[v]() )
```
SquidDev: I'm personally not a fan of this - environment variables should just be string -> string mappings, not arbitrary code.

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -114,7 +115,22 @@ else
 end
 
 local function run( _sCommand, ... )
-    local sPath = shell.resolveProgram( _sCommand )
+    local checkvar = { ... }
+    table.insert( checkvar, 1, _sCommand )
+    for k, v in ipairs( checkvar ) do
```
SquidDev: It would probably be neater if this was a separate function, rather than inserting and removing the current command from the `checkvar` table.

src/main/resources/assets/computercraft/lua/rom/programs/echo.lua
```diff
@@ -0,0 +1 @@
+print( ... )
```
SquidDev: This file probably shouldn't be here - it's useful for testing, but has been rejected before (#209).

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
Wilma456: The reason for this is, that the Vars may be changing after starting the shell. Like $time (if it get implemented). You don't have to run a updater for this Var in the background, when you have a function.

src/main/resources/assets/computercraft/lua/rom/programs/echo.lua
Wilma456: In the PR you're liked we don't have Vars in the Shell. In this PR we have Vars in the Shell.

SquidDev: The point still stands though - shell scripting isn't a thing and so it doesn't really serve much purpose.

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
SquidDev: I feel this would be more elegantly handled via [command substitution](http://tldp.org/LDP/abs/html/commandsub.html). So `foo $(time)` would execute the `time` command and pass its output as an argument to `foo`. If we go this route, then you can also implement the various pipes (`>`, `<`, `|`).

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
dan200: Yep, this is declaring a global!

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -114,7 +115,22 @@ else
...
+        if v:find( "$", 1, true ) == 1 then
```
dan200: "if v:sub(1,1) == "$" then"
would be more effecient

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -114,7 +115,22 @@ else
...
+            v = v:sub( 2, -1)
```
dan200: v:sub( 2 ) is equivalent to this

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -114,7 +115,22 @@ else
...
+            if type( tShellenv[v] ) == "string" or type( tShellenv[v] ) == "number" or type( tShellenv[v] ) == "boolean" then
+                v = tostring( tShellenv[v] )
+            elseif type( tShellenv[v] ) == "function" then
```
dan200: Allowing setenv() to be a function seem like a big can of worms. I'd rather you just restrict to the string/number/boolean

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -114,7 +115,22 @@ else
 end
 
 local function run( _sCommand, ... )
-    local sPath = shell.resolveProgram( _sCommand )
+    local checkvar = { ... }
+    table.insert( checkvar, 1, _sCommand )
```
dan200: Let's not let you use a variable as the command name, that's what aliases are for.

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -163,6 +180,19 @@ local function tokenise( ... )
 end
 
 -- Install shell API
+
+function shell.setenv( key, value )
+    tShellenv[key] = value
```
dan200: Check args here

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -163,6 +180,19 @@ local function tokenise( ... )
...
+end
+
+function shell.getenv( key )
+    if key == nil then
```
dan200: Check key is a string here, don't allow nil

src/main/resources/assets/computercraft/lua/rom/programs/echo.lua
dan200: If variables exist, the ability to echo them provides the ability to see their values, which is valuable.
However, that points out a big missing piece in this PR: Currently there's no way to set variables from the shell!

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
Wilma456: You want, that the user don't get a list of all Vars?

Wojbie:
What is the point of even adding this to CC. There is not shell scripting. I don't see usecase for this.

dan200:
Requested changes have not been made after 2 months. Closing PR.

PR #297, opened 2017-05-31
Fire Forge events where appropriate

### LINKED ISSUE #100: Fortune Ores, Turtles still get the Ore
I issued this on Fortune Ores Git [(Issue # 5)](https://github.com/DemoXinMC/FortuneOres2/issues/5) but basically got directed here, they said "ComputerCraft should be posting HarvestDropsEvent when breaking blocks."
Basically the issue is when the turtle mines an ore block it drops as the ore and not as a F.O. cluster, Other mining machines do work in the latest release of F.O. but even with the latest version of CC turtles don't get the clusters.
[See Turtle Slot One](http://imgur.com/AvIO5Vo)

### LINKED ISSUE #103: turtles can break blocks protected by rftools block protector [1.7.10]
A few days ago someone asked on the Computercraft forums about how to stop a turtle from mining into a base and break things and I suggested to test with RFtools.

Today I decided to test myself to make sure it worked and discovered that a block protected by a Block Protector can be destroyed by a mining turtle.

I asked this earlier today on rftools github and got the answer that: "As far as I know the fault is actually on ComputerCraft's side for not firing the appropriate BreakBlock event first"

This was tested on FTB Infinity Evolved 2.5.0

SquidDev:
- `BlockEvent.BreakEvent` and `BlockEvent.HarvestDropsEvent` are fired when digging (the former only being fired if `turtlesObeyBlockProtection` is true).
 - `AttackEntityEvent` is fired when attacking.
 - Various `PlayerInteractEvent.*` events are fired when placing.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/turtle/core/TurtlePlaceCommand.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/core/TurtlePlaceCommand.java
@@ -24,13 +24,16 @@ import net.minecraft.tileentity.TileEntity;
 import net.minecraft.tileentity.TileEntitySign;
 import net.minecraft.util.ActionResult;
 import net.minecraft.util.EnumActionResult;
+import net.minecraft.util.EnumFacing;
 import net.minecraft.util.EnumHand;
 import net.minecraft.util.math.BlockPos;
-import net.minecraft.util.text.TextComponentString;
-import net.minecraft.util.EnumFacing;
 import net.minecraft.util.math.Vec3d;
+import net.minecraft.util.text.TextComponentString;
 import net.minecraft.world.World;
 import net.minecraft.world.WorldServer;
+import net.minecraftforge.common.ForgeHooks;
+import net.minecraftforge.event.entity.player.PlayerInteractEvent;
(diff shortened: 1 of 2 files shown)


lurquhar1221:
Yes, please! This would help with sponge compatibility and the such, and would allow mods to better protect areas from turtles.

inklit:
This is pretty much required for proper Sponge compatibility.
Great to see this here.

dan200:
Has much testing been done on this with other mods? I worry there will be a lot that will assume the player is a real one and try to do things like open GUIs for them.

SquidDev:
I haven't tested how well this plays with others, but plenty of other mods fire the same events without issues. For instance, the same events will be fired by Thermal Expansion Autonomous activators, Extra Utilities mechanical users, Buildcraft quarries, etc... 

The `FakePlayer` class (which `TurtlePlayer` extends from) already handles people opening GUIs. Many mods will also include an `instanceof FakePlayer` check to ensure they don't do "actual" player specific code.

dan200:
Ah, that's reassuring then. Let's merge and see how it goes :)

SquidDev:
Great. Thanks!

PR #298, opened 2017-06-01
Replace BitAPI with a LuaJ implementation of bit32

SquidDev:
This removes CC's `bit` library and adds LuaJ's implementation of `bit32`, stubbing the original `bit` library.  This has several advantages:

 - Performance: as CC's conversion code isn't used, the `bit32` functions are significantly faster - my rather ad-hoc profiling shows a gain of somewhere between 1.5x and 2x.
 - Better "standard compliance": the LuaJ implementation of `bit32` now accepts variadic arguments and implements `extract` and `replace`.

I realise you lose the nice abstraction of `ILuaAPI`, but I don't feel that is a great loss - `bit` won't be needed should we ever switch to Lua 5.2/5.3 anyway. 

I'm happy to move this to the LuaJ directory and rebuild the LuaJ jar should you want.


dan200:
Idea seems good. How about moving this into our actual LuaJ sources?

SquidDev:
Sure. Would you also like me to rebuild the LuaJ jar then?

PR #299, opened 2017-06-02
Fix Bug with term.setCursorPos

Wilma456:
I fixed 2 Bugs:
1. If you call this function, without 2 numbers, you get a error in the window API and not in your Program
2. If you call, this function with 2 numbers lower then 1 (e.g. term.setCursorPos(0,0) ), CraftOS will hang forever and need to press Ctrl+R or rejoin the world.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/window.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/window.lua
@@ -243,6 +243,12 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     end
 
     function window.setCursorPos( x, y )
+        if type( x ) ~= "number" or type( y ) ~= "number" then
+            error( "Expected number, number", 2 )
+        end
+        if x < 1 or y < 1 then
+            error( "Position out of range", 2 )
+        end
         nCursorX = math.floor( x )
         nCursorY = math.floor( y )
         if bVisible then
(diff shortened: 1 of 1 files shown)


SquidDev:
@Wilma456 I couldn't manage to reproduce the hang with moving the cursor to 0, 0 - it's vanishes off screen but the computer still responds to input. Could you provide some sample code? Furthermore, I'm pretty sure you should be allowed to move your cursor outside the bounds - it would probably be better to fix whatever is causing the hang.

Wilma456:
Try it with term.setCursorPos(-10,-10). I just run that in the LUA Shell. Btw. I fond out, that if I press Ctrl+T longer than normal, I can kill the program.

SquidDev:
@Wilma456 Even -10, -10 doesn't block it - just press enter 11 times to get back to 1, 1.

src/main/resources/assets/computercraft/lua/rom/apis/window.lua
```diff
@@ -243,6 +243,12 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     end
 
     function window.setCursorPos( x, y )
+        if type( x ) ~= "number" or type( y ) ~= "number" then
+            error( "Expected number, number", 2 )
+        end
+        if x < 1 or y < 1 then
```
dan200: Remove this check. out of range cursors are fine

Wilma456:
This is a feature and not a bug? I had removed the second fix now.

PR #302, opened 2017-06-03
Switch wget to use binary mode.

Wojbie:
Allows wget to download all types of files not only text ones.

This makes use of changes SquidDev made to http and fs to make wget get all types files correctly.


xAnavrins:
Have you tested this, you sure it won't just save the first byte?

Wojbie:
Yes tested and retested. There were changes to http api and fs api in #238 that make this work.

Wojbie:
Opse.

SquidDev:
@xAnavrins Binary handles now have a `readAll` method, which this will use. The `write` method now accepts a string too, which will be saved as its byte representation.

PR #304, opened 2017-06-04
Add checks to window API

Wilma456:


src/main/resources/assets/computercraft/lua/rom/apis/window.lua
```diff
@@ -416,6 +425,9 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     end
 
     function window.reposition( nNewX, nNewY, nNewWidth, nNewHeight )
+        if type( nNewX ) ~= "number" or type( nNewY ) ~= "number" or type( nNewWidth ) ~= "number" or type( nNewWidth ) ~= "number" then
```
SquidDev: There are two checks for `nNewWidth`, instead of one for width and one for height. On a side note, it might be worth caching the `type` global, to reduce the performance impact.

PR #307, opened 2017-06-05
Add settings for Colours in some programs

Wilma456:
I'm just a friend of personalisation.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -814,28 +814,30 @@ if bAPIError then
 end
 
 -- Set default settings
-settings.set( "bios.use_multishell", true )
 settings.set( "shell.allow_startup", true )
 settings.set( "shell.allow_disk_startup", (commands == nil) )
 settings.set( "shell.autocomplete", true )
-settings.set( "shell.promptColor", "yellow" )
-settings.set( "shell.textColor", "white" )
-settings.set( "shell.backgroundColor", "black" )
+settings.set( "shell.prompt_color", "yellow" )
+settings.set( "shell.text_color", "white" )
+settings.set( "shell.background_color", "black" )
 settings.set( "edit.autocomplete", true ) 
 settings.set( "edit.default_extension", "lua" )
-settings.set( "edit.backgroundColor", "black" )
-settings.set( "edit.textColor", "white" )
-settings.set( "edit.highlightColor", "yellow" )
-settings.set( "edit.keywordColor", "yellow" )
-settings.set( "edit.commentColor", "green" )
-settings.set( "edit.stringColor", "red" )
+settings.set( "edit.background_color", "black" )
+settings.set( "edit.text_color", "white" )
+settings.set( "edit.highlight_color", "yellow" )
+settings.set( "edit.keyword_color", "yellow" )
+settings.set( "edit.comment_color", "green" )
+settings.set( "edit.string_color", "red" )
 settings.set( "paint.default_extension", "nfp" )
 settings.set( "lua.autocomplete", true )
-settings.set( "lua.promptColor", "yellow" )
-settings.set( "lua.textColor", "white" )
+settings.set( "lua.prompt_color", "yellow" )
+settings.set( "lua.text_color", "white" )
 settings.set( "list.show_hidden", false )
 settings.set( "list.file_color", "white" )
 settings.set( "list.directory_color", "green" )
+if term.isColour() then
+    settings.set( "bios.use_multishell", true )
+end
 
 if _CC_DEFAULT_SETTINGS then
     for sPair in string.gmatch( _CC_DEFAULT_SETTINGS, "[^,]+" ) do
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -824,6 +824,8 @@ settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
 if term.isColour() then
     settings.set( "bios.use_multishell", true )
+    settings.set( "list.file_color", "white" )
+    settings.set( "list.directory_color", "green" )
 end
 if _CC_DEFAULT_SETTINGS then
     for sPair in string.gmatch( _CC_DEFAULT_SETTINGS, "[^,]+" ) do
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -826,6 +826,17 @@ if term.isColour() then
     settings.set( "bios.use_multishell", true )
     settings.set( "list.file_color", "white" )
     settings.set( "list.directory_color", "green" )
+    settings.set( "edit.backgroundColor", "black" )
+    settings.set( "edit.textColor", "white" )
+    settings.set( "edit.highlightColor", "yellow" )
+    settings.set( "edit.keywordColor", "yellow" )
+    settings.set( "edit.commentColor", "green" )
+    settings.set( "edit.stringColor", "red" )
+    settings.set( "shell.promptColor", "yellow" )
+    settings.set( "shell.textColor", "white" )
+    settings.set( "shell.backgroundColor", "black" )
+    settings.set( "lua.promptColor", "yellow" )
+    settings.set( "lua.textColor", "white" )
 end
 if _CC_DEFAULT_SETTINGS then
     for sPair in string.gmatch( _CC_DEFAULT_SETTINGS, "[^,]+" ) do
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -814,30 +814,29 @@ if bAPIError then
 end
 
 -- Set default settings
+settings.set( "bios.use_multishell", true )
 settings.set( "shell.allow_startup", true )
 settings.set( "shell.allow_disk_startup", (commands == nil) )
 settings.set( "shell.autocomplete", true )
+settings.set( "shell.promptColor", "yellow" )
+settings.set( "shell.textColor", "white" )
+settings.set( "shell.backgroundColor", "black" )
 settings.set( "edit.autocomplete", true ) 
 settings.set( "edit.default_extension", "lua" )
+settings.set( "edit.backgroundColor", "black" )
+settings.set( "edit.textColor", "white" )
+settings.set( "edit.highlightColor", "yellow" )
+settings.set( "edit.keywordColor", "yellow" )
+settings.set( "edit.commentColor", "green" )
+settings.set( "edit.stringColor", "red" )
 settings.set( "paint.default_extension", "nfp" )
 settings.set( "lua.autocomplete", true )
+settings.set( "lua.promptColor", "yellow" )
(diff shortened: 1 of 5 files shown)


SquidDev:
Would it be sensible to extend this to syntax highlighting in `edit`? There might be some other places colour is used which could be "themed".

Wilma456:
@SquidDev I think yes, but I want to wait until this PR is close, because I don't wont a conflict.

SquidDev:
You can just add it to this one - turn it into a more general theming PR.

src/main/resources/assets/computercraft/lua/rom/programs/list.lua
```diff
@@ -27,7 +27,7 @@ table.sort( tDirs )
 table.sort( tFiles )
 
 if term.isColour() then
-    textutils.pagedTabulate( colors.green, tDirs, colors.white, tFiles )
+    textutils.pagedTabulate( colors[settings.get( "list.directory_color" )], tDirs, colors[settings.get( "list.file_color" )], tFiles )
```
Lignum: There's mighty error potential here, you should check what you index `colors` with.

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -824,6 +824,8 @@ settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
 if term.isColour() then
     settings.set( "bios.use_multishell", true )
+    settings.set( "list.file_color", "white" )
+    settings.set( "list.directory_color", "green" )
```
Lignum: Haven't tested it, but it doesn't look like this would work. Wouldn't this make it white & green every time, regardless of what the user sets?

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -824,6 +824,8 @@ settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
 if term.isColour() then
```
Lignum: This check no longer matters in relation to the colours, since greyscale terminals can now have their palette set to any greyscale colour, meaning you might still want to set your directory colour to green, since that could be set to a shade of grey.

src/main/resources/assets/computercraft/lua/rom/programs/list.lua
Wilma456: I think, that should the settings API do. I work on it.

Wilma456:
I add settings for edit,shell and lua

Wojbie:
Just a note here but with all this settings it would make sense to add pagination to set command.

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -826,6 +826,17 @@ if term.isColour() then
     settings.set( "bios.use_multishell", true )
     settings.set( "list.file_color", "white" )
     settings.set( "list.directory_color", "green" )
+    settings.set( "edit.backgroundColor", "black" )
```
SquidDev: Should these be `underscore_case` rather than `camelCase`? The other settings are, so it would be nice to be consistent.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -31,12 +31,12 @@ local bRunning = true
 -- Colours
 local highlightColour, keywordColour, commentColour, textColour, bgColour, stringColour
 if term.isColour() then
```
SquidDev: As @Lignum said before, we don't really need the `isColour` check any more.

src/main/resources/assets/computercraft/lua/bios.lua
Wilma456: The reason, why this is not underscore_case, is that the programs has her config for Colours in the code. I toke the names from there.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
Wilma456: Done

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -814,30 +814,29 @@ if bAPIError then
 end
 
 -- Set default settings
+settings.set( "bios.use_multishell", true )
```
SquidDev: `bios.use_multishell` should still be in the `term.isColour()` section.

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -814,17 +814,29 @@ if bAPIError then
...
+settings.set( "edit.backgroundColor", "black" )
+settings.set( "edit.textColor", "white" )
+settings.set( "edit.highlightColor", "yellow" )
+settings.set( "edit.keywordColor", "yellow" )
+settings.set( "edit.commentColor", "green" )
+settings.set( "edit.stringColor", "red" )
 settings.set( "paint.default_extension", "nfp" )
 settings.set( "lua.autocomplete", true )
+settings.set( "lua.promptColor", "yellow" )
+settings.set( "lua.textColor", "white" )
```
Wojbie: I really think we should move this to 'underscore_case'. I believe settings should be consistent with each other.

src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -104,9 +104,9 @@ end
 -- Colours
 local promptColour, textColour, bgColour
 if term.isColour() then
-    promptColour = colours.yellow
-    textColour = colours.white
-    bgColour = colours.black
+    promptColour = colors[ settings.get( "shell.prompt_color" ) ]
```
Wojbie: Now that I think about it there should be some fallback here. Cause if this setting is incorrect shell crashes and whole computer is unusable. I propose `promptColour = colors[ settings.get( "shell.prompt_color" ) ] or colours.yellow`. It would be nice to secure all stuff that takes color from settings like this but a shell is a must.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
SquidDev: Not quite: you've removed the check in the bios, but here and in the other programs you only load the colours from settings on an advanced computer - normal ones still use the default.

Wilma456:
Normal Computers don't have Colours, so you don't need this settings.

SquidDev:
@Wilma456 To quote @Lignum's earlier comment:

> This check no longer matters in relation to the colours, since greyscale terminals can now have their palette set to any greyscale colour, meaning you might still want to set your directory colour to green, since that could be set to a shade of grey.

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -817,14 +817,28 @@ end
...
+settings.set( "edit.keyword_color", "yellow" )
+settings.set( "edit.comment_color", "green" )
+settings.set( "edit.string_color", "red" )
 settings.set( "paint.default_extension", "nfp" )
 settings.set( "lua.autocomplete", true )
+settings.set( "lua.prompt_color", "yellow" )
+settings.set( "lua.text_color", "white" )
 settings.set( "list.show_hidden", false )
+settings.set( "list.file_color", "white" )
+settings.set( "list.directory_color", "green" )
```
xAnavrins: There's already a default colors in the edit and other programs
`colors[ settings.get( "edit.highlight_color" ) ] or colors.yellow`
and I feel like it would be better to not have these in the settings file by default.
If I don't want to customize themes, then it shouldn't bloat the settings file.
We could reference the settings names in their respective help files though.

Wojbie: Yes this would be nice way to reduce default .settings file size.

SquidDev: An alternative solution would be adding a way to provide default values, something like `settings.setDefault`. This would serve as a fallback for `settings.get` if no value is found (or default argument is provided). Obviously defaults wouldn't be saved to the `.settings` file, meaning it is less bloated.

Wojbie: Isn't what exactly same thing as what we achieve by using `colors[ settings.get( "edit.highlight_color" ) ] or colors.yellow`? I think xAnavrins and I are suggesting removing those default from bios.lua and instead documenting them inside help files. There is a fallback build in already into edit.

SquidDev: Wojbie: Yes and no. So anywhere `settings.get` is used, it'll still return "white" (or whatever) if nothing is configured. The only difference in functionality is that such values are not stored to `.settings` unless changed.

Wojbie: Ohhh so per example another program can check that and not get nil. That would be nice solution to problem.

dan200: One of the benefits of explicitly setting the settings at the start is that the "set" program gives you a nice list of everything that's customisable, without you having to dig about in source code to know about them

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
dan200: On both normal and advanced computers, it needs to load colours from setings, *but* the defaults need to be different for each! On a normal computer, I want my directory name to be white, not whatever shade of gray green comes out to.

dan200: So, if bios.lua is going to set colours, it needs to set them differently depending on Advanced or Normal

src/main/resources/assets/computercraft/lua/bios.lua
MineRobber9000: What about a `settings.getWithDefault(sSetting,vDefault)` (not sure about the naming tradition) that returns a default if the setting isn't set? I could use it in some of my programs.

Wojbie: getWithDefault of what exactly?

Wojbie:
Isint that already in settings.get? You should read wiki. If setting is not set then get will return its second argument.

src/main/resources/assets/computercraft/lua/bios.lua
Wojbie: Isint that already in settings.get? You should read wiki. If setting is not set then get will return its second argument.

MineRobber9000: I just checked and that is indeed what happens. I'm an idiot.

src/main/resources/assets/computercraft/lua/rom/programs/lua.lua
```diff
@@ -19,11 +19,13 @@ local tEnv = {
...
+else
+    term.setTextColour( colors[ settings.get( "lua.prompt_color" ) ] or colors.white)
 end
 print( "Interactive Lua prompt." )
 print( "Call exit() to exit." )
-term.setTextColour( colours.white )
+term.setTextColour( colors[ settings.get( "lua.text_color" ) ] or colors.white )
 
 while bRunning do
     --if term.isColour() then
```
MineRobber9000: This should be reimplemented (using `lua.prompt_color`, not sure why this was removed in the first place)

dan200:
The requested changes to this PR have not been made in 2 months. closing.

PR #308, opened 2017-06-05
move treasure disks to rom/programs/legacy

lupus590:


dan200:
why?

lupus590:
Just in case people wanted them, or because I could. I just thought of the idea and managed to implement it before thinking otherwise.

dan200:
Treasure disks haven't been removed from the mod.

lupus590:
>Merged. Though Treasure Disks have been deprecated and *disabled by default for a few versions now* and I may end up just removing them altogether.
#279

dan200:
Correct. They're disabled by default, not removed. Anyone who wants these
programs can just enable disks again.

PR #311, opened 2017-06-07
Add sources to os help

Wilma456:


SquidDev:
Perhaps it would be useful to give a brief description of each one? They are mostly self explanatory, but it wouldn't hurt.

oeed:
Good to have, but I think formatting needs work.

dan200:
This is something better documented on the CC wiki

Issue #83, opened 2016-01-13
[1.78] Phantom Peripherals


BombBloke:
Connect a peripheral to a computer via a pair of wired modems, and then break the peripheral. The computer will still act as if the peripheral is connected - eg, peripheral.wrap("printer_0") returns a table populated with usable functions even if the referenced printer no longer exists.

Peripherals placed directly next to computers before breaking do not trigger this. Breaking the modems (or "disconnecting" them) also doesn't seem to cause a problem.

Bug also existed under 1.76.

dan200:
According to SquidDev, this is fixed

SquidDev:
@dan200 Ahhh, sorry I may have confused things. #227 provides a commit (b39d765b3af178a0a9776a1345cd5bb59afca726) which fixes this, it just hasn't been merged yet.

PR #312, opened 2017-06-11
Allows placement of wired modem on any block that has peripheral.

Wojbie:
This was created to remove need for hacky solutions like Peripheral Proxy from use. Allows any wired modem to be placed on any block that has peripheral on selected side no matter what type of side is it. As side effect allows placing wired modem on turtles. Modem will break normally if block stops being peripheral or is moved/broken.

This feature would allow for non solid peripheral to be connected to wired networks.
In base mod would work only for turtles but for other mods that add peripherals to normal blocks like chests, doors or minecart tracks or for mods that want to use fancy block models with non solid sides this feature would allow them to bypass adding buggy stopgap measures like  Peripheral Proxy or Full block modems.

In image below in place of turtle imagine any non solid peripheral you ever used.

This still keeps rule about not placing wired modem on non solid surfaces. Only adds case so peripheral block get to bypass that rule.

Sidenote: Github got really confused about changes i made in TileCable.java due to added indentation level.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/peripheral/modem/TileCable.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/modem/TileCable.java
@@ -349,10 +349,10 @@ public class TileCable extends TileModemBase
         {
            
             EnumFacing dir = getDirection();
-            if( !worldObj.isSideSolid(
+            if( worldObj.isAirBlock(getPos().offset( dir )) || ( !worldObj.isSideSolid(
             getPos().offset( dir ),
             dir.getOpposite()
-            ) && PeripheralUtil.getPeripheral( worldObj, getPos().offset( dir ), dir.getOpposite() ) == null
+            ) && PeripheralUtil.getPeripheral( worldObj, getPos().offset( dir ), dir.getOpposite() ) == null )
             )
             {
                 switch( getPeripheralType() )
@@ -366,6 +366,12 @@ public class TileCable extends TileModemBase
                     }
                     case WiredModemWithCable:
                     {
+                        if( m_peripheralAccessAllowed == true ) //Make sure to disconnect peripheral when breaking
+                        {
+                            m_peripheralAccessAllowed = false;
+                            updateAnim(); 
+                            networkChanged();
+                        }
                         // Drop the modem and convert to cable
                         ((BlockGeneric)getBlockType()).dropItem( worldObj, getPos(), PeripheralItemFactory.create( PeripheralType.WiredModem, getLabel(), 1 ) );
                         setLabel( null );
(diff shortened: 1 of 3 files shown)


apemanzilla:
It looks a little strange, but I like it.

apemanzilla:
It looks a little strange with things like turtles, but I still think this would be a good change.

SquidDev:
The only potential issue I see here is when something is considered a peripheral on the server but not on the client (the old versions of OpenPeripheral used to be server only). Here you wouldn't be able to place a modem on a peripheral despite it is considered one by the server.

I don't think this is that big of an issue, just worth bearing in mind.

Wojbie:
All checks all server side only. Turtles per example are not considered peripherals on clients. On client side it just goes trough and then server decides if it wants to cancel that. That's why all those ``world.isRemote``s are there.

Wojbie:
Sorry for that.. Git accident.

Wojbie:
That not disconnection bug was in CC ever since we had modem peripherals ;D Never noticed it myself.

src/main/java/dan200/computercraft/shared/peripheral/modem/TileCable.java
```diff
@@ -349,12 +349,18 @@ public void onNeighbourChange()
...
             EnumFacing dir = getDirection();
-            if( !worldObj.isSideSolid(
+            if( worldObj.isAirBlock(getPos().offset( dir )) || ( !worldObj.isSideSolid(
             getPos().offset( dir ),
             dir.getOpposite()
-            ) && PeripheralUtil.getPeripheral( worldObj, getPos().offset( dir ), dir.getOpposite() ) == null
+            ) && PeripheralUtil.getPeripheral( worldObj, getPos().offset( dir ), dir.getOpposite() ) == null )
             )
             {
+                if( m_peripheralAccessAllowed == true ) //Make sure to disconnect peripheral when breaking
```
SquidDev: I'm pretty sure this can be shifted into the `WiredModemWithCable` section - normal `WiredModem`s cannot have a peripheral.

SquidDev:
For reference, the above bug is in #83, so that can be closed when this is merged.

src/main/java/dan200/computercraft/shared/peripheral/modem/TileCable.java
Wojbie: You are right. Moved to `WiredModemWithCable` section.

dan200:
This looks ugly. And connecting to turtles opens up other questions, what should the modems do when the turtle moves away? Break? Or stay hanging in mid-air? Both are bad answers imo.

Wojbie:
This causes them to break.
Hmm Was this closed only cause it looks ugly or You don't think this functionality is needed @dan200?
If first i will look into nice looking solution?

PR #313, opened 2017-06-11
HTTP blacklist and IP address support

### LINKED ISSUE #170: [Real world security] Http blacklist or similar
There are certain things you may not want your server users able to poke around with. For example, most routers are running on `192.168.1.1`. I really don't want people on my server able to poke around and access my router's admin console (I have a password but let's be honest here, how many people actually change the default password).

Many people running servers probably aren't using a hosting service and having the computer craft http module enabled currently gives players full access to the server's internal network.

Example:
```lua
http.get("http://192.168.1.1"):readAll()
```
This will dump the HTML of most server's router web interfaces.
It's a simple bit of code to go from here to tinkering with the server hosters router.

SquidDev:
There is a little bit more explanation of the changes in the commit messages, but the general gist is:

 - Adds support for blacklisting domains
 - Adds support for blacklisting & whitelisting IP addresses and IP ranges.
 - Reuse threads for HTTP requests

It is worth noting that the implementation now requires `http.checkURL` to yield. There is a `http.checkUrlAsync` method which will not yield but instead queue a `http_check` event.

Thanks to the new implementation, it is reasonably trivial to limit the number of concurrent HTTP requests (across the entire server, rather than per-computer like file handle limits). I don't know if this is something which would be worth implementing. 

I'm not entirely happy with my CIDR notation system, but it seems to work OK. Any comments welcome.

Closes #170.


Wojbie:
I do admit it would be nice to be able to have ``async`` version of ``checkURL``. Just so pepole don't have to mess with coroutines.

src/main/java/dan200/computercraft/ComputerCraft.java
```diff
@@ -106,7 +108,8 @@
 
     // Configuration options
     public static boolean http_enable = true;
-    public static String http_whitelist = "*";
+    public static AddressPredicate http_whitelist = new AddressPredicate( "*" );
+    public static AddressPredicate http_blacklist = new AddressPredicate( );
```
dan200: Wasn't the point of this feature so we could add localhost and it's various aliases by default?

src/main/java/dan200/computercraft/ComputerCraft.java
```diff
@@ -252,10 +256,34 @@ public void preInit( FMLPreInitializationEvent event )
...
+            if( currentProperty != null && !currentProperty.isList() )
+            {
+                Config.http_whitelist.setValues( currentProperty.getString().split( ";" ) );
+            }
+        }
+        Config.http_whitelist.setComment( "A list of wildcards for domains or IP ranges that can be accessed through the \"http\" API on Computers.\n" +
+            "Set this to \"*\" to access to the entire internet. Example: \"*.pastebin.com\" will restrict access to just subdomains of pastebin.com.\n" +
+            "You can use domain names (\"pastebin.com\"), wilcards (\"*.pastebin.com\") or CIDR notation (\"127.0.0.0/8\")." );
+
+        Config.http_blacklist = Config.config.get( Configuration.CATEGORY_GENERAL, "http_blacklist", new String[] {
```
dan200: Oh here it is. It seems error prone having this list in two places?

SquidDev: Ahhhh, missed that. I suppose I could extract the defaults into a separate string array?

dan200: Yeah, stick them up top somewhere, and construct the http_blacklist global using them.

SquidDev:
Done. That's a little neater :).

Issue #270, opened 2017-05-21
[Enhancement] More peripherals on pocket computers


KnightMiner:
As it has been brought up a couple times by dan with no real discussion, I feel it is worth bringing up the idea of allowing more peripherals on pocket computers before someone starts a PR for it (no point wasting effort if it is unwanted). I would say no more than two is required, as most of the time you are going to want a modem to communicate with other computers, then some peripheral unique to the program (speaker, sensor of some form, etc.). 

I could see the argument that limiting it brings out more creativity, but it feels much less natural to have to constantly swap the peripheral in order to perform more than one function which I feel really limits functionality.

----

While bringing that up, as turtles have two peripherals, which with this change would make the same as pocket computers, should they get a third one so they remain having more than pocket computers? (as they do seem to be more powerful computers) I feel a peripheral would fit well visually on the top, though tools might look awkward up there, but I am not sure turtles how much turtles need to have more peripherals than pocket computers.

Restioson:
Where would you put the second peripheral for the pocket?

KnightMiner:
Well, the current is "back", so probably top or bottom. Not sure how to show it visually unless we actually showed peripherals on the texture (quite a bit more complicated)

SquidDev:
One thing to consider is what to do about the "modem light" when both upgrades are trying to change it. A couple of possibilities:

 - Whoever changed it last determines the colour of the light. This seems the least appealing solution as modems set the modem colour every tick.
 - Split the current light into two pixels.
 - Determine the "average" colour of the two upgrades and use that.

---

I personally think two upgrades on turtles is enough. I feel there is always going to be a temptation to add more upgrade slots, just as they make things more convenient. Two allows for a tool and a modem, meaning you can communicate with other turtles and thus share responsibilities between them - something I think is more interesting than having some "all in one" turtle.

lupus590:
On the subject of turtles. Perhaps having them being able to use tools directly? Kind of like CCTweaks Tool Host but built into the turtle (and also how OC does its robots if I understand correctly).

Wojbie:
Could you explain this idea in more details?

lupus590:
@Wojbie Basically being able to have two peripherals and a tool(e.g. a pickaxe), with the tool being used 'from' the turtle's inventory.

SquidDev:
[...]

PR #314, opened 2017-06-11
Fix Pocket API working outside of player inventory

Iunius118:
This PR makes Pocket API not equip/unequip upgrades when the pocket computer is outside of the player inventory (e.g. dragging, dropped, placed in a chest).
This avoids duping/losing items of upgrade when they are equiped/unquiped.

This PR will fix these problems:

**Problem 1**
DESCRIPTION:
When a pocket computer is outside of the player inventory, the pocket computer fails to equip an upgrade and the upgrade item is lost.

EXPECTED RESULT:
The upgrade item should remain in the player inventory when the upgrade was failed to equip.

REPRODUCTION STEPS:
1. Have a pocket computer and an upgrade item
2. Run this code: `sleep(2) pocket.equipBack()`
3. Drop the pocket computer while sleeping
4. The pocket computer tries to equip the upgrade but fails

VIDEO: [fail_equip_and_lose_item](https://user-images.githubusercontent.com/18502177/27318325-af7bf80e-55c6-11e7-9751-77bda66121f3.gif)

**Problem 2**
DESCRIPTION:
When an upgraded pocket computer is outside of the player inventory, the pocket computer fails to unequip its upgrade and the upgrade item is duplicated.

EXPECTED RESULT:
The upgrade item should not be duplicated when the upgrade was failed to unequip.

REPRODUCTION STEPS:
1. Have an upgraded pocket computer
2. Run this code: `sleep(2) pocket.unequipBack()`
3. Drop the pocket computer while sleeping
4. The pocket computer tries to unequip but fails

VIDEO: [fail_unequip_and_dupe_item](https://user-images.githubusercontent.com/18502177/27319116-d2b32a78-55ca-11e7-8a49-5389259bc0b0.gif)

**EDIT**: Add a description about some problems I will fix by this PR.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/computer/core/ServerComputer.java
+++ b/src/main/java/dan200/computercraft/shared/computer/core/ServerComputer.java
@@ -101,11 +101,6 @@ public class ServerComputer extends ServerTerminal
         m_ticksSincePing = 0;
     }
 
-    public boolean isUpToDate()
-    {
-        return m_ticksSincePing == 0;
-    }
-
     public boolean hasTimedOut()
     {
         return m_ticksSincePing > 100;
--- a/src/main/java/dan200/computercraft/shared/computer/core/ServerComputer.java
+++ b/src/main/java/dan200/computercraft/shared/computer/core/ServerComputer.java
@@ -101,6 +101,11 @@ public class ServerComputer extends ServerTerminal
         m_ticksSincePing = 0;
     }
 
+    public boolean isUpToDate()
+    {
+        return m_ticksSincePing == 0;
+    }
+
     public boolean hasTimedOut()
     {
         return m_ticksSincePing > 100;
--- a/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
+++ b/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
@@ -13,6 +13,7 @@ import dan200.computercraft.api.lua.LuaException;
 import dan200.computercraft.api.pocket.IPocketUpgrade;
 import dan200.computercraft.core.apis.ILuaAPI;
 import dan200.computercraft.shared.pocket.core.PocketServerComputer;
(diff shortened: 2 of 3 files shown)


Iunius118:
I'm sorry for not writing a description about some problems I will fix by this PR.

SquidDev:
I wonder if it would be better to scan the inventory for the presence of the stack rather than using the timeout. My one worry is that depending on the timeout being 0 is depends on Lua tasks being ticked before the computer is ticked and that the inventory hasn't been changed since the last tick. The latter shouldn't occur in vanilla, but other mods could cause the player (for example) to drop all items in an event handler.

If you are going to stick with a timeout, it might be worth exposing it in `IPocketAccess` too.

Iunius118:
> The latter shouldn't occur in vanilla, but other mods could cause the player (for example) to drop all items in an event handler.

Sorry, I was not aware of this. It is exactly as you say. If a previous Lua task by other mods deletes the pocket computer stack, the pocket.equipBack() task will equip an upgrade to the deleted computer and decrease the upgrade stack.

> I wonder if it would be better to scan the inventory for the presence of the stack rather than using the timeout.

Yes, I will take your advice, thanks.

Iunius118:
I realized my PR was blocking development of pocket computer. So I will close this PR and give up. Now I think these problems I tried to fix would be better to be solved by the contributors developing pocket computer.

I’m sorry to disturb the project.

SquidDev:
> I realized my PR was blocking development of pocket computer.

If you're referring to #270, I wouldn't worry - that is something I'm not planning on working on for a while. Don't worry, you're not blocking anything :). Seeing as this is a pretty major bug, I think it would be worth reopening this.

> I’m sorry to disturb the project.

This is the last thing you need to be apologising for - you found an important bug and fixed it. That alone is great - thank you! Now we just need to wait for Dan to merge it :).

Iunius118:
Thanks, SquidDev.
However, I deleted forked repo and my branch, it's my bad. So this PR is dead because I can't fix my codes anymore.

PR #315, opened 2017-06-12
Rewrite argument validation

SquidDev:
This adds a utility class `ArgumentHelper`, which provides convenience methods for parsing arguments. This has the following advantages:

 - Makes argument parsing more consistent between functions. For instance, `nil` will allways fall back to the default value. Previously this was only the case for some methods.
 - Makes argument parsing simpler: all `instanceof` and array length checks are in the helper class, making argument validation much simpler.
 - Provides more useful error messages: we now generate errors in the same format as Lua. For instance, `bad argument #1 (expected string, got number)` instead of `Expected string`. This makes it easier for the end user to track down problems (and IMO is more consistent).

I've also added similar checks to the Lua code. I'm less sure about that - it was more to be consistent than anything else. This doesn't use any helper methods, the type checks are just inlined at the top of each function.

I've tested things pretty thoroughly, but I'd really appreciate others having a read through the code as well - just to ensure I haven't made a silly mistake.

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/ArgumentHelper.java
+++ b/src/main/java/dan200/computercraft/core/apis/ArgumentHelper.java
@@ -1,0 +1,246 @@
+package dan200.computercraft.core.apis;
+
+import dan200.computercraft.api.lua.LuaException;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Map;
+
+/**
+ * Various helpers for arguments
+ */
+public final class ArgumentHelper
+{
+    private ArgumentHelper()
+    {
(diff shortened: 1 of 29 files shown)


Wojbie:
Just a suggestion but if you are fixing up the error messages would you mind fixing capitalisation at 
[this](https://github.com/dan200/ComputerCraft/blob/master/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java#L560) and [that](https://github.com/dan200/ComputerCraft/blob/master/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java#L560) line? It is inconsistent with rest of error but don't seem like problem requiring its own PR.

PR #319, opened 2017-06-15
Add Default to read()

Wilma456:
Edit:
Here a example code. It shows how it can be used by a GUI.
```
term.clear()
term.setCursorPos(1,1)
print("Click to edit text. Exit with Ctrl+T.")
local text = ""
term.setCursorPos(1,3)
term.write("Text:")
while true do
local ev,me,x,y = os.pullEvent("mouse_click")
if y == 3 then
  term.clearLine()
  term.setCursorPos(1,3)
  term.write("Text:")
  text = read(nil,nil,nil,text)
end
end
```
Edit 2:
read

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -276,12 +276,17 @@ function printError( ... )
     end
 end
 
-function read( _sReplaceChar, _tHistory, _fnComplete )
+function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault )
     term.setCursorBlink( true )
 
-    local sLine = ""
+    local sLine
+    if type( _sDefault ) == "string" then
+        sLine = _sDefault
+    else
+        sLine = ""
+    end
     local nHistoryPos
-    local nPos = 0
+    local nPos = #sLine
     if _sReplaceChar then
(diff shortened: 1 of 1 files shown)


lupus590:
So what does this do? Does this mean that the 'text box' that read provides has data already inputted which the user can remove?

Wilma456:
@lupus590 Yes. The text from _sDefault is already in the the text box. And the User can edit them. In the current Version, you only can create a empty text box. You can add my code to the bios.lua (Open Computercraft.jar with a ZIP Program) and try my example or your own code. I think this is useful for GUIs. In RL the textbox don't get cleared when you click them.

dan200:
Read isn't really for "text boxes", it's for command line guis

SquidDev:
> Read isn't really for "text boxes", it's for command line guis

It could still be useful in that scenario. For instance, someone could create a REPL with allows multi-line input, with the appropriate indent being prepended automatically.

Wilma456:
@dan200 It is used as textbox in many programs, because it's the only function to read text in CC and may people don't want to write a function from scratch.

But if you want a example for the shell:
Without default:
```
print("Enter Path. Leave blank for default")
local path = read()
if path == "" then
   path = "/myprogram"
end
```
With default:
```
print("Enter Path")
local path = read(nil,nil,nil,"/myprogram")
```

PR #320, opened 2017-06-18
Fix Bug with peripheral.find()

Wilma456:
Since #315 peripheral.find() (and edit who use this function) doesn't work. This PR fix this.


SquidDev:
Ahhhh, really sorry about this 😦.

PR #321, opened 2017-06-18
Add Checks to Bios

Wilma456:


MineRobber9000:
I personally would make the messages more like usual CC error messages ("Expected string")

Wojbie:
@MineRobber9000 Error system changed few commits ago. This is new error setup.

PR #324, opened 2017-06-19
Fix Bug with peripheral.find()

BombBloke:
Since #315 peripheral.find() doesn't work. This PR fix this.


BombBloke:
I've also skimmed through all the other files #315 affected, and as far as I can see these are the only two negatively affected.

Wojbie:
Ok this time peripheral.find() is fixed. For sure!

Wilma456:
In #320 I only want to made edit work. I did not see this. Sorry.

Wojbie:
Hey don't worry @Wilma456. Whole point of open source is that many eyes beat pair of eyes.

SquidDev:
If anyone should be apologising it should be me - I was the one who wrote the code. Sorry!

PR #325, opened 2017-06-19
Add shell.loadAPI() and shell.unloadAPI()

Wilma456:
The difference to os.loadAPI() is that shell.loadAPI() load the API in the shell environment and not in _G. So you have access to the shell and multishell API and all global Vars in the Shell.


Wojbie:
I really don't see point of this function now that we have require.

BombBloke:
I don't see the point of these functions *without* require. For what purpose would you need this behaviour? For what purpose would you *want* this behaviour?

I suspect you're thinking of APIs as a way of spreading a single script across multiple files, whereas I would consider them to be a way of creating libraries multiple scripts can make use of independently.

dan200:
I'd rather not. loadAPI is for global scoped stuff, require is for program scoped stuff. I'd prefer people use require now we have it.

PR #326, opened 2017-06-20
Update whatsnew and changelog

JakobDev:


SquidDev:
As [mentioned here](https://github.com/dan200/ComputerCraft/commit/922818dfa532441f9dfa17eb872aa764653c0f32#commitcomment-22048615), we don't need to put every little change into the changelog. For instance, instead of the 5 "use xyz extensions on files" lines, you could just have "Add file extensions to various lua files and assets."

The whole changelog needs a clean up anyway, but Dan says he'll do that closer to the release.

PR #327, opened 2017-06-20
Add sources to time

Wilma456:


src/main/resources/assets/computercraft/lua/rom/programs/time.lua
```diff
@@ -1,3 +1,15 @@
...
-local nDay = os.day()
+local tArgs = { ... }
+local source
+if tArgs[1] == nil then
+    source = "ingame"
+else
+    if tArgs[1] == "ingame" or tArgs[1] == "local" or tArgs[1] == "utc" then
+        source = tArgs[1]
+    else
+        printError("Not a valid source")
```
lupus590: Perhaps you should list the valid options here.

dan200:
As the time sources feature is a fourth-wall-breaking feature, i'd rather keep it as a "behind the scenes" API feature for programs that need it, rather than part of the OS programs.

PR #329, opened 2017-06-22
Add equip and unequip for Pocket Computers

Wilma456:
We had this Programs for Turtles and now can Pocket Computers also use equip, so I add this programs for Pocket Computers.

The User shouldn't notice a different between equip for Pocket and equip for Turtles.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/help/equip.txt
+++ b/src/main/resources/assets/computercraft/lua/rom/help/equip.txt
@@ -1,4 +1,5 @@
-equip is a program for Turtles. equip will equip an item from the Turtle's inventory for use as a tool of peripheral.
+equip is a program for Turtles and Pocket Computer. equip will equip an item from the Turtle's inventory for use as a tool of peripheral. On a Pocket Computer you don't need to write a side.
 
 ex:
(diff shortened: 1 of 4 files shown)


SquidDev:
It might be worth also having a side argument, even if it is limited to "back". I'm planning to create a PR implementing #270 at some point, which will introduce another side.

Wilma456:
I will wait for your PR

dan200:
Merged. @SquidDev feel free to ammend this program in your pr

hugeblank:
@SquidDev Can I ask how the progress of the side argument is coming? I plan on making a sonar application that sends out a sound that gets faster the closer you get to another computer. While that's possible with what is already merged, just by equipping and unequipping, it would be much easier to just have both peripherals on the computer. :P

SquidDev:
@hugeblank I'm wanting to get #314/#331 fixed before doing any more work on pocket computers. However, I'm not sure which of the approaches outlined in the latter issue is the "best" one, and therefore am waiting for Dan to have a little bit more free time so that can be discussed.

hugeblank:
@SquidDev Absolutely understandable. Bug squashing before you add a feature that will consequently add more.

PR #330, opened 2017-06-22
Add Charlimit to read()

Wilma456:
I think this is useful for textboxes and it will make read() more realistic.

Example Code:
```
term.setBackgroundColor(colors.green)
term.setTextColor(colors.black)
term.clear()
paintutils.drawFilledBox(2,2,23,2,colors.blue)
term.setCursorPos(2,2)
term.write("Textbox without Limit:")
paintutils.drawFilledBox(2,3,23,3,colors.white)
paintutils.drawFilledBox(2,5,23,5,colors.blue)
term.setCursorPos(2,5)
term.write("Textbox with Limit:")
paintutils.drawFilledBox(2,6,23,6,colors.white)
term.setCursorPos(2,3)
read()
term.setCursorPos(2,6)
read(nil,nil,nil,nil,21)
term.setCursorPos(1,1)
term.setBackgroundColor(colors.black)
term.clear()
```

Running this code:
read

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -283,7 +283,7 @@ function printError( ... )
     end
 end
 
-function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault )
+function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault, _nCharlimit )
     if _sReplaceChar ~= nil and type( _sReplaceChar ) ~= "string" then
         error( "bad argument #1 (expected string, got " .. type( _sReplaceChar ) .. ")", 2 ) 
     end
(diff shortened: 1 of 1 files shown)


SquidDev:
IMO if people need text-box like behaviour then they should writer their own, rather than just relying on `read` (or if you must, create a new window and use `read` inside that). Otherwise you run the risk of having a mess of arguments. To quote Dan from #319:

> Read isn't really for "text boxes", it's for command lines

Cruor:
If you really want a `drawFilledBox` the function should be named `paintutils.rectangle` instead.
`paintutils.rectangle(mode, x, y, width, height, color)`, where mode is "fill" or "line".

BombBloke:
Using a window has an additional benefit in that it allows the text to scroll within the defined space, whereas this patch apparently doesn't.

Cruor, [drawFilledBox](http://www.computercraft.info/wiki/Paintutils.drawFilledBox) is already a part of CC, has been since v1.64. Same update added [drawBox](http://www.computercraft.info/wiki/Paintutils.drawBox).

apemanzilla:
With all these changes, why not instead add an additional function as part of the `textutils` API that, instead of taking a thousand arguments, takes a table of options? For example:
`local output = textutils.prompt({maxLength=10, default="hello"})`

Cruor:
@BombBloke my bad, was not aware of that. `textutils` is an abomination though, honestly.

dan200:
read() is for console input, not text boxes. Just write your own GUI library or "require" in someone elses

PR #331, opened 2017-06-22
Fix Pocket API working outside of player inventory

SquidDev:
This was cherry picked from #314 - see there for an explanation of the changes made and why.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
+++ b/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
@@ -13,6 +13,7 @@ import dan200.computercraft.api.lua.LuaException;
 import dan200.computercraft.api.pocket.IPocketUpgrade;
 import dan200.computercraft.core.apis.ILuaAPI;
 import dan200.computercraft.shared.pocket.core.PocketServerComputer;
+import dan200.computercraft.shared.pocket.items.ItemPocketComputer;
 import dan200.computercraft.shared.util.InventoryUtil;
 import dan200.computercraft.shared.util.WorldUtil;
 import net.minecraft.entity.player.EntityPlayer;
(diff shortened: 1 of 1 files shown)


Iunius118:
Thank you very much!

src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -78,11 +78,12 @@ public void shutdown()
                     {
                         if( !(m_computer.getEntity() instanceof EntityPlayer) )
                         {
-                            throw new LuaException( "Cannot find player" );
+                            return new Object[] { false, "Cannot find player" };
                         }
 
                         EntityPlayer player = (EntityPlayer) m_computer.getEntity();
                         InventoryPlayer inventory = player.inventory;
+                        if( !m_computer.inInventory() ) return new Object[] { false, "Not in an inventory" };
```
dan200: Why is this check necessary? Why does m_computer.getEntity() return the player when the player isn't holding the item anymore? That's what needs to be fixed.

SquidDev: That would be better, then I'm not entirely sure of the best course of action here. We can't really track when the pocket computer enters/exits a player's inventory - the only notification we get is the item's `update` method. 

Consequently, our only real strategy to determine whether the entity is valid is to scan the entire inventory for the pocket computer stack. We could add such a check to the `getEntity` method, returning `null` if it is not in our current entity's inventory. However, that does mean it is no longer thread-safe, which may be a problem in some situations.

Iunius118: How about finding the pocket computers not updated to set their private fields to `null` on `ServerTickEvent(Phase.END)`?

Iunius118:
This problem (pocket computer not held by a player isn't updated) may affect not only Pocket API but also pocket upgrade peripherals. They can't keep up to date if their `IPocketUpgrade.update()` isn't called.

SquidDev:
@Iunius118 Yeah, that is an issue I'm aware of. I'm not really sure of a feasible solution - I guess you could fire the `update` methods in the `ServerComputer` instead, but that isn't an ideal solution.

SquidDev:
@dan200 I've rebased onto master, which fixes issues introduced by the 1.11.2 transition. 

It's become apparent that the current proposed fix isn't ideal. I'm wondering if you have any thoughts on the above comments & questions, so we can find a better solution to these problems.

PR #332, opened 2017-06-23
Fix term.redirect

Wilma456:
If you call term.redirect with a table, who not contain all function, your Computer will Crash. This PR fix this. And I Update the error with missing table to the new style.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/term.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/term.lua
@@ -18,11 +18,11 @@ term.redirect = function( target )
         error( "term is not a recommended redirect target, try term.current() instead", 2 )
     end
 	for k,v in pairs( native ) do
-		if type( k ) == "string" and type( v ) == "function" then
-			if type( target[k] ) ~= "function" then
-				error( "Redirect object is missing method "..k..".", 2 )
-			end
-		end
+		if type( target[k] ) ~= "function" then
+				target[k] = function()
+					error( "Redirect object is missing method "..k..".", 2 )
+				end
+        end
 	end
 	local oldRedirectTarget = redirectTarget
 	redirectTarget = target
--- a/src/main/resources/assets/computercraft/lua/rom/apis/term.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/term.lua
@@ -12,7 +12,7 @@ local term = {}
 
 term.redirect = function( target )
 	if target == nil or type( target ) ~= "table" then
-		error( "Invalid redirect target", 2 )
+		error( "bad argument #1 (expected table, got " .. type( target ) .. ")", 2 )
 	end
     if target == term then
         error( "term is not a recommended redirect target, try term.current() instead", 2 )
@@ -20,9 +20,7 @@ term.redirect = function( target )
 	for k,v in pairs( native ) do
 		if type( k ) == "string" and type( v ) == "function" then
 			if type( target[k] ) ~= "function" then
-				target[k] = function()
-					error( "Redirect object is missing method "..k..".", 2 )
-				end
+				error( "Redirect object is missing method "..k..".", 2 )
 			end
 		end
 	end


src/main/resources/assets/computercraft/lua/rom/apis/term.lua
```diff
@@ -12,17 +12,15 @@ local term = {}
...
     if target == term then
         error( "term is not a recommended redirect target, try term.current() instead", 2 )
     end
 	for k,v in pairs( native ) do
 		if type( k ) == "string" and type( v ) == "function" then
 			if type( target[k] ) ~= "function" then
-				target[k] = function()
-					error( "Redirect object is missing method "..k..".", 2 )
-				end
+				error( "Redirect object is missing method "..k..".", 2 )
```
lupus590: Is it a goof idea to change how this works?

BombBloke: I don't think it is.

Current behaviour is backwards compatible with older buffers, which may not offer functions such as the relatively recent [term.blit()](http://www.computercraft.info/wiki/Term.blit). Since these older buffers tend to be used alongside older scripts which won't call that function anyway, they don't run into any problems.

This change would forcibly break them, and I don't see any gain.

Wilma456: @BombBloke  Have you any other way, how to prevent CraftOS from crashing, when this the table don't have all arguments?

Wojbie: I do believe cashing and erroring is correct solution for malformed terminal object/table

BombBloke: Yeah, don't redirect CraftOS to a terminal object unless you know it's suitable for the purpose.

All you're doing here is shifting the crash from a "might happen" to a "will happen" situation. If the terminal object isn't suitable, then *that's* the problem that needs to be addressed.

SquidDev: I don't think it's the end of the world if CraftOS crashes, as long as it prints out a message beforehand. Anyway, that's the worst case scenario - smarter programs (such as `monitor`) should catch the error and restore to the original term before printing the error message, which avoids such crashes.

lupus590: If CraftOS crashing is the issue, perhaps CraftOS should be modified to protect it from crashes caused by this?

dan200: Yep. We can't merge this change. For examples of code that would break with this change, look at how term.redirect is used by edit for printing, or textutils.pagedPrint

Wilma456:
I have a Idea: If a function missing, we could insert the old function, to prevent CraftOS from crashing. Are you like the idea?

Wojbie:
Wait wat? Insert old function from where? Could you explain in more details what are you suggesting here?

dan200:
Keep the old behaviour. Number 1 rule of CC is not to break existing worlds or programs.

Wilma456:
Changed back

src/main/resources/assets/computercraft/lua/rom/apis/term.lua
```diff
@@ -12,19 +12,17 @@ local term = {}
...
 term.redirect = function( target )
 	if target == nil or type( target ) ~= "table" then
-		error( "Invalid redirect target", 2 )
+		error( "bad argument #1 (expected table, got " .. type( target ) .. ")", 2 )
 	end
     if target == term then
         error( "term is not a recommended redirect target, try term.current() instead", 2 )
     end
 	for k,v in pairs( native ) do
-		if type( k ) == "string" and type( v ) == "function" then
```
Wojbie: Why is this check fur k == string removed? Is there a reason for it?

Wilma456:
I will make a new PR who only contains the table chek

PR #333, opened 2017-06-23
Better Style of Syntax error

Wilma456:
I think the new looks more like native LUA.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -19,7 +19,7 @@ if _VERSION == "Lua 5.1" then
                     end
                     return result
                 else
-                    return nil, err
+                    return nil, name..err:sub(25+#name,-1)
                 end
             else
                 local result, err = nativeload( x, name )
@@ -30,7 +30,7 @@ if _VERSION == "Lua 5.1" then
                     end
                     return result
                 else
-                    return nil, err
+                    return nil, name..err:sub(25+#name,-1)
                 end
             end
         end )
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -19,7 +19,11 @@ if _VERSION == "Lua 5.1" then
                     end
                     return result
                 else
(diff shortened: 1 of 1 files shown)


SquidDev:
You don't need the `-1` on the `sub` calls - Lua will default to the end of the string.

Worth noting that this is also fixed by #163 (/me does puppy eyes at @dan200).

SquidDev:
It's also worth noting that `name` can be nil, which will result in `lua` exiting.

Wilma456:
@dan200  Have you any example who name is nil, so I can look how to fix this?

SquidDev:
@Wilma456 You're allowed to use `load` - something like `load("return 1")` is legal. Trying to do `load("1")` crashes with `bios.lua:22: attempt to get length of nil`.

Wilma456:
@SquidDev Should be fixed now

SquidDev:
@Wilma456 Trying `load("x")` produces `nilg"]: '=' expected` on my machine, which is far from ideal :). 

> From my testing, `return false, err:sub(14)` is sufficient to strip the `bios.lua` element off. Just wondering why you went for the current method rather than this?

Wilma456:
`if err == nil then  return` should protect from this. Crazy.

Wojbie:
Why did you put read char limit into this PR?

Wilma456:
@Wojbie Sorry, this was in the same file and I had not see this. It is now removed.

Wojbie:
You forgot to push.

Wilma456:
No. In my Repository it is fixed. https://github.com/Wilma456/ComputerCraft-1/blob/syntaxerr/src/main/resources/assets/computercraft/lua/bios.lua But is not shown here. Maybe it's a Bug of Github. I wait a few hours.

SquidDev:
> `if err == nil then return` should protect from this. Crazy.

That won't do anything: `load` should never return a nil error message so that check is redundant. The issue is that your `:sub` code is invalid.

> From my testing, `return false, err:sub(14)` is sufficient to strip the `bios.lua` element off. Just wondering why you went for the current method rather than this?

dan200:
This is very hacky

Issue #90, opened 2016-02-26
[1.79] table.pack does not set n.


SquidDev:
`table.pack` should set the size of the table (as per the [Lua reference manual](http://www.lua.org/manual/5.3/manual.html#pdf-table.pack)), but does not 

```lua
table.pack = function( ... ) return { ... } end
```
Should be
```lua
table.pack = function( ... ) return {n=select('#',...), ... } end
```

dan200:
Fixed for 1.78pr1

PR #336, opened 2017-06-23
Removed the ~= nil checks from the bios checks. This will allow false…

Scoopta:
Removed the ~= nil checks from the bios checks. This will allow false to be used in place of nil. I make this change primarily because I have code that submits arguments with table.unpack and used false instead of nil. All of that code is now broken. Apparently my text editor also stripped extra spaces from the ends of stuff.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -9,7 +9,7 @@ if _VERSION == "Lua 5.1" then
         if mode ~= nil and mode ~= "t" then
             error( "Binary chunk loading prohibited", 2 )
         end
-        local ok, p1, p2 = pcall( function()        
+        local ok, p1, p2 = pcall( function()
             if type(x) == "string" then
                 local result, err = nativeloadstring( x, name )
                 if result then
@@ -38,7 +38,7 @@ if _VERSION == "Lua 5.1" then
             return p1, p2
         else
             error( p1, 2 )
-        end        
+        end
     end
     table.unpack = unpack
     table.pack = function( ... ) return { n = select( "#", ... ), ... } end
@@ -184,8 +184,8 @@ end
 
 -- Install globals
 function sleep( nTime )
-    if nTime ~= nil and type( nTime ) ~= "number" then
-        error( "bad argument #1 (expected number, got " .. type( nTime ) .. ")", 2 ) 
+    if nTime and type( nTime ) ~= "number" then
+        error( "bad argument #1 (expected number, got " .. type( nTime ) .. ")", 2 )
     end
     local timer = os.startTimer( nTime or 0 )
     repeat
@@ -194,13 +194,13 @@ function sleep( nTime )
 end
 
 function write( sText )
-    if sText ~= nil and type( sText ) ~= "string" and type( sText ) ~= "number" then
-        error( "bad argument #1 (expected string, got " .. type( sText ) .. ")", 2 ) 
+    if sText and type( sText ) ~= "string" and type( sText ) ~= "number" then
+        error( "bad argument #1 (expected string, got " .. type( sText ) .. ")", 2 )
     end
 
-    local w,h = term.getSize()        
+    local w,h = term.getSize()
     local x,y = term.getCursorPos()
-    
+
     local nLinesPrinted = 0
     local function newLine()
         if y + 1 <= h then
@@ -212,7 +212,7 @@ function write( sText )
         x, y = term.getCursorPos()
         nLinesPrinted = nLinesPrinted + 1
     end
-    
+
     -- Print the line with proper word wrapping
     while string.len(sText) > 0 do
         local whitespace = string.match( sText, "^[ \t]+" )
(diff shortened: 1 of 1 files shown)


Wojbie:
Well first thing is. Can you explain exactly how this change is important for CC other then not making you rewrite your code?

And for second qutestion. - Why are you using false instead of nil?

Scoopta:
I explained why I'm using false. I need to use tables to pack and pass variables. It's considerably easier than varargs in this particular segment of code and I can't put nil into a table. As far as why it's important? It shouldn't affect anything. It shouldn't break anything and the checks should behave identically to the way they do now. The only difference is you can pass false or nil instead of just nil. It's not particularly important but at the same time I don't really see a reason not to pull it.

Wojbie:
@Scoopta 
This about the meaning of values. 
When function expects a string and gets a boolean that means something went wrong. It got value it did not expect.
When function expects string and you gave it a nil that means you wanted to give it “absence of a value". This makes function assume default functionality there or error if there is no default functionality for that part.
I hope this explains my stance on this issue, I do believe even if this " It shouldn't break anything" its still a bad programing practice and not something we should be following.
Unless you have any other reasoning for this PR other then " I can't put nil into a table" then i will say its a bad code.
As for putting nils in table? `table.pack("One","Two",nil,"4")` will give you table with nil in 3th positions and total amount of arguments passed to in under `n` key. You could try using this method to construct your tables.

Scoopta:
@Wojbie that does not even come close to working.
`test = table.pack(1, 2, nil, 4)
table.unpack(test)`
I need that to be 1, 2, nil, 4...I get 1, 2. That's my point. I get your point that they mean two different things but nil is literally the only value that can be passed to mean "I don't want to replace." If false isn't any good then allow an empty string to mean "do not replace." As it is space can be used for replacing with nothing. I need to do read(table.unpack(_table)). I don't know how many arguments are given to me and they're not passed in a way where varargs would work and there's no real way to make varargs work.

Wojbie:
@Scoopta You need to use 2nd and 3th argument of unpack. By default it prints until first nil but you can tell it to print until specified number. Like key n that is given to you by table.pack function
``
a = table.pack(1, 2, nil, 4)
 print(table.unpack(a,1,a.n))
``
Does exactly what you are trying to achieve.

Scoopta:
@Wojbie That doesn't work in old versions of CC. Ugh version checking...oh well...it seems to work in newer versions. Chances are my code will break on some version somewhere because of this but whatever. I understand the false argument but that's still way cludgier than "" IMO but I guess you won't change your mind. I think I'mma go do my own read and try to remove my dependency on bios all together.

BombBloke:
"a.n or table.maxn(a)" should work under any version.

Scoopta:
@BombBloke I wasn't referring to a.n. I was referring to table.unpack. In the current official build table.unpack always stops at nil no matter what I specify. As I said...it doesn't really matter. I'm just not going to use read.

SquidDev:
@Scoopta `table.unpack` seems to work fine on 1.79. The only issue is that `print` stops on the first `nil` element.

```lua
local x = table.pack(1, 2, nil, 3)
local a, b, c, d = table.unpack(x, 1, x.n)
assert(d == 3)
```

Worth noting that `table.pack` was broken on earlier versions of the mod (pre 1.78), so you might want to include the patch in #90.

Scoopta:
@SquidDev ...that would explain...a lot. Well I'm already part way through the change anyway and one of the long term goals is to eventually remove bios as a dependency of this project anyway so it's not that big of a deal. I was probably going to have to do this anyway.

Scoopta:
@SquidDev I still don't personally see a problem with passing it false instead of nil. I only half see it as a problem of bad practice. In lua they more or less "mean" the same. I get the semantics but in an if statement there's no difference. That being said I'd be lying if I said I don't do ~= nil in my own code although that's really more because I come from a background of languages where nil/null is not false and cannot be treated as such and so it's just habit.

SquidDev:
>  I get the semantics but in an if statement there's no difference.

Whilst `nil` is falsey, it isn't the same as *false*. `nil` represents the absence of a value meanwhile `false` is a value. This is useful if you're trying to emulate default arguments in a function:

```lua
local function doTheThing(doIt)
    if doIt == nil then -- Nothing was given, use true as the default.
        doIt = true
    elseif type(doIt) ~= "boolean" then
        error("expected boolean", 2)
    end
    -- Do the thing!
end
``` 

> I come from a background of languages where nil/null is not false and cannot be treated as such and so it's just habit.

I'm not convinced that `nil` should be falsey, but then I'm a strong typing fanatic, so I'd want anything which isn't a boolean to produce a runtime error :). At least it isn't Javascript where you have `null`, `undefined` and `false`.

I think the killer argument here is that treating `false` equivalent to `nil` defines convention - the Lua standard library treats them differently. For instance, compare these two calls:

```lua
table.concat({"x", "y"}, nil) -- xy
table.concat({"x", "y"}, false) -- bad argument #2 to 'concat' (string expected, got boolean)
```

Whilst using `false` *may* work on *some* libraries, you're eventually going to hit problems somewhere else, so it's just better to do it "the right way".

PR #337, opened 2017-06-25
Fixed possible Typo.

Techman749:
Fixed a possible typo in the License file. 

It originally said: 
"All damages caused from the use or misuse of this mad fall on the user."
Not sure if this was intentional but I assumed you meant:
"All damages caused from the use or misuse of this may fall on the user."

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -36,7 +36,7 @@ This mod is provided 'as is' with no warranties, implied or otherwise. The owner
 of this mod takes no responsibility for any damages incurred from the use of
 this mod. This mod alters fundamental parts of the Minecraft game, parts of
 Minecraft may not work with this mod installed. All damages caused from the use
-or misuse of this may fall on the user.
+or misuse of this mod fall on the user.
 
 3. Play rights
 --------------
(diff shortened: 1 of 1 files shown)


Lignum:
This does appear to be a typo, however I think it's supposed to be "mod", not "may".

Techman749:
Yeah, I could see it being that way too. I assume @dan200 hand-wrote the license or modified an existing one and must've missed this part. I know this is a minor issue compared to other things, but a mispronounced license can be a potential problem. Especially if something bad does happen.

Wojbie:
I agree with @Lignum. This should be defiantly mod not may. It would make even more sense to use code instead off mod in this sentence.

LICENSE
```diff
@@ -36,7 +36,7 @@ This mod is provided 'as is' with no warranties, implied or otherwise. The owner
 of this mod takes no responsibility for any damages incurred from the use of
 this mod. This mod alters fundamental parts of the Minecraft game, parts of
 Minecraft may not work with this mod installed. All damages caused from the use
-or misuse of this mad fall on the user.
+or misuse of this may fall on the user.
```
lupus590: "mod" not "may" as others have already said

PR #338, opened 2017-06-26
Add Checks to Textutils API

Wilma456:


src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -149,10 +158,32 @@ local function tabulateCommon( bPaged, ... )
 end
 
 function tabulate( ... )
+    tArgs = { ... }
+    for i=1,#tArgs,2 do
```
SquidDev: You're creating a global here, which is never good. It might be better to use `select('#', ...)` and `select(i, ...)` instead - that way it won't break on `nil`s.

Wilma456: Just forget the local. Sorry. What did you mean with select()?

SquidDev: Instead of wrapping the varargs in the table and indexing it, you can use [`select`](https://www.lua.org/pil/5.2.html) to get the length and index it.

For instance, consider `tabulate("foo", "bar", nil, nil)` - this won't produce an error, as the length of the table is only 2. Using `select('#', ...) will return 4 - and so you'll error as expected. [`print` does something similar](https://github.com/dan200/ComputerCraft/blob/master/src/main/resources/assets/computercraft/lua/bios.lua#L262-L269).

BombBloke: The checks in tabulate / pagedTabulate could simply be performed into tabulateCommon with an error level of 3.

Moving them there would also allow use of [arg.n](https://www.lua.org/pil/5.2.html), avoiding an extra function call (you can only refer to either ... or arg within a vararg function: not both). Downside is that's broken after Lua 5.1.

Wojbie: Well BombBloke you can emulate arg.n thing in later version of lua by simply going `local arg = table.pack(...)`. This causes exactly same functionality as arg did.

BombBloke: My point was "let's make less calls", but it's not an issue while we remain on Lua 5.1 anyway.

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -149,10 +158,32 @@ local function tabulateCommon( bPaged, ... )
 end
 
 function tabulate( ... )
+    tArgs = { ... }
```
dan200: Please remove these checks, and the similar ones in pagedTabulate, and put checks in tabulateCommon with an error level of 3 instead.

Wilma456:
@dan200 Now Done!

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -99,6 +108,16 @@ end
 
 local function tabulateCommon( bPaged, ... )
     local tAll = { ... }
+    for i=1,#tAll,2 do
```
dan200: This is still wrong... the API doesn't require you to go "number, table, number, table". you can use any combination.

Wilma456:
@dan200  Should be fixed now

PR #339, opened 2017-06-26
Add colors.isValid()

Wilma456:
This function checks, if a colour is valid. It might be useful for adding checks to the APIs.

```
function example( nColor )
    if type(  nColor ) ~= "number" then
        error( "bad argument #1 (expected number, got " .. type( nColor ) .. ")", 2 )
    end
    if not colors.isVaild( nColor ) then
         error( "Invalid color", 2 )
    end
    --do something
end
```

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/colors.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/colors.lua
@@ -48,3 +48,18 @@ function rgb8( r, g, b )
         error( "Expected 1 or 3 numbers", 2 )
     end
 end
+
+local tColorcheck = {1}
+for i=1, 15 do
+    tColorcheck[2^i] = true
+end
+function isValid( color )
+    if type( color ) ~= "number" then
+        error( "bad argument #1 (expected number, got " .. type( color ) .. ")", 2 )
+    end
+    if tColorcheck[color] then
+        return true
+    else
+        return false
+    end
+end


src/main/resources/assets/computercraft/lua/rom/apis/colors.lua
```diff
@@ -48,3 +48,18 @@ function rgb8( r, g, b )
...
+end
+function isValid( color )
+    if type( color ) ~= "number" then
+        error( "bad argument #1 (expected number, got " .. type( color ) .. ")", 2 )
+    end
+    if tColorcheck[color] then
+        return true
+    else
+        return false
+    end
```
Wojbie: This whole if could be one `return tColorcheck[color] or false`

Wojbie:
By is valid you mean it's one of screen colors right? Cause colors can be combined.

src/main/resources/assets/computercraft/lua/rom/apis/colors.lua
SquidDev: Or `return tColorcheck[color] ~= nil`, which I find a little nicer :).

dan200:
Not a fan of adding a barely useful function that we don't even use.
Especially as calling code can just do the following, which is just as brief:
if type(colors[x]) == "number" then print( "x is a color" ) end

BombBloke:
That works if x is a string, though Wilma's function takes numbers and there aren't any numberic keys in the colours table.

Still, I see such checks being most commonly performed by display buffers, which are bound to have their own lookup tables localised anyways.

Wilma456:
@dan200 This doesn't works with numbers

PR #341, opened 2017-06-28
Added support for notifications in chat

CoolMan119:
As you might know in the new versions of CC there are new speakers. No programs that came with CraftOS used the speaker in any way. I have made it so the chat program can use speakers for notifications. In a chat you can toggle notifications with the `mute` command. It would be very useful so you don't miss a message if someone sends one.


CoolMan119:
Major errors, will try again later.

PR #342, opened 2017-06-28
Replace direct equality with InventoryUtil.areItemsEqual

### LINKED ISSUE #340: turtle.suck() and turtle.drop() give incorrect return values when an inventory is full
turtle.drop(), turtle.dropUp(), turtle.dropDown(), turtle.suck(), turtle.suckDown(), and turtle.suckUp() will sometimes give incorrect return values when putting items into an inventory.

- Inserting X into an item full of Y will result in "false, No space for items" (Expected)
- Inserting X into an item full of X will result in "true" (Unexpected) (No items were moved)

This appears to be the same for both turtle.suck() and turtle.drop().

SquidDev:
Some IItemHandler.insertItem implementations clone the item, so we must
check whether the object is equal instead. It would be possible to
compare stack sizes instead, I just feel this is a little clearer.

Fixes #340

### DIFF
--- a/src/main/java/dan200/computercraft/shared/util/InventoryUtil.java
+++ b/src/main/java/dan200/computercraft/shared/util/InventoryUtil.java
@@ -163,13 +163,13 @@ public class InventoryUtil
         }
 
         // Inspect the slots in order and try to find empty or stackable slots
-        ItemStack remainder = stack;
+        ItemStack remainder = stack.copy();
         for( int slot : slots )
         {
             if( remainder == null ) break;
             remainder = inventory.insertItem( slot, remainder, false );
         }
-        return remainder;
+        return areItemsEqual( stack, remainder ) ? stack : remainder;
     }
 
     private static ItemStack takeItems( int count, IItemHandler inventory, int[] slots )


dan200:
Could we do this check inside InventoryUtils.storeItems instead of every bit of calling code? I like the convention of returning the same ItemStack object to indicate no change

SquidDev:
Updated. Yeah, this was a little neater :).

src/main/java/dan200/computercraft/shared/util/InventoryUtil.java
```diff
@@ -169,7 +169,7 @@ private static ItemStack storeItems( ItemStack stack, IItemHandler inventory, in
             if( remainder == null ) break;
             remainder = inventory.insertItem( slot, remainder, false );
         }
-        return remainder;
+        return areItemsEqual( stack, remainder ) ? stack : remainder;
```
dan200: for extra safety, line 166 could be "remainder = stack.copy()", incase the code modifies the itemstack in-place

SquidDev: Hrrmrm, that does violate the contract of `insertMethod` (the docs say not to modify the stack), but better safe than sorry I guess :).

SquidDev: Done!

PR #344, opened 2017-06-29
Add Checks to disk,gps,help and keys

Wilma456:


src/main/resources/assets/computercraft/lua/rom/apis/disk.lua
```diff
@@ -1,5 +1,8 @@
 
 local function isDrive( name )
+    if type( name ) ~= "string" then
```
SquidDev: Is this really needed? `peripheral.getType` will error anyway.

Wilma456: Yes. Without this you get a error in the Disk API and not in your Program.

PR #348, opened 2017-07-03
Update term.redirect check to new style

Wilma456:


lupus590:
This is a lot better than #332

PR #350, opened 2017-07-04
Replaced "no value" with "nil" in ArgumentHelper

Wilma456:
If you call a function who is coded in LUA with a missing Argument, you get the error: "bad argument #1 (expected string, got nil)" if you call a function who is coded in Java with a missing Argument, you get the error: "bad argument #1 (expected string, got no value)". With this PR, the two error messages become the same look.


SquidDev:
I originally used `no value` as that is what PUC Lua uses. I'm not fussed either way, just explaining my rational.

> LUA 

Looks like a "Lua Uppercase Accident" :p.

PR #351, opened 2017-07-04
Fix to #349 - Monitors causing miscolouration of tile entities.

### LINKED ISSUE #349: Monitors causing miscolouration of tile entities
Hey - with a monitor in the field of vision, the glColor does not get reset before rendering other tile entities, leading to things like chests being broken.

Here, green text is being drawn on a monitor under me, and the chests are green as well.

image

And here, with the monitor out of frame, the chests return to normal colour.

image

I think this can easily be fixed by running a glColor3f(1, 1, 1) after you've rendered monitors.

Wojbie:
This fixes problem reported in #349.
I was one of people that were having this bug so i was able to test it and confirm this fix.
Thanks to @half-cambodian-hacker-man for diagnosing the issue cause and @SquidDev for this fix.

Closes #349

### DIFF
--- a/src/main/java/dan200/computercraft/client/render/TileEntityMonitorRenderer.java
+++ b/src/main/java/dan200/computercraft/client/render/TileEntityMonitorRenderer.java
@@ -177,6 +177,7 @@ public class TileEntityMonitorRenderer extends TileEntitySpecialRenderer<TileMon
                             }
                         }
                         GlStateManager.callList( origin.m_renderDisplayList );
+                        GlStateManager.resetColor();
 
                         // Draw text
                         mc.getTextureManager().bindTexture( FixedWidthFontRenderer.font );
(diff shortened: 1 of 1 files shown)


half-cambodian-hacker-man:
Have we tested this?

SquidDev:
@half-cambodian-hacker-man Wojbie says it fixes it on his machine. If you could double check, that would be great - you can [download a 1.10.2 version here](https://cc.crzd.me/maven/dan200/computercraft/ComputerCraft-Wojbie-ComputerCraft-Fix-to-349-Monitors-causing-miscolouration-of-tile-entities/1.80pr0-build0/ComputerCraft-Wojbie-ComputerCraft-Fix-to-349-Monitors-causing-miscolouration-of-tile-entities-1.80pr0-build0.jar).

Wojbie:
I did test it on both 1.10.2 and current. It works.

PR #355, opened 2017-07-06
Make pastebin.lua use HTTPS

MineRobber9000:
This PR makes `pastebin` use HTTPS for all requests. This makes file upload more secure<sup>[citation needed]</sup> and prevents relying on the HTTP->HTTPS redirect in the pastebin API.

Old PR message:

> Attempting to use pastebin with plain HTTP results in a 301 redirect to the HTTPS version. This PR makes the pastebin command work again.
> 
> Before this PR:
>  - `pastebin get` - works
>  - `pastebin run` - works (internally uses same pastebin file get method from `get`)
>  - `pastebin put` - fails
> 
> After this PR:
>  - `pastebin get` - works
>  - `pastebin run` - works (internally uses same pastebin file get method from `get`)
>  - `pastebin put` - works (theoretically, haven't had time to test it)

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
@@ -21,7 +21,7 @@ end
 local function get(paste)
     write( "Connecting to pastebin.com... " )
     local response = http.get(
-        "http://pastebin.com/raw/"..textutils.urlEncode( paste )
+        "https://pastebin.com/raw/"..textutils.urlEncode( paste )
     )
         
     if response then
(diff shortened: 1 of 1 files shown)


Wojbie:
When did that happen? It was working ok 9 hours ago?

MineRobber9000:
@Wojbie This simply makes `pastebin` follow the rules and request via HTTPS. (So I guess it's a "follow the access directions change")

Keanu73:
@MineRobber9000 Me and @Wojbie have re-tested this, and we have confirmed that `pastebin put` doesn't work still. And even **more** worrying, we have sent a raw HTTP request to the Pastebin API, and it has returned with the response 'Could not connect.'

ohno

MineRobber9000:
I'm leaving this open because `pastebin` *should* use HTTPS anyways. It doesn't conflict with #379 so I'll leave it here.

BombBloke:
Just to be clear as to where the "should" comes from: as of August 1st, pastebin's API will refuse all non-https connections.

https://pastebin.com/api

MineRobber9000:
@BombBloke so I was right in my assumption that pastebin would eventually quit redirecting HTTP API calls.

PR #358, opened 2017-07-07
Add shell.createEnviroment()

Wilma456:
This give other Programs access to the shell internal function createShellEnv(). This is useful, if you want to run a program in a sandbox and the shell Environment. I see no reason, why give other Programs not access to this function.

Edit:
Example Code:
Without this function:
```
local sandta = {}
sandta.shell = shell
sandta.multishell = multishell
sandta.require = require
sandta.package = package
sandta.fs = Table with sandboxed functions
os.run(sandta,program)
```
The Problem is, that the Program you run has no access no global Vars.

With this function:
```
sandta = shell.createEnviroment()
sandta.fs = Table with sandboxed functions
os.run(sandta,program)
```
The Program has access to all global Vars.

Another usecase is, that people can code functions like #325 self.


Wojbie:
Can you explain in detail how this function would be useful for someone? To make proper sandbox world require tons more code on top of this.

Wilma456:
@Wojbie I have edit the Post.

Wojbie:
So basically you want to add another function to shell to save writing 5 lines of code??

Wilma456:
@Wojbie Here is a example code:
Code of /print:
```
print(example)
```
Code of /run
```
example = "hello"
shell.run("/print")
os.run({},"/print")
```
If you run /run, you will get hello and nil, If os.run() share global Vars, you should get hello and hello.

Restioson:
Couldn't you write your own function to do this?

Wilma456:
No. I need Vars from the shell, which are local. dan200 has coded this function for shell.run() I only made it public, that other people can write their own functions.

Wojbie:
@Wilma456 this is expected. It's how it is supposed to work. If you want to write example to globals you need to go '_G.example = "hello"' . You will find your code works then.  I think you need to look exactly how environments work. You are making some false assumptions.
Sorry I can't be more help atm - on mobile.

Restioson:
Side note, this:
```lua
local sandta = {}
sandta.shell = shell
sandta.multishell = multishell
sandta.require = require
sandta.package = package
sandta.fs = Table with sandboxed functions
os.run(sandta,program)
```

can become:
```lua
local sandta = {shell = shell, mutlishell = multishell, require = require, package = package, fs = { Table with sandboxed functions } }
os.run(sandta, program)
```

BombBloke:
> The Problem is, that the Program you run has no access no global Vars. / The Program has access to all global Vars.

`createShellEnv()` doesn't return "the shell's environment table", it makes a *new* one - which *doesn't* contain the full list of globals that may've been in the previous environment, same as your first example there doesn't.

If you wanted multiple scripts to use the same environment tables (as shell.run() automatically arranged for you prior to [this commit](https://github.com/dan200/ComputerCraft/commit/61b2ed36a96c4c1977d39b9bc27c644f311a886c)), it seems to me that it'd be easier to use getfenv() to fetch the current one than it'd be to build a new one, especially if you want your other scripts to share with the one doing all the os.run()'ing.

dan200:
This seems a rare enough use-case that you could just copy createShellEnvironment into your program if you needed it.

PR #362, opened 2017-07-09
Command Event

KingofGamesYami:
Adds a `/computer <id> <msg>` command, which queues a `computer_command <msg>` event on the correct computer, if possible.

This allows certain things like text and book interfaces with computers.

I will gladly accept modifications restricting this to command computers only; I can't figure out how to make that work.


SquidDev:
Just a couple of suggestions:

 - Instead of storing a list of all computers and looping through them, just use `ComputerCraft.serverComputerRegistry.lookup`. This means you won't get memory leaks with the `Computer`.
 - `ServerComputer`s receive the `ComputerFamily` in the constructor. If you save this as a field and provide a getter, you can do ensure this only works for command computers.
 - It looks like you support passing multiple arguments in the computer event, so it might be worth adjusting the documentation to show that - I think Minecraft uses  `[arg]...` but I'll double check.
 - `Integer.valueOf` will throw an ugly exception if the ID isn't a number. You should catch the `NumberFormatException` and throw a `CommandException`.
 - Similarly, instead of printing an error message when the computer cannot be found, throw a `CommandException`.
 - I feel the `computer` event name is a bit too general - I'd stick with `command_computer`.
 - The command should probably override `CommandBase` instead of implementing `ICommand` - this way you get sensible defaults for methods like `compareTo` - your current implementation will possibly break sorting in the `/help` command.

BombBloke:
This sort of strikes me as being half a chat box (as implemented by various add-on mods since [MiscPeripherals](http://www.computercraft.info/forums2/index.php?/topic/4587-cc153mc152-miscperipherals-33/)) - it'd be great to have this functionality regardless, but would it maybe be better to add an official version of that classic peripheral at last?

KnightMiner:
One issue I see with this is a lack of a permissions check meaning it makes using a pocket computer as a remote less useful as it's harder to implement/more expensive. I think it would be good for command computers only, but make the ID optional letting the computer decide to handle it as otherwise you are telling users on your server to add an arbitrary number before commands which is never good to remember.

----

I have been thinking about how to bring chatboxes to cc vanilla in a way that prevents discouraging pocket computers. A few notes I have on how I would implement it:
* Always have a range limitation on the chatbox. For a base that is chat controlled, these can be periodically placed for full coverage (might want some message internal ID so the computer can check for duplicates)
* To still allow freedom, add a pocket variant that can receive messages while it's in the players inventory. The messages can be redirected using rednet to the intended computer or just run on the pocket computers itself (as a Siri like system). This needs a second peripheral on pocket computers.
* For the sake of servers, there will be a admin/command version that cannot be crafted but has no range limitation. This allows censoring/server commands/whatever else.
* I am not sure if sending chat messages as part of this makes more sense or if it would be better to move that to the speaker. Maybe rename the chatbox to a microphone, and add sound detection if it can be done cleanly/efficiently

SquidDev:
> Make the ID optional letting the computer decide to handle it as otherwise you are telling users on your server to add an arbitrary number before commands which is never good to remember.

@KnightMiner I got the impression from the PR that this is designed to be an "invisible" command, much like vanilla's [`/trigger`](http://minecraft.gamepedia.com/Commands#trigger). It would only be used inside [`clickEvent`](http://minecraft.gamepedia.com/Commands#Raw_JSON_text)s (such as the output of `/tellraw` or in written books). 

An example would be in Taken - I have the ability to switch the player into a "spectator mode", teleporting them back to where they started when they press a button. This would normally be implemented by the `/trigger` command and polling the scoreboard every tick. Here we could just wait for an event.

Chat boxes would be also nice, but do serve a separate purpose - they designed to be used by the average user for more "public" interactions with a computer.

KnightMiner:
Fair enough, though you would still have to ensure your computers ID does not change or all your tellraw needs to be updated. Probably could be a local variable if all commands are run in one computer.

KingofGamesYami:
@SquidDev I've implemented (at least locally) everything in your first post
@BombBloke @KnightMiner Yeah this is pretty much a replacement for `/trigger` like @SquidDev said
I did figure out the appropriate permissions level for that now.

src/main/java/dan200/computercraft/ComputerCommand.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public void execute(MinecraftServer server, ICommandSender sender, String[] args) throws CommandException {
+        if( args.length < 2 ){
+            throw new CommandException( "Usage: /computer <id> <value1> [value2]..." );
+        }
+        try {
+            ServerComputer computer = ComputerCraft.serverComputerRegistry.lookup(Integer.valueOf(args[0]));
+            if( computer.getFamily().equals( ComputerFamily.Command ) ){
```
SquidDev: This can just be `==` as it's an enum. Also worth noting that `.lookup` may return `null`, so you should probably have a `null` check here too.

KingofGamesYami: Done :)

SquidDev:
@KingofGamesYami This'll need rebasing onto master: it's been updated to 1.11.2 which introduces some incompatibilities.

src/main/java/dan200/computercraft/ComputerCommand.java
```diff
@@ -0,0 +1,44 @@
...
+        return "computer <id> <value1> [value2]...";
+    }
+
+    @Override
+    public void execute(MinecraftServer server, ICommandSender sender, String[] args) throws CommandException {
+        if( args.length < 2 ){
+            throw new CommandException( "Usage: /computer <id> <value1> [value2]..." );
+        }
+        try {
+            ServerComputer computer = ComputerCraft.serverComputerRegistry.lookup(Integer.valueOf(args[0]));
```
SquidDev: Having another think about this, I think it would be better to do find *all* computers with the given ID, queuing for those which are commands. If no such computers are found then it'd error. Most of the time they'll only be one computer with the given ID, but some programs (*cough* Taken *cough*) spawn multiple computers with the same ID which makes this unusable.

KingofGamesYami:
@SquidDev I finally got around to writing and testing a fix.  Sorry about the wait, I've been busy with college.

src/main/java/dan200/computercraft/ComputerCommand.java
```diff
@@ -0,0 +1,50 @@
...
+        try {
+            id = Integer.valueOf(args[0]);
+        }catch( NumberFormatException e ){
+            throw new CommandException( "Invalid ID" );
+        }
+        boolean found_valid_computer = false;
+        for( ServerComputer computer : ComputerCraft.serverComputerRegistry.getComputers() ){
+            if( computer.getID() == id && computer.getFamily() == ComputerFamily.Command ){
+                computer.queueEvent("computer_command", ArrayUtils.remove(args, 0));
+                found_valid_computer = true;
```
Lignum: There shouldn't be any computers with duplicate ids, so add a `break` here? Or, since we now have Java 8, if we make that assumption, this can be a whole lot nicer:
```Java
Optional<ServerComputer> optServerComp = ComputerCraft.serverComputerRegistry.getComputers().stream()
    .filter(c -> c.getID() == id)
    .filter(c -> c.getFamily() == ComputerFamily.Command)
    .findFirst();

if(optServerComp.isPresent())
{
    optServerComp.get().queueEvent("computer_command", ArrayUtils.remove(args, 0));
}
else
{
    throw new CommandException( "Computer #" + args[0] + " is not a Command Computer" );
}
```

KingofGamesYami: Did you read *anything* in the comments and reviews?  I previously supported 1 computer instance, and now support more - at SquidDev's request.  It is definitely possible to have more than 1 computer with the same ID.

SquidDev: In fairness, that review is now hidden as you've made changes. Sometimes GH's features are counter-productive :/. 

@Lignum For reference, this was requested because Taken spawns pocket computers with the same ID as a command computer, meaning in its original form the command failed.

Lignum: Hmm, alright, in that case a `Stream` won't be any nicer than what we have now. Nevermind then.

apemanzilla: It could still be done with a `Stream` somewhat nicely:

```
ComputerCraft.serverComputerRegistry.getComputers().stream()
    .filter(c -> c.getID == id)
    .filter(c -> c.getFamily() == ComputerFamily.Command)
    .peek(c -> c.queueEvent("computer_command", ArrayUtils.remove(args, 0))
    .findFirst().orElseThrow(() -> new CommandException("No command computer with ID " + args[0] + " found");
```

PR #363, opened 2017-07-09
Add Defaultprograms to extensions

Wilma456:
If you just type "picture.nfp" the shell will open the file in paint and not execute it..txt files will be open in edit. Other programs can add their own filetypes with the settings API.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -915,8 +915,10 @@ settings.set( "edit.default_extension", "lua" )
 settings.set( "paint.default_extension", "nfp" )
 settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
+settings.set( "filetype.txt", "edit {}" )
 if term.isColour() then
     settings.set( "bios.use_multishell", true )
+    settings.set( "filetype.nfp", "paint {}" )
 end
 if _CC_DEFAULT_SETTINGS then
     for sPair in string.gmatch( _CC_DEFAULT_SETTINGS, "[^,]+" ) do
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -915,8 +915,10 @@ settings.set( "edit.default_extension", "lua" )
 settings.set( "paint.default_extension", "nfp" )
 settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
+settings.set( "filetype.txt", "edit {}" )
```
SquidDev: Why `{}`? I'd personally go for `%s`, which means you can just do `settings.get(xyz):format(sCommand)`.

Lignum:
I'm not sure if `settings` is the best place to store this, that could lead to bloat pretty fast. I think a `shell.associate` function would work better, similar to `shell.setAlias`.

lupus590:
@Lignum and how will you get it to persist over reboots?

I propose adding a new settings file just for file associations.

apemanzilla:
@lupus590 presumably, the same way aliases and path persist over reboots - via startup scripts.

BombBloke:
"Startup scripts" such as [this one](https://github.com/dan200/ComputerCraft/blob/master/src/main/resources/assets/computercraft/lua/rom/startup.lua), presumably.

Personally I think .settings is fine, as I doubt many users will be adding extensions to it anyway, and those which do are unlikely to add more than... what, two, three?

src/main/resources/assets/computercraft/lua/bios.lua
BombBloke: I'm a fan of %1, myself!

apemanzilla: Eh, I'd rather keep it consistent with `string.format` which uses `%s`.

Cruor: @SquidDev Python uses {} for the "new" formating strings i.e. `"foo {} {key}".format("bar", baz = "baz")` either way, i agree that it should be consistent with `string.format`.

dan200:
I don't like this, it confuses the distintion between programs and files

PR #365, opened 2017-07-10
Added Spanish and Galician languages

manu-03:
Spanish files have the same text but there would be a slight difference between the slang spoken in each country. I am Spanish and I can't really know that differences, so I just copied ´es_ES.lang´ to ensure every player who has their game in Spanish can enjoy the translation (easily understandable in every Hispanic country).


Wilma456:
dann200 don't wont other lang files. In his opinion they should provided by a resource pack.

xAnavrins:
See #5

manu-03:
Alright

PR #369, opened 2017-07-12
Add /apis to package.path

Wilma456:
With this PR we had a standard folder for apis.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
@@ -30,7 +30,7 @@ local function createShellEnv( sDir )
         string = string,
         table = table,
     }
-    package.path = "?;?.lua;?/init.lua"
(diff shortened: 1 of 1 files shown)


CrazedProgrammer:
Having a hardcoded path for global APIs is not only inextensible but also a bad practice.

lupus590:
could the path be loaded from settings?

apemanzilla:
@lupus590 For consistency I'd say it should be left the way it is, the way `shell.path` works.

SquidDev:
@apemanzilla I agree, though worth noting that the include path is entirely hard coded. It would be nice to have a `shell.setRequirePath` or something.

apemanzilla:
Can't you just do `package.path = ...`?

SquidDev:
@apemanzilla Yes, but only in the main program - as it is a new instance every program run, changing it in startup will not help. That being said, I can see potential for people confusing the two methods, so maybe it is something which should be avoided.

dan200:
Nope. APIs and libraries are different things

PR #371, opened 2017-07-12
Add os.isKeyDown()

Wilma456:
I think it's more flexible than events.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -683,6 +683,14 @@ function os.reboot()
     end
 end
 
+local tKeydown = {}
+function os.isKeyDown(nKey)
+    if type( nKey ) ~= "number" then
+        error( "bad argument #1 (expected number, got " .. type( nKey ) .. ")", 2 ) 
+    end
+    return tKeydown[nKey] or false
+end
+
 -- Install the lua part of the HTTP api (if enabled)
 if http then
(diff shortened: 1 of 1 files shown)


SquidDev:
Just a couple of potential problems:

 - This will not handle multiple people holding a key at once - the first person to release it will trigger a release. Perhaps it would be better to have a counter determining the number of people holding it.
 - It is possible to close a GUI whilst holding a key, meaning the `key_up` event is never fired. This means `os.isKeyUp` will always return true.

It might also be better if this was in the `key` API - though I realise the event loop would also have to be moved there, which is a little ugly. I guess rednet already does that though so it isn't *too* bar.

apemanzilla:
IMO this shouldn't be added, if you need this behavior you can easily add it yourself with 20 lines of code -
 especially given the way that this doesn't always produce the expected behavior.

xAnavrins:
This kind of thing is very simple to implement by yourself, I don't think it warrants its place in the bios/top level coroutine.

MineRobber9000:
If this is implemented, it should be in Java code, seeing as Java code can directly access what keys are being pressed<sup>[citation needed]</sup> which would solve SquidDev's second problem. As for the first problem, well, I can't help you there. Perhaps a counter, as apemanzilla suggested.

dan200:
I'm not opposed to this feature, but it needs to be implemented properly (sum together inputs from all players reliably. which is quite a bit of work). This is not that.

PR #373, opened 2017-07-14
Add shell.setPackagePath() and shell.getPackagePath()

Wilma456:
Please look at the comments of #369. This PR add this function, so you can set package.path in the full shell and not only in the running program.

Before somebody ask, why we need shell.getPackagePath(), please look at line 15.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
@@ -12,6 +12,7 @@ local sDir = (parentShell and parentShell.dir()) or ""
 local sPath = (parentShell and parentShell.path()) or ".:/rom/programs"
 local tAliases = (parentShell and parentShell.aliases()) or {}
 local tCompletionInfo = (parentShell and parentShell.getCompletionInfo()) or {}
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -440,6 +441,17 @@ function shell.aliases()
     return tCopy
 end
 
+function shell.setPackagePath( sPath )
+    if type( sPath ) ~= "string" then
+        error( "bad argument #1 (expected string, got " .. type( sPath ) .. ")", 2 )
+    end
+    sPackagePath = sPath
+end
+
```
Wojbie: This function would not change contents of package path of currentvrunning program. It will only affect next ones.

SquidDev: I think that's the point, though I don't think it's a massive issue as long as it is documented. That being said, it might be a little clearer if it was get/setDefaultPackagePath, though that is a little verbose.

dan200:
No thanks. If you want to change the require path for your program, just modify package.path. If you want to change the require path for other programs, you shouldn't.

PR #375, opened 2017-07-14
Adds the ability to load custom bios.lua files from resource packs

### LINKED ISSUE #240: Sugestion: Ability to remove CraftOS
Before trying to implement this, I was wondering if anyone would find it useful for the ability to remove CraftOS off of any computer (like a `rm craftos` or something). What I mean by this is that every computer still comes preloaded with craftos, but you can remove it from each individual computer **without** modifying the jar file **or** using a resource pack. This could be useful when developing a computercraft os (more than just GUI for CraftOS). It possibly could be similar, but different to OpenComputers where you can install the different OSes, but with this CraftOS would be installed by default.

SquidDev:
[See here for context](https://github.com/dan200/ComputerCraft/pull/272#issuecomment-303023045)

`Computer.initLua` now delegates to `IComputerEnvironment.createResourceFile` which, by default, looks in the following locations:

 - Resouce pack files
 - The "debug" folder
 - The original ComputerCraft jar

I'm doing it this rather convoluted way to remain consistent with the `rom` mount loading, as well as ensuring the "core" doesn't add a dependency on Minecraft.

I believe this closes #240, but I'd want to discuss this too :).

Note that this is **not** indented as a replacement for #347, but something to go along side it.

### DIFF
--- a/src/main/java/dan200/computercraft/ComputerCraft.java
+++ b/src/main/java/dan200/computercraft/ComputerCraft.java
@@ -72,10 +72,10 @@ import net.minecraftforge.fml.common.network.FMLEventChannel;
 import net.minecraftforge.fml.common.network.NetworkRegistry;
 import net.minecraftforge.fml.common.network.internal.FMLProxyPacket;
 import net.minecraftforge.fml.relauncher.Side;
+import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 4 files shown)


ecktec:
Like that one, would remove any arguments against #347 .

PR #376, opened 2017-07-14
Fix argument checks made in PR #304 and #338

CrazedProgrammer:
This PR fixes some faulty argument checks made in the PRs referenced above.  
These checks didn't align with the documentation/wiki and resulted in errors where they shouldn't happen.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
@@ -332,7 +332,7 @@ function unserialize( s )
 end
 
 function serializeJSON( t, bNBTStyle )
-    if type( t ) ~= "table" or type( t ) == "string" or type( t ) == "number" or type( t ) == "boolean" then
+    if type( t ) ~= "table" and type( t ) ~= "string" and type( t ) ~= "number" and type( t ) ~= "boolean" then
         error( "bad argument #1 (expected table/string/number/boolean, got " .. type( t ) .. ")", 2 )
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -329,11 +332,11 @@ function unserialize( s )
 end
 
 function serializeJSON( t, bNBTStyle )
-    if type( t ) ~= "string" then
-        error( "bad argument #1 (expected string, got " .. type( t ) .. ")", 2 )
+    if type( t ) ~= "table" or type( t ) == "string" or type( t ) == "number" or type( t ) == "boolean" then
```
dan200: you clearly haven't tested this code? there's no way this line would work.

lupus590: Replace all of the  `==` with `~=`

Wojbie: And or with and.
`if type( t ) ~= "table" and type( t ) ~= "string" and type( t ) ~= "number" and type( t ) ~= "boolean" then`

Wojbie: Sadly CrazedProgramer is on vacation right now so its not getting changed in some time.

CrazedProgrammer:
Fixed it. I'm really sorry about this.

PR #377, opened 2017-07-15
Better Errors for fs API

Wilma456:
The Errors of the fs API now contain the Path.

### DIFF
--- a/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
+++ b/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
@@ -730,7 +730,7 @@ public class FileSystem
         }
         if( match == null )
         {
-            throw new FileSystemException( "Invalid Path (" + path + ")" );
+            throw new FileSystemException( "/" + path + ": Invalid Path" );
         }
         return match;
     }
--- a/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
+++ b/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
@@ -110,7 +110,7 @@ public class FileSystem
                 }
                 else
                 {
-                    throw new FileSystemException( "Not a directory" );
+                    throw new FileSystemException( "Not a directory (" + path + ")" );
                 }
             }
             catch( IOException e )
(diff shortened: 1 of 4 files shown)


SquidDev:
I'm not sure this current syntax reads great, maybe a couple of alternatives:

 - `'foo' is not a directory`. This doesn't work that well for other error messages like `Access denied` though.
 - `Not a directory (when listing 'foo')` - this works for everything, but is rather verbose
 - `foo: not a directory` - This is what Lua does on my machine, again it works with everything.

dan200:
Agree with SquidDev: including the path is good, but need better strings

lupus590:
Error string suggestion: `received fs path foo: not a directory`

Wilma456:
@dan200 I had updated the messages to a new style

fserror

lupus590:
The new format is similar to Vanilla Lua while still being clear that `/foo` is not a part of the file which errored.

src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
```diff
@@ -730,7 +730,7 @@ private MountWrapper getMount( String path ) throws FileSystemException
         }
         if( match == null )
         {
-            throw new FileSystemException( "Invalid Path" );
+            throw new FileSystemException( "Invalid Path (" + path + ")" );
```
Lignum: Shouldn't this be `"/" + path + ": Invalid path"` to be consistent with the others?

Lignum:
Don't repeat yourself; that should probably be a function:
```Java
static String formatFSError( String path, String message )
{
    return String.format( "/%s: %s", path, message );
}
```
It's cleaner code, and that'll let us switch the format more easily if needed.

Wojbie:
Is this really such big thing that it should require its own function?

Lignum:
@Wojbie I suppose it's a matter of taste. Coming from functional programming, I'm a fan of generalising as much as possible.

src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
Wilma456: Forget it, sorry

Wilma456:
I think, dan200 should decide, if we add a function for this.

PR #378, opened 2017-07-15
Add ability to load raw data from a string [paintutils]

MineRobber9000:
This removes the need to create a temporary file just to load an image from a string. (see [hence](https://github.com/CrazedProgrammer/chpal/blob/master/pacpal.lua#L39))

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
@@ -15,7 +15,6 @@ local function parseLine( tImageArg, sLine )
         tLine[x] = tColourLookup[ string.byte(sLine,x,x) ] or 0
     end
     table.insert( tImageArg, tLine )
-    return tImageArg
 end
 
 function parseImage( sRawData )
@@ -24,7 +23,7 @@ function parseImage( sRawData )
     end
     local tImage = {}
     for sLine in ( sRawData .. "\n" ):gmatch( "(.-)\n" ) do -- read each line like original file handling did
-        tImage = parseLine( tImage, sLine )
+        parseLine( tImage, sLine )
     end
     return tImage
 end
@@ -35,7 +34,6 @@ function loadImage( sPath )
     end
 
     if fs.exists( sPath ) then
-        local tImage = {}
         local file = io.open( sPath, "r" )
         local sContent = file:readAll()
         file:close()
--- a/src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
@@ -44,7 +44,7 @@ function loadImage( sPath )
             sLine = file:read()
         end
         file:close()
-        return parseImage( sContent:sub(2) ) -- remove first newline and delegate parsing of images to parseImage function as suggested by @SquidDev in discussion of PR #378
+        return parseImage( sContent:sub(2) ) -- remove first newline and delegate image parse to parseImage
     end
     return nil
 end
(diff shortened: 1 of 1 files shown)


SquidDev:
I feel it would be nicer if it were a separate method (like `.parseImage`), rather than doing `.loadImage("just-a-stub-string", actualContents)`. It might be possible to then get `.loadImage` just to read the file into a string and delegate to `.parseImage`.

src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
```diff
@@ -9,25 +9,49 @@ for n=1,16 do
...
-            for x=1,sLine:len() do
-                tLine[x] = tColourLookup[ string.byte(sLine,x,x) ] or 0
-            end
-            table.insert( tImage, tLine )
+            sContent = sContent .. "\n" .. sLine
             sLine = file:read()
         end
         file:close()
-        return tImage
+        return parseImage( sContent:sub(2) ) -- remove first newline and delegate parsing of images to parseImage function as suggested by @SquidDev in discussion of PR #378
```
SquidDev: You don't need that much context in the comment :p. Though seeing as you've got the `parseLine` function, you could just use that.

src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
```diff
@@ -9,25 +9,49 @@ for n=1,16 do
...
+    end
+    local function split( str, delim ) -- modified split function from https://codea.io/talk/discussion/2118/split-a-string-by-return-newline
+        if string.find( str, delim ) == nil then return { str } end
+        local result, pat, lastpos = {}, "(.-)" .. delim .. "()", nil
+        for part, pos in string.gfind( str, pat ) do table.insert( result, part ); lastpos = pos; end
+        table.insert( result, string.sub( str, lastpos ) )
+        return result
+    end
+    local tImage = {}
+    for _, sLine in pairs( split( sPath, "\n" ) ) do -- read each line like original file handling did
```
SquidDev: I wonder if it would be cleaner to iterate using `:gmatch` instead of splitting the string and then looping over it. You also don't need to reassign `tImage` on the next line.

MineRobber9000: issue is, `:gmatch` won't give you the last item (i.e; for string `"0123\n4567"`, `:gmatch` with a newline as the delimiter only gives `"0123"` and not `"4567"`.)

BombBloke: "[^\n]+"

src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
```diff
@@ -9,25 +9,49 @@ for n=1,16 do
...
     end
 
-    local tImage = {}
     if fs.exists( sPath ) then
-        local file = io.open(sPath, "r" )
+        local tImage = {}
+        local file = io.open( sPath, "r" )
         local sLine = file:read()
+        local sContent = ""
         while sLine do
```
BombBloke: Personally I'd run an io.lines() loop and call parseLines() directly.

MineRobber9000: I believe that passing the file contents to parseImage makes more sense in the context of what it does. I will, however, replace the while loop with a :readAll call.

BombBloke: In that case, why even *have* parseLine? Or have it return anything, for that matter... remember tImageArg doesn't hold a table, but a pointer leading to one.

loadImage is additionally creating a redundant tImage table that doesn't need to be there.

PR #380, opened 2017-07-18
Add textutils.formatDate()

Wilma456:
Since #183 we can get the number of days since 1970. This function give you the current date.

Example Code:
```
local nDay, nMonth, nYear = test(os.day("local"))
print("The Date is "..nDay.."."..nMonth.."."..nYear)
```
If you want to test this function with other dates visit https://www.epochconverter.com/date-difference and calculate the difference between 1.1.1970 and the date that you wish to test and add 1.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
@@ -52,6 +52,57 @@ function formatTime( nTime, bTwentyFourHour )
     end
 end
 
+local function isLeapYear( nNum )
+nNum = nNum / 4
+local a, b = math.modf( nNum )
+if b == 0 then
+    return true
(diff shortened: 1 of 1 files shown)


SquidDev:
It's worth noting that the leap year calculation is a little more complex. A leap year occurs if it is a multiple of 4 but not a multiple of 100 *unless* it is a multiple of 400.

I wonder if it would be better if there was an option for `os.day` to return a table, ala PUC Lua. That way we're not depending on ad-hoc calculations. Dates and times are very tricky things.

dan200:
The "local" and "utc" options to os.day/time are fourth wall breaking, and thus not really something that is "canonically" part of CraftOS. As this doesn't make sense for regular minecraft days (What month is "day 2"?), i'm hesitant to include this.

PR #381, opened 2017-07-18
Better Error for "Invalid color"

Wilma456:
the "Invalid color" error now contain the wrong color.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/window.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/window.lua
@@ -274,7 +274,7 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
         if type( color ) ~= "number" then 
             error( "bad argument #1 (expected number, got " .. type( color ) .. ")", 2 )
         elseif tHex[color] == nil then
-            error( "Invalid color", 2 )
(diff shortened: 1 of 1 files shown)


SquidDev:
Would it also be a good idea to include the value in the `TermAPI` error messages? I realise that is an out of bounds error instead, but might be a little more consistent.

src/main/resources/assets/computercraft/lua/rom/apis/window.lua
```diff
@@ -321,7 +321,7 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
         if type( color ) ~= "number" then
             error( "bad argument #1 (expected number, got " .. type( color ) .. ")", 2 )
         elseif tHex[color] == nil then
-            error( "Invalid color", 3 )
+            error( "Invalid color (got " .. color .. ")", 2 )
```
lupus590: the 2 was a 3, is that an intended change?

Wilma456: Yes. There must be a 2 and not a 3.

SquidDev: @lupus590 I presume the 3 was a relic from when `window.setBackgroundColor` and `window.setBackgroundColour` were separate functions which deligated to this.

PR #384, opened 2017-07-26
Move bios.lua to /rom

Wilma456:
I see no reason, why the User can't read the code of bios.lua ingame. I've personally learned a lot while reading the code of CraftOS ingame. With this PR, all files of CraftOS are in /rom.


Wojbie:
Can you confirm that this won't interfere with #375 ?

SquidDev:
It shouldn't - the path in Computer.java has been changed. That being said, it would be good to double check this all works, if you haven't already.

Wilma456:
@lupus590 @SquidDev  Don't worry. I can confirm that everything works and that bios.lua can be replaced by a resourcepackk.

lupus590:
@Wilma456 Why am I tagged in that last post? Surely you meant @Wojbie

Wilma456:
@lupus590 Sorry

SquidDev:
I personally think #347 is a better solution to this problem - this way you're not exposing all the sandbox-related code (such as Lua 5.1 compat or metatable protection) to the user, but all the interesting stuff is still accessible.

Wilma456:
@SquidDev If someone want the code code to break the sandbox, he has much experience in LUA and he know, where to find the file. But a beginner who wants to read the code behind print(),write() or read() does this not know.

SquidDev:
@Wilma456 I agree with you entirely. I'm not suggesting separating the sandbox and CraftOS code for security reasons (an idea which is, as you rightly point out, absurd). However, I do think it would be "neater" if the two are separate as they do different things. In fact, it would probably be easier for those who want to go digging into the bios if all the metatable protection code is elsewhere.

Vexatos:
I agree with @SquidDev, the sandboxing magic isn't really something that a user should see.

roger109z:
@Vexatos couldn't they easily come here to read the sandboxing code if they were really trying to hack it?

BombBloke:
@roger109z: Sure, but there's a difference between digging into the mod's GitHub, and being exposed to those details while the mod's actually running. It's one thing for a user to *ignore* the fourth wall, but quite another to go ahead and break it *for* them.

xAnavrins:
What's the point of exposing it if you can't even modify it. 
Now that servers can make their own bios files, if some operator decides to put some moderation mechanism or sandboxing, that rely on being secret, they might want to keep it secret.

Otherwise, if you want to see the original bios, it's easily acquirable.

dan200:
Bios is somewhat behind the scenes, and it can now be replaced by resource packs anyway. So this isn't necessary.

PR #386, opened 2017-07-26
added a fifth argument to read

roger109z:
Simply added a fifth argument to read to control if print is fired at the end or not. This can be very useful if you are trying to put a text field at the bottom of the screen.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -283,7 +283,7 @@ function printError( ... )
     end
 end
 
-function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault )
+function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault, _bNewLine )
     if _sReplaceChar ~= nil and type( _sReplaceChar ) ~= "string" then
         error( "bad argument #1 (expected string, got " .. type( _sReplaceChar ) .. ")", 2 ) 
     end
@@ -295,6 +295,9 @@ function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault )
     end
     if _sDefault ~= nil and type( _sDefault ) ~= "string" then
         error( "bad argument #4 (expected string, got " .. type( _sDefault ) .. ")", 2 ) 
+    end
+    if _bNewLine ~= nil and type( _bNewLine ) ~= "boolean" then
+        error( "bad argument #5 (expected boolean, got " .. type( _bNewLine ) .. ")", 2)
     end
     term.setCursorBlink( true )
 
@@ -548,7 +551,9 @@ function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault )
     local cx, cy = term.getCursorPos()
     term.setCursorBlink( false )
     term.setCursorPos( w + 1, cy )
-    print()
+    if _bNewLine == nil or true then
+        print()
+    end
     
     return sLine
 end


SquidDev:
To quote #330:

> IMO if people need text-box like behaviour then they should writer their own, rather than just relying on read (or if you must, create a new window and use read inside that). Otherwise you run the risk of having a mess of arguments. To quote Dan from #319:
>
> > Read isn't really for "text boxes", it's for command lines

Whilst I can there may be times you don't want to print a newline, its rare enough that I'm not sure its worth adding a new argument - it's already a complicated signature as is. Just copy the `read` function and specialise it for your use-case, rather than creating a "super-duper READ-O-MATIC 4000 with turbo text entry".

roger109z:
I feel like this could be useful in many situations where you just don't want the read function to newline especially in command lines maybe you want more control of the cursor

Lignum:
> especially in command lines maybe you want more control of the cursor

Not really, as has been stated, `read()` is simply a utility to create simple CLIs, which almost universally function in this fashion. If you want something more powerful, you'd be better off writing your own prompt either way. Heck, I'd be surprised if there wasn't some library buried deep in the forums that allows you to do crazy things like that.

roger109z:
I would rather have it just not print at the end but prs can't break current programs in essence I want to take control away from read and give it to the program running it

Lignum:
Sure, but this is the fifth parameter already. `read` was designed for a very specific purpose, and not printing at the end would not be part of that purpose. Either way, I've suggested the addition of a `textutils.prompt` in #387, which I think would be a good compromise.

roger109z:
I feel like read should just be a barebones function that takes user input and provides an output and if you want more functionality you would have to go to something like textutils.prompt

dan200:
read() already has too many features. Just write your own library

lupus590:
@dan200  What do you think of a pull request to 'trim the fat' off of `read`?

BombBloke:
I'd quite like additional functionality that wraps completed filenames within quotes when they contain spaces. In its current state, though, the cleanest way I can see to do that is to add (wait for it...) a fifth argument, as the shell should have this behaviour and the Lua console shouldn't.

hugeblank:
@BombBloke what do you mean by that? Doesn't fs.list() already do what you want? I don't think I'm understanding exactly what you want...

Wojbie:
What he means is that if there is a space in filenames it should get autocompleted in the quotes. Sadly current system don't allow for that kind of modification of input line by anticompetitive functions. Currently they can only add not modify. This has side effect that if you auto complete filename with spaces you have to add quotes yourself or the command is not valid/pointing at wrong file. Per example `` edit file with spaces.lua `` points to file not "file with spaces.lua". Yet current auto competition is happy to compete it like that. It would make sense to think about ways to "solve" this issue.

hugeblank:
@wojbie ah... that makes sense. Thanks for the clarification! 

And yeah that would be a doozie to work out. If I wasn't in a very loud and busy place I would offer a concept of how to fix it...

BombBloke:
It's easy to fix, it's just a question as to where the toggle should go. Either `read` gets another parameter, or autocomplete functions gain a method of telling `read` to quote certain completions.

Wojbie:
Or a method to tell read to replace whole like with one provided instead of appending to make it more powerful. I might try PR that later today.

BombBloke:
How're things going with this, Wojbie? If you got stuck, I'd be quite happy to write it myself.

Wojbie:
[Take a look at it.](https://github.com/Wojbie/ComputerCraft/tree/Advanced-Autocompletion-Additions)
I just need to add some polish and check if i did not miss any bugs and it will be ok now. I think.
Always open to suggestions/mad yelling at stupid (or not stupid) errors.

Wojbie:
@BombBloke 
[Take a look at it.](https://github.com/Wojbie/ComputerCraft/tree/Advanced-Autocompletion-Additions)
I just need to add some polish and check if i did not miss any bugs and it will be ok now. I think.
Always open to suggestions/mad yelling at stupid (or not stupid) errors.

PR #387, opened 2017-07-26
Limit start point of scrolling for read()

hugeblank:
Adds the ability to reposition the starting point along the x axis for scrolling of characters. This would be particularly useful for command line GUIs that don't want the entire x axis being put towards the input area.

To clarify: I _have_ seen PR [303](https://github.com/dan200/ComputerCraft/pull/330). This works nothing like that; it allows for scrolling, not blocking characters being typed. And I also know that this could be done using the window API, but the way I see it, it is a lot more efficient having the scrolling built into the read function for the user. Plus the addition of this argument is optional, it has no bearing on past programs, much like a majority of the read function already.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -353,7 +353,7 @@ function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault, _nLimit )
         term.setCursorPos( sx, cy )
         local sReplace = (_bClear and " ") or _sReplaceChar
         if sReplace then
-            term.write( string.rep( sReplace, math.max( string.len(sLine) - nScroll, 0 ) ) )
+            term.write( string.sub( string.rep( sReplace, math.max( string.len(sLine) + nScroll, 0 ) ),  nScroll + 1, nScroll + w ) )
         else
             term.write( string.sub( sLine, nScroll + 1, nScroll + w ) )
         end
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -283,7 +283,7 @@ function printError( ... )
     end
 end
 
-function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault )
+function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault, _nLimit )
     if _sReplaceChar ~= nil and type( _sReplaceChar ) ~= "string" then
         error( "bad argument #1 (expected string, got " .. type( _sReplaceChar ) .. ")", 2 ) 
     end
@@ -295,6 +295,9 @@ function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault )
     end
     if _sDefault ~= nil and type( _sDefault ) ~= "string" then
         error( "bad argument #4 (expected string, got " .. type( _sDefault ) .. ")", 2 ) 
+    end
+    if _nLimit ~= nil and type( _nLimit ) ~= "number" then
+        error( "bad argument #5 (expected number, got " .. type( _nLimit ) .. ")", 2 ) 
     end
     term.setCursorBlink( true )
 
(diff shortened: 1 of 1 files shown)


SquidDev:
To quote #330:

> IMO if people need text-box like behaviour then they should writer their own, rather than just relying on read (or if you must, create a new window and use read inside that). Otherwise you run the risk of having a mess of arguments. To quote Dan from #319:
>
> > Read isn't really for "text boxes", it's for command lines

Whilst there is a use case for it, its rare enough that I'm not sure its worth adding a new argument - it's already a complicated signature as is. 

Remember, `read` isn't designed for GUIs - it's just a simple tool for the command prompt. Just copy the `read` function and specialise it for your use-case, rather than creating a "super-duper READ-O-MATIC 4000 with turbo text entry".

hugeblank:
@SquidDev Understandable, but I feel it has some form of use for people that want to work with the command line, especially with large monitors.

Lignum:
I agree with @SquidDev, however, judging by the amount of PRs relating to making `read()` more powerful, it appears that there really does appear to be a great need for this sort of thing. If anything, this would be a job for `textutils`. Perhaps a `textutils.prompt` function which takes a table to allow for a wide range of customisations would satisfy these needs.

BombBloke:
> there really does appear to be a great ~~need~~ **desire** for this sort of thing.

Anyone submitting a PR that adds on to `read` has *in the process* demonstrated that they don't "need" that functionality to be coded directly into the mod at all.

Lignum:
It's not strictly necessary, of course, which is why I suggested it be added to `textutils`, which, as its name suggests, is a utility library full of functions like `pagedPrint` and `tabulate`. If any of the features from the read PRs (and more) were added, I think it would best find its home among these functions as a `textutils.prompt`. Of course, there's still the question whether it's worth the additional clutter, and I'm not entirely sure about that myself, which is why I'm putting this up for debate. Ultimately, we're evaluating if we want to stick to ComputerCraft's philosophy of keeping everything small and simple, or if we're following the desire of the community, as you mentioned. Since we've generally been drifting off from the former a bit, I tend towards thinking that adding a function like that couldn't hurt, though I am still torn myself.

hugeblank:
I'm actually with @Lignum here, this probably would be better in textutils. I'll see what I can do about setting it up, that is unless there is a PR already for said `textutils.prompt`

apemanzilla:
@hugeblank @CrazedProgrammer has been working on some improvements to the `texutils` API already, so we should wait and see if he's up for implementing it.

hugeblank:
@apemanzilla alright well when/if they get merged, I'll add it :P

dan200:
read() already has too many parameters.

lupus590:
Maybe "read() already has too many parameters." should go in the readme?

hugeblank:
I second this ^ 

But as I said the concept of a scrolling start point should be implemented in some sort of input function. 

Has said input function already been implemented?

lupus590:
@hugeblank Regarding the input function, perhaps it should go in `textutils`. Or maybe a new API should be made called `IOUtils`?

hugeblank:
@lupus590 It could go in `ioutils`, but what else would? It would be pointless having an API with only 1 function.

lupus590:
Or maybe a `GUIUtils`? given how many people ask for help with button API's this may be useful.

It might be worth making an issue/forum thread asking for function suggestions for the new APIs.

apemanzilla:
I'd just put it in `textutils` since it already contains some other interactive functions like `pagedPrint`

hugeblank:
I think i'll put it in uiutils (less letters the better imo). That way a button API could be added, along with an advanced read function, and the like

Wojbie:
That realy feels like stuff that user should make or find on forums. Not something to provide on default.

hugeblank:
Yes, but I would argue that that doesn't mean it shouldn't exist in API form. 
For example, `paintutils`: `drawLine`, `drawBox`, and `drawFilledBox` are all things that users should make or find on the forums, but they exist by default.

Wojbie:
Those are rather simple things. Button API or advanced read are more complicated.

hugeblank:
A button API is rather simple...

```
local buttons = {}
button = {
	add = function(x1, y1, x2, y2, name)
		buttons[#buttons+1] = {x1, y1, x2, y2, name}
	end,
	check = function()
		_, _, cx, cy = os.pullEvent("mouse_click")
		for i = 1, #buttons do
			if buttons[i][1] <= cx and buttons[i][2] <= cy and buttons[i][3] >= cx and buttons[i][4] >= cy then
				return buttons[i][5]
			end
		end
	end
}
```

This would work perfectly fine (of course with better variable names).

PR #389, opened 2017-07-28
Make io.open() return Error

Wilma456:
If fs.open() failed, it will return nil and a a error message like "No such file"  or "Out of space". This PR make, that io.open() returned this error too.


SquidDev:
It'd probably be worth adding the same check to `loadfile`, rather than the current "file not found", especially when #377 is merged

PR #390, opened 2017-07-28
Show fs error in paint and edit

JakobDev:
If saving failed, you now get a error, that shows why it failed.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -292,7 +292,7 @@ local tMenuFuncs = {
                 sStatus="Saved to "..sPath
             else
                 if fileerr then
-                    sStatus="Error saving to "..sPath.." ("..fileerr..")"
+                    sStatus="Error saving to "..fileerr
                 else
                     sStatus="Error saving to "..sPath
                 end
@@ -774,3 +774,4 @@ end
 term.clear()
 term.setCursorBlink( false )
 term.setCursorPos( 1, 1 )
+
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -93,7 +93,7 @@ local function save( _sPath )
     -- Save
     local file = nil
     local function innerSave()
-        file = fs.open( _sPath, "w" )
+        file, fileerr = fs.open( _sPath, "w" )
```
Lignum: `fileerr` should not be global.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -107,7 +107,7 @@ local function save( _sPath )
     if file then 
         file.close()
     end
-    return ok, err
+    return ok, err, fileerr
```
SquidDev: Perhaps instead of storing `fileerr` in an upvalue and returning, you should have `error(fileerr, 0)`. That way you don't need the separate `err` and `fileerr` values. It would be worth checking how it handles out of space errors though - those will be generated through `file.write` instead.

SquidDev:
It's worth considering how this will interact with #377 - I'm presuming you'll get something like `Error saving to bild.nfp: (/bild.nfp: Out of space)`, which could be a little weird. Obviously we should wait for that to be merged before worrying about it though :).

lupus590:
@SquidDev Thinking ahead for that case, should the programs modify the error message to remove the file name from the error (the part in brackets)?

SquidDev:
@Wilma456 What does this look like now that #377 has been merged? Would it be better to no longer show the file name, as it'll be included in the error message?

Wilma456:
Here's the new Style

PR #391, opened 2017-07-28
Make fs.getSize() work for Directories

Wilma456:


lupus590:
My gut tells me that there was a reason that this was not originally in the mod.

KnightMiner:
How consistent is this with checking each file in the directory? Basically, does it take into account the minimum file I saw mentioned before?

SquidDev:
@KnightMiner `fs.getSize` on files doesn't take into account the minimum size either. One wonders what the correct approach would be here - ideally `fs.getFreeSpace("/") + fs.getSize("/") - fs.getSize("rom")` would be the capacity of the computer, but again that could be confusing.

Lignum:
That's not all though, you'd still have to change [this](https://github.com/dan200/ComputerCraft/blob/master/src/main/java/dan200/computercraft/core/filesystem/JarMount.java#L36) line in `JarMount`, and I'm sure there's other mount types in need of changing as well.

apemanzilla:
I'd rather keep it the way it is. It's trivial to implement in Lua with some recursion if you need to get the total size of a directory's contents.

Wojbie:
And it makes a interesting thing for someone to do.

BombBloke:
I suspect that changing the output for directories in this manner may break existing space-calculating scripts. Given that it's already *possible* to write such scripts, why would we even need such a change?

Certainly I wouldn't want `fs.getSize()` to return a minimum value of 500: that'd simply make it harder to figure out how much more data you can actually store on a drive. Currently you can round up easily whenever you feel you need to, but working in the opposite direction would be a right nuisance. Though I can see that this PR in its current state isn't touching *that* behaviour.

> ideally `fs.getFreeSpace("/") + fs.getSize("/") - fs.getSize("rom")` would be the capacity of the computer, but again that could be confusing.

Better to ignore *any* directory that represents a drive. rom, disk, disk_2, drives added by mods or user scripts... basically any item on the root for which fs.getDrive() doesn't return "hdd" should be discounted.

Wilma456:
I think the solution for the / problem is, that fs.getSize("/") keep returning 0.

dan200:
Breaking change for little benefit.

PR #392, opened 2017-07-30
Fix Bug with term.getPaletteColour()

Wilma456:
The Problem is, that the Colour Palette is global and not bound to a window. You can trigger the Bug with this code:
```
local win1 = window.create(term.current(),1,1,20,10)
local win2 = window.create(term.current(),20,1,20,10)
win1.setTextColor(colors.red)
win1.write("Hello World")
win2.setTextColor(colors.red)
win2.write("Hello World")
win1.setPaletteColor(colors.red,0.4,0.4,0.4)
win2.setPaletteColor(colors.red,0.5,0.5,0.5)
win1.setCursorPos(1,2)
win1.write(win1.getPaletteColor(colors.red))
win2.setCursorPos(1,2)
win2.write(win2.getPaletteColor(colors.red))
```
If you run this code, you will see, that the 2 Window have the same Palette but getPaletteColor return different values.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/window.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/window.lua
@@ -311,8 +311,12 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     window.setPaletteColor = window.setPaletteColour
 
     function window.getPaletteColour( colour )
-        local tCol = tPalette[ colour ]
-        return tCol[1], tCol[2], tCol[3]
+        if type( colour ) ~= "number" then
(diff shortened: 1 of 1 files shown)


Wojbie:
Well i do believe that children palette is supposed to be separate from parent.
And then every time window is redrawn it updates its parent pallete to his own pallete using updatePalette().
What you are proposing would cause window to ignore its internal palette and use one of its parent?

Lignum:
> Well i do believe that children palette is supposed to be separate from parent.
And then every time window is redrawn it updates its parent pallete to his own pallete using updatePalette().

Yeah, that's exactly the intention. There's not really a sane way to manage palettes when a parent has multiple windows with different palettes, so windows just enforce their own palette once they get drawn. This isn't a bug.

Issue #35, opened 2015-12-11
[1.76pr2] cd ... crashes shell


BombBloke:
... or cd with any other number of additional periods.

Under Windows 8.1 with an NTFS file system, when the shell is working within a folder other than the root of the drive, shell.resolve("...") returns "current directory + ..." (as opposed to the expected "current directory alone"). It should act in the same manner as shell.resolve(".").

Because fs.exists( shell.resolve( "..." ) ) always (correctly) resolves as true, typing "cd ..." into the command line leads to the shell switching to a non-existant directory (as shell.resolve() brings it to the wrong location). Auto-complete promptly crashes shell when the next character is typed ("Java Exception Thrown: java.lang.NullPointerException").

![cddotdotdot](https://cloud.githubusercontent.com/assets/12162603/11746154/a3ed8e9c-a070-11e5-8ccf-710bcb3a420a.gif)

dan200:
Ooh, dangerous! I didn't know about the Windows "..." shortcut. I've fixed it by treating "..." as ".." on all platforms.

dan200:
(Please let me if this works for you when pr3 is released, as I don't have a Windows system to test on)

BombBloke:
"..." (or "....", or ".....", or "......", etc...) should act as ".", not "..". As far as I'm aware, only ".." should act as ".." does.

oddstr13:
In CMD
![cmd](https://dl.dropboxusercontent.com/u/4306267/ShareX/2015/12/2015-12-12_07-31-15.png)
---
In Cygwin with bash (still NTFS filesystem)
![cygwin bash](https://dl.dropboxusercontent.com/u/4306267/ShareX/2015/12/2015-12-12_07-33-55.png)
---
In explorer after running `touch test` in cygwin
![explorer](https://dl.dropboxusercontent.com/u/4306267/ShareX/2015/12/2015-12-12_07-37-28.png)
Notice how I had to open the directory twice to actually get there, but that it does in fact work.

I have not been able to navigate to that folder in cmd using `cd ...` or `cd .../...`, but `type ...\test` outputs the content of the file.

Seems to me that this is a untested edge case in Windows, and that there are some bugs here and there.

You should either treat `"..."` as an actual directory, and then make sure to handle the edge cases properly,
or you could strip the occurrence of `r"\.\.\.+"` all together (or replace with `"."`, which is how the `cd` in cmd.exe seem to be working)

[...]

PR #394, opened 2017-07-30
Modify logic behind ... and more in file path.

Wojbie:
Changes all equal or longer then 3 multidots setups to be treated as .
This removes other potentially dangerous situations and brings it closer to windows in how it treats said dots. This makes sure that even on systems that have wierd magic meanings to ... and onward CC will work with predictable path no matter the system.

This fixed the bugs expressed in #35 , http://www.computercraft.info/forums2/index.php?/topic/16882-computercraft-beta-versions-bug-reports/page__st__40__p__213941#entry213941 and http://www.computercraft.info/forums2/index.php?/topic/28870-command-crashes-computer/

### DIFF
--- a/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
+++ b/src/main/java/dan200/computercraft/core/filesystem/FileSystem.java
@@ -765,14 +765,15 @@ public class FileSystem
         Stack<String> outputParts = new Stack<String>();
         for( String part : parts )
         {
-            if( part.length() == 0 || part.equals( "." ) )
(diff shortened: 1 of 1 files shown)


dan200:
This is a good change, but I don't like the performance hit of compiling a regular expression several times on every path operation. Precompile the pattern!

Wojbie:
Hopefully compiling it once for each path operation is ok? I completely did not think about that for loop there for some reason.
Anyhow requested changes implemented.

dan200:
Maybe make the pattern a static class member?

Wojbie:
I was not sure what was a standard for placement of  static class member. Hope this is correct way?
Anyhow requested change implemented.

PR #395, opened 2017-07-30
Websocket support

### LINKED ISSUE #161: Suggestion - Web Sockets
Seeing as we're getting new features added now, I'd like to suggest websockets. They've been suggested on the forums in the past, and would be a much better way to maintain communication with a server compared to the current method of long-polling HTTP requests, as seen in [jamROGUE](https://github.com/Lemmmy/jamROGUE). They're also significantly easier to restrict than raw TCP sockets, which would be good for server admins, and can also support SSL for security.

There's a [forum thread](http://www.computercraft.info/forums2/index.php?/topic/28086-websocket-implementation/) suggesting them, and they've also been implemented in CCTweaks by @SquidDev already which makes it really easy to port over to the core mod.

SquidDev:
This uses Netty's websocket functionality, meaning we do not have to depend on another library.

## API
As websockets do not fit neatly into the standard polling socket model, the API is significantly more event based than CCTweaks's. One uses `http.websocket` to connect, which will wait until a connection is established and then returns the connection object (an async variant is available).

Once you have a websocket object, you can use `.send(msg)` to transmit a message. Incoming messages will fire a "websocket_message" event, with the URL and content as arguments. A convenience method (`.receive()`) exists to aid waiting for valid messages.

A `websocket_closed` event is also fired if the server forcibly terminates the connection, or it times out.

I'd really appreciate feedback on this - I'm not entirely sure it's the best approach, but IMO it's neater than CCTweaks's.

## Example
```lua
local res, err = http.websocket("wss://echo.websocket.org")
if not res then error(err, 0) end

for i = 1, 10 do
    res.send("hello " .. i)
    print(res.receive())
end

res.close()
```

## Possible enhancements
These are things which probably should be implemented at some point, but I'm looking for some thoughts on them first. Not all of them are entirely websocket related.

 - Limit the number of open sockets. I think it is worth discussing how we can best limit both the number of open sockets _and_ the number of active HTTP requests. It should be relatively trivial to handle either.
 - Automatically close the socket when it is GCd, like we do for file handles. We'd probably have to split up the connection into the `ILuaObject` part and the socket holding part, detecting when the former is no longer referenced.
 - Have a separate blacklist (and whitelist?) for websockets. There may be some sites for which one would like to allow simple requests, but not websockets.

---

Closes #161


lupus590:
The API description which you gave at the start of your post sounds like it mimics the rednet/modem API's in interface, which I would consider a plus as that is what some users of the mod will be more familiar with when using the API.

MineRobber9000:
I made what I believe to be the first public software to use this PR. [It makes a file handle interface to the WebsocketConnection object.](https://pastebin.com/PnE7Kq9n)

PR #396, opened 2017-07-30
Simple code hints in edit

Lignum:
With this PR, edit will show syntax errors and warnings when saving.

As of now, the only warning that exists is `os.loadAPI` being deprecated, though more can be easily added by adding a pattern to the `tWarnings` table (it's a bit silly, but probably better than doing bytecode analysis in the long run).

I'd imagine this would be useful for ComputerCraft's mainly beginner-level audience, who often struggle with syntax errors more than actual runtime errors. Additionally, we can use this to discourage certain coding practices, such as the use of `os.loadAPI`, though I was also thinking about `sleep(0)` calls and similar.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -28,6 +28,13 @@ local scrollX, scrollY = 0,0
 local tLines = {}
 local bRunning = true
 
+local tMessages = {}
+local tMessageColours = {
+    error = colours.red,
+    warning = colours.yellow,
+    info = colours.blue
+}
+
 -- Colours
 local highlightColour, keywordColour, commentColour, textColour, bgColour, stringColour
 if term.isColour() then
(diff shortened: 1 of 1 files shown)


lupus590:
What is this like on non-advanced grey computers?

Lignum:
Good question, the colours would probably work due to recent changes, but that doesn't matter since you can't click the hints... Therefore, I think it would make sense to just disable the feature entirely on normal computers. Since advanced computers generally grant you editing luxuries like syntax highlighting, that's kind of logical anyway, isn't it?

MaoamZedong:
Why not pop up all warnings/errors when pressing Ctrl?

Wilma456:
Edit is no IDE. I don't want warnings, when I edit a non LUA file. But a Opt-In with a setting would be OK.

Lignum:
@Wilma456 You presumably wouldn't want syntax highlighting either. So I suppose it would make sense to disable syntax highlighting and code hints if the file that's being edited doesn't have a .lua extension. But that's for another PR.

xAnavrins:
@Wilma456 Says the person who made edit save files with a .lua extension by default.

This is really neat, and it may also reduce the amount of new posts in Ask a pro :stuck_out_tongue:

SquidDev:
@xAnavrins I'm pretty sure that was Dan rather than Wilma. Wilma added a setting to turn it off.

viluon:
@SquidDev not to mention a large portion of CC users is stuck on older versions present in modpacks. It will take months, if not years, before such a change would show a large scale effect.

Cool idea @Lignum! Although I don't really like the looks of it... Perhaps placing the warning/error below the affected line would look nicer?

What I would definitely want to see is improved handling of whitespace, and perhaps even warnings on trailing whitespace, very long lines, broken indentation and the like. Making whitespace visible would help too, although that is, once again, a matter of another PR.

PR #397, opened 2017-07-30
Type checking utility function

Lignum:
This PR adds a `check` utility function to `bios.lua`, used as such:
```Lua
local function add(x, y)
    check(x, "number", y, "number")
    return x + y
end
```

It's also possible to specify multiple possible types for certain arguments:
```Lua
check(x, "number,string,boolean", y, "string,table", z, "table,nil")
```
By including `nil` as a possible type, the argument becomes optional.
Intuitively, you can use `"any"` to mean any type:
```Lua
check(x, "string", y, "any", z, "boolean")
``` 
This is useful for maintaining the parameter numbering in the error output.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -666,9 +666,7 @@ function os.unloadAPI( _sName )
     end
 end
 
-function os.sleep( nTime )
-    return sleep( nTime )
-end
+os.sleep = sleep
 
 local nativeShutdown = os.shutdown
 function os.shutdown()
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -283,8 +283,10 @@ function printError( ... )
     end
 end
 
+local fnTablePack = table.pack
+
 function check( ... )
-    local tArgs = table.pack( ... )
+    local tArgs = fnTablePack( ... )
     local nArgs = tArgs.n
     if nArgs % 2 ~= 0 then
         return error( "amount of check args is not divisible by 2" )
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -291,7 +291,6 @@ function check( ... )
     end
 
     for i=1,nArgs,2 do
-        local nArg = i / 2 + 1
         local sParam = tArgs[ i ]
         local sActualType = type( sParam )
         local sExpectedTypes = tArgs[ i + 1 ]
@@ -306,7 +305,7 @@ function check( ... )
         if tExpectedTypes and not tExpectedTypes[ sActualType ] then
             local sErr = string.format(
                 "bad argument #%d (expected %s; got %s)",
-                nArg, sExpectedTypes, sActualType
+                i / 2 + 1, sExpectedTypes, sActualType
             )
             return error( sErr, 2 )
         end
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -284,6 +284,10 @@ function printError( ... )
 end
 
 local fnTablePack = table.pack
+local fnType = type
+local fnStringLower = string.lower
+local fnStringGmatch = string.gmatch
+local fnStringFormat = string.format
 
 function check( ... )
     local tArgs = fnTablePack( ... )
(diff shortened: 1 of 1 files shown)


Wojbie:
Hmm. This does make the functions cleaner.

Lignum:
> But does any includes nil or not?

In my implementation it just omits the whole type checking altogether, so yes, `nil` is included.

Lignum:
Also, quick poll: Would you prefer `{ "string", "table", "number" }` over `"string,table,number"`?
:+1: = yes, :-1: = no

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -641,30 +646,28 @@ function os.loadAPI( _sPath )
...
-    end
+    check( _sName, "string" )
     if _sName ~= "_G" and type(_G[_sName]) == "table" then
         _G[_sName] = nil
     end
 end
 
 function os.sleep( nTime )
-    sleep( nTime )
+    return sleep( nTime )
```
BombBloke: os.sleep = sleep?

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -283,19 +283,38 @@ function printError( ... )
...
-    if _tHistory ~= nil and type( _tHistory ) ~= "table" then
-        error( "bad argument #2 (expected table, got " .. type( _tHistory ) .. ")", 2 ) 
-    end
-    if _fnComplete ~= nil and type( _fnComplete ) ~= "function" then
-        error( "bad argument #3 (expected function, got " .. type( _fnComplete ) .. ")", 2 ) 
-    end
-    if _sDefault ~= nil and type( _sDefault ) ~= "string" then
-        error( "bad argument #4 (expected string, got " .. type( _sDefault ) .. ")", 2 ) 
+function check( ... )
+    local tArgs = table.pack( ... )
```
BombBloke: `check` will be called a LOT, probably better to avoid redundant function calls like table.pack - suggest using either {...} or arg instead

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -283,19 +283,38 @@ function printError( ... )
...
+    local nArgs = tArgs.n
+    if nArgs % 2 ~= 0 then
+        return error( "amount of check args is not divisible by 2" )
+    end
+
+    for i=1,nArgs,2 do
+        local nArg = i / 2 + 1
```
BombBloke: This usually isn't used, could stand to be moved down

Wilma456:
Why is the error level hardcoded to 2?

SquidDev:
Whilst this is shorter, I do worry about the performance. I wonder if a better compromise would be a simple `checkArg(idx, val, type[, level])`, where type is a single type. This way the code is still cleaner for most cases, and you can write the more complex checks by hand. I don't see this as a bad thing as some functions are always going to require the manual check anyway, like `textutils.tabulate`

src/main/resources/assets/computercraft/lua/bios.lua
SquidDev: It's worth noting that `arg` is considered depricated in Lua 5.1. If `table.pack` is cached in an upvalue, then the performance should be similar. It might be worth investigating arg vs select vs table.pack.

src/main/resources/assets/computercraft/lua/bios.lua
Lignum: Yes

src/main/resources/assets/computercraft/lua/bios.lua
Lignum: Sure, will do

src/main/resources/assets/computercraft/lua/bios.lua
Lignum: Oh, you meant that as code. That makes sense, I'll change it

BombBloke:
@Wilma456: Tail calls knock the previous function out of the stack. It's as if `check` called `error` directly, so using level 2 lays the "blame" on the function that called `check`.

src/main/resources/assets/computercraft/lua/bios.lua
BombBloke: Deprecated it may be, but it's the fastest option I know of. `{...}` followed by `#` to get the length isn't as fast as `arg` either, but I find either language feature to be faster than making a function call.

apemanzilla:
Has anyone determined that this actually causes a significant performance impact? I doubt that either this or Squid's method would be significant enough to matter except in extreme high-performance areas (e.g. hashing implementations) in which you can just avoid using it.

PR #398, opened 2017-07-30
Small cleanup in ComboMount

Lignum:
I changed the constructor to take varargs instead of an array and changed all the loops to iterate forwards, so that the mount parts are prioritised in the order they are passed to the constructor, not in the reverse order.

### DIFF
--- a/src/main/java/dan200/computercraft/ComputerCraft.java
+++ b/src/main/java/dan200/computercraft/ComputerCraft.java
@@ -79,10 +79,7 @@ import java.io.*;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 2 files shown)


SquidDev:
If you're doing this, you should reverse the order mounts are passed in when constructing this - otherwise you won't be able to override rom files with resource packs.

Lignum:
Hm, I hate to give up the intuition here, so I came up with this compromise, which I hope is an agreeable solution.

SquidDev:
An alternative solution woild be to reverse the lookup order in getResourceMount - so it lists resource packs, debug and then the mod jar. This way resource packs are loaded in alphabetical order, which might be more intuitive.

dan200:
Iterating in reverse order is very much deliberate: the whole point is that later mounts can override files in earlier mounts.

Lignum:
Oh, well, in that case, I'm closing this PR. I don't deem the varargs constructor important enough to be in need of merging.

Issue #219, opened 2017-05-08
Shell resolution on sub-directories


SquidDev:
When you have a program in a sub-directory, the shell auto complete does not include it unless you give the absolute path. This can be reproduced as follows:

 - `mkdir test`
 - `edit test/program` and save
 - Type `test` in the shell. Note that it does not auto complete to `test/program`.

The reason for this is that `shell.programs` does not include sub-directories. For most directories on the path, this makes sense. However, for the current directory this is rather annoying.

It's also worth nothing that running `turtle/excavate 10` on a normal computer will actually work, as so far as the program will start. This doesn't seem like desired behaviour, though I guess it *could* be useful.

The first point can be "fixed" by adding `fs.complete` to this branch of `completeProgram`. This does feel rather ugly as you're special casing a specific element of the path. Instead, I propose the following rules to `shell.resolveProgram` and `completeProgram`:

 - If a program contains `/` then resolve and complete relative to the file system root. This includes paths starting with `/` as well as `./`, `../` and `test/`.
 - Otherwise, resolve using the path. Auto-completion should attempt to use both the path and the current directory. So `tes` would still complete to `test/program`.

There is an argument for removing the current directory from the path and requiring `./`. [This bug](http://www.computercraft.info/forums2/index.php?/topic/28587-) - which is triggered by moving to the `/rom/help` directory and being unable to get out as `cd` resolves to `/rom/help/cd` - does surface relatively regularly. I don't think it should be changed, but it is something worth bearing in mind.

SquidDev:
@Wilma456 I noticed you've reacted to this issue, just wondering why? I mainly created this issue to get other people's thoughts on the matter, so I'd really like to know other people's opinions too :).

Wilma456:
@SquidDev Sorry, I misunderstand you. I had thought that you mean, that all programs in Subdirectories are added that shellpath. I had read that Issue late in the evening.

MineRobber9000:
I'd like to suggest that instead of removing "." from the path, it should be later in the list (allowing programs in the current dir called by name but also prevents the /rom/help command override situation)

dan200:
Help files have a .txt extension now, so that problem is already solved!

PR #402, opened 2017-08-02
Tweak shell program resolution slightly

SquidDev:
See #219 for a more detailed explanation and rational

 - Path containing `/` or `\` are resolved relative to the current directory, rather than using the path. Paths starting with `/` still resolve relative to the root directory.
 - Shell completion will also include sub-directories of the current directory.


Wojbie:
I don't think it currently works correctly.
Try this sequence of events:
``cd rom
write/autocomplete to programs/time.lua``
Expected result : runs time.lua
Got result: No such program

Worst thing is that it auto completed correctly.

SquidDev:
@Wojbie Apparently I failed to test execution from places which weren't the root directory, should be fixed now. Thanks for spotting it!

PR #404, opened 2017-08-04
Fix Bug in copy.lua, mkdir.lua and rename.lua

Wilma456:
If the Destination exists, copy will crash with the Error "copy.lua:14: File exists". This PR fix this.

Edit:
I have noticed, that mkdir and rename have the same problem. This is also fixed in this PR.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/copy.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/copy.lua
@@ -11,7 +11,7 @@ local tFiles = fs.find( sSource )
 if #tFiles > 0 then
     for n,sFile in ipairs( tFiles ) do
         if fs.exists( sDest ) == true then
-            printError( "File exists" )
+            printError( "Destination exists" )
         elseif fs.isDir( sDest ) then
             fs.copy( sFile, fs.combine( sDest, fs.getName(sFile) ) )
         elseif #tFiles == 1 then
--- a/src/main/resources/assets/computercraft/lua/rom/programs/copy.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/copy.lua
@@ -10,7 +10,9 @@ local sDest = shell.resolve( tArgs[2] )
 local tFiles = fs.find( sSource )
 if #tFiles > 0 then
     for n,sFile in ipairs( tFiles ) do
-        if fs.isDir( sDest ) then
+        if fs.exists( sDest ) == true then
+            printError( "File exists" )
+        elseif fs.isDir( sDest ) then
             fs.copy( sFile, fs.combine( sDest, fs.getName(sFile) ) )
         elseif #tFiles == 1 then
             fs.copy( sFile, sDest )
--- a/src/main/resources/assets/computercraft/lua/rom/programs/mkdir.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/mkdir.lua
@@ -7,7 +7,7 @@ end
 local sNewDir = shell.resolve( tArgs[1] )
 
 if fs.exists( sNewDir ) == true then
-    printError( "File exists" )
+    printError( "Destination exists" )
     return
 end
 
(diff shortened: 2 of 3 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/copy.lua
```diff
@@ -10,7 +10,9 @@ local sDest = shell.resolve( tArgs[2] )
 local tFiles = fs.find( sSource )
 if #tFiles > 0 then
     for n,sFile in ipairs( tFiles ) do
-        if fs.isDir( sDest ) then
+        if fs.exists( sDest ) == true then
+            printError( "File exists" )
```
lupus590: This error message can be confusing if the destination is a folder, perhaps change it to `"Destination exists"`.

src/main/resources/assets/computercraft/lua/rom/programs/copy.lua
```diff
@@ -10,7 +10,9 @@ local sDest = shell.resolve( tArgs[2] )
 local tFiles = fs.find( sSource )
 if #tFiles > 0 then
     for n,sFile in ipairs( tFiles ) do
-        if fs.isDir( sDest ) then
+        if fs.exists( sDest ) == true then
```
SquidDev: You don't need `== true`. Just `if fs.exists( sDest ) then` is enough.

Wilma456: I know, but I think == true is a better style.

SquidDev: Hrmrrmrm, why? I'm genuinely curious - it seems rather verbose to me. Each to their own I guess :).

src/main/resources/assets/computercraft/lua/rom/programs/rename.lua
```diff
@@ -6,4 +6,9 @@ end
 
 local sSource = shell.resolve( tArgs[1] )
 local sDest = shell.resolve( tArgs[2] )
+
+if fs.exists( sDest ) == true then
+    printError( "File exists" )
```
lupus590: `"Destination exists"` here too.

src/main/resources/assets/computercraft/lua/rom/programs/mkdir.lua
```diff
@@ -5,5 +5,11 @@ if #tArgs < 1 then
 end
 
 local sNewDir = shell.resolve( tArgs[1] )
+
+if fs.exists( sNewDir ) == true then
+    printError( "File exists" )
```
lupus590: `"Destination exists"` here too.

Wilma456:
The Strings are now Changed.

btw:
Does anybody know, why the second commit doesn't have my Profile Picture and Linked not to my Profile?

SquidDev:
> Does anybody know, why the second commit doesn't have my Profile Picture and Linked not to my Profile?

Running `git show --format=full 5c7ab57782d61412f5910fc97a73fd4423346078` produces the following output:

```
commit 5c7ab57782d61412f5910fc97a73fd4423346078
Author: Wilma456 (Jakob0815) <wilma456@gmx.de>
Commit: GitHub <noreply@github.com>

    Change "File exists" to "Destination exists"
```

As you can see, the committer is "GitHub <noreply@github.com>". I'm not sure why this occurred, but I presume that's responsible for it not being associated with your account.

Wilma456:
@SquidDev Thank You. I will make a updated Version of this PR to prevent issues with credits generation.

apemanzilla:
Credit generation shouldn't be affected. You're already correctly marked as the committer in other commits so you'll be in the list already, and it ignores commits from `GitHub`.

src/main/resources/assets/computercraft/lua/rom/programs/copy.lua
viluon: The two styles produce a different result in bytecode[1], but have basically the same effect.

[1]:
```
viluon@deluon:~$ luac5.1 -l -o /dev/null -
local x

if x then end

if x == true then end

main <stdin:0,0> (5 instructions, 20 bytes at 0xc98230)
0+ params, 2 slots, 0 upvalues, 1 local, 1 constant, 0 functions
	1	[3]	TEST     	0 0 0
	2	[3]	JMP      	0	; to 3
	3	[5]	EQ       	0 0 -1	; - true
	4	[5]	JMP      	0	; to 5
	5	[5]	RETURN   	0 1
```

PR #408, opened 2017-08-08
Make Multishell Scrollable

JakobDev:
You can now Scroll through the Programlist, if the list bigger than your Screen.

multi

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
@@ -8,7 +8,7 @@ local nCurrentProcess = nil
 local nRunningProcess = nil
 local bShowMenu = false
 local bWindowsResized = false
-local nScrollpos = 1
+local nScrollPos = 1
 local bScrollRight = false
 
 local function selectProcess( n )
@@ -117,15 +117,15 @@ local function redrawMenu()
         parentTerm.setCursorPos( 1, 1 )
         parentTerm.setBackgroundColor( menuOtherBgColor )
         parentTerm.clearLine()
-        local nCharcou = 0
+        local nCharCount = 0
         local nSize = parentTerm.getSize()
-        if nScrollpos ~= 1 then
+        if nScrollPos ~= 1 then
             parentTerm.setTextColor( menuOtherTextColor )
             parentTerm.setBackgroundColor( menuOtherBgColor )
             parentTerm.write( "<" )
-            nCharcou = 1
+            nCharCount = 1
         end
-        for n=nScrollpos,#tProcesses do
+        for n=nScrollPos,#tProcesses do
             if n == nCurrentProcess then
                 parentTerm.setTextColor( menuMainTextColor )
                 parentTerm.setBackgroundColor( menuMainBgColor )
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
```diff
@@ -115,7 +117,15 @@ local function redrawMenu()
         parentTerm.setCursorPos( 1, 1 )
         parentTerm.setBackgroundColor( menuOtherBgColor )
         parentTerm.clearLine()
-        for n=1,#tProcesses do
+        local nCharcou = 0
```
dan200: Change to "nCharCount"

src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
```diff
@@ -8,6 +8,8 @@ local nCurrentProcess = nil
 local nRunningProcess = nil
 local bShowMenu = false
 local bWindowsResized = false
+local nScrollpos = 1
```
dan200: Change to "nScrollPos"

src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
```diff
@@ -262,15 +282,26 @@ while #tProcesses > 0 do
...
-            for n=1,#tProcesses do
-                local tabEnd = tabStart + string.len( tProcesses[n].sTitle ) + 1
-                if x >= tabStart and x <= tabEnd then
-                    selectProcess( n )
-                    redrawMenu()
-                    break
+            if x == 1 and nScrollpos ~= 1 then
+                nScrollpos = nScrollpos - 1
+                redrawMenu()
+            elseif x == term.getSize() and bScrollRight == true then
```
dan200: remove "== true"

lupus590: Thinking optimisation, you may want to change the line to `elseif bScrollRight and x == term.getSize() then`. Doing this will mean that if bScrollRight then Lua will not bother calling the function to check if x is equal to the size of the terminal.

Wilma456:
@dan200 Changes are now done

Wilma456:
I added a patch from @BombBloke which allow the scroll with the mousewhell.

src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
```diff
@@ -315,7 +315,15 @@ while #tProcesses > 0 do
...
-        if not (bShowMenu and y == 1) then
+        if bShowMenu and sEvent == "mouse_scroll" and y == 1 then
+            if p1 == -1 and nScrollPos ~= 1 then
+                nScrollPos = nScrollPos - 1
+                redrawMenu()
+            elseif bScrollRight and p1 == 1 then
+                nScrollPos = nScrollPos + 1
+                redrawMenu()
+            end
+        elseif not (bShowMenu and y == 1) then
```
KnightMiner: I feel like this would be clearer as `elseif not bShowMenu or y ~= 1 then`, the `and` in a `not` is a bit more confusing to read. Arguably you could just move `sEvent == "mouse_scroll"` into a separate `if` and make this just an `else` statement though.

SquidDev:
Just trying this out today and noticed a issue with the scroll offset not changing when tabs are closed:



 - Open lots of tabs in multishell (I ran `bg` about 10 times).
 - Scroll to the far right but keep the focus in the first tab.
 - Start closing tabs.

I think the best thing to do would be to ensure the offset is valid before each menu redraw, clamping it if required.

Wilma456:
@SquidDev Thanks for reporting. The Bug is now fixed.

SquidDev:
Thought I'd commented this a couple of weeks ago but apparently I didn't. Sorry.

@Wilma456 The fix works, but I feel it would be nicer closing a tab scrolled the selection to the left again, so that the whole tab list is always full - rather than just having one tab on the far left.

PR #412, opened 2017-08-10
Add Check to textutils.tabulate/pagedTabulate

JakobDev:


src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -122,6 +122,9 @@ local function tabulateCommon( bPaged, ... )
     for n, t in ipairs( tAll ) do
         if type(t) == "table" then
             for n, sItem in pairs(t) do
+                if type( sItem ) ~= "number" and type( sItem ) ~= "string" then
+                    error( "textutils.tabulate/pagedTabulate only allow strings and numbers in the table", 3 ) 
+                end
```
dan200: Please only allow strings here, and format the error message more standardly.

BombBloke: Why only strings, all of a sudden? What happened to "no breaking changes"?

Wilma456: What should the message look like? e.g bad argument #Position.Position in Table (expected string, got foo)?

Wilma456:
@dan200 Changes are now done

PR #413, opened 2017-08-10
Give CC a custom User Agent

Wilma456:


SquidDev:
This'll only work for `http.get` and `http.post` as `http.request` doesn't go through `wrapRequest`. IMO it would be nicer to do this on the Java side: just add `connection.setRequestProperty( "user-agent", "computercraft" );` after this line:

https://github.com/dan200/ComputerCraft/blob/579f7443a8c47d71efd71da41744239b5019f320/src/main/java/dan200/computercraft/core/apis/http/HTTPRequest.java#L140

Wilma456:
Now Moved to Java Side

src/main/java/dan200/computercraft/core/apis/http/HTTPRequest.java
```diff
@@ -138,6 +138,7 @@ public void run()
 
             // Set headers
             connection.setRequestProperty( "accept-charset", "UTF-8" );
+            connection.setRequestProperty( "user-agent", "ComputerCraft/1.80" );
```
Guard13007: Does this overwrite a custom user-agent specified by user code? Or is this applied before custom headers?

SquidDev: Before custom headers, so one should be able to overwrite it.

Wilma456: I have tested it before I had made the Commit and it works great.

Restioson: Just need to be wary about updating the version number of ComputerCraft in the useragent if we include it there...

Wilma456: Of Course

SquidDev:
One could use `${version}` like `ComputerCraft.java`. That way it'll update automatically.

Wilma456:
@SquidDev Crazy. I had the same idea and testing it, while you wrote this.

Wilma456:
I tested it and it does not work. This is the User Agent:
```
ComputerCraft/${version}
```
To test it, I have downloaded a Page with this source code with wget from localhost:
```
<?php
echo $_SERVER['HTTP_USER_AGENT'];
?>
```

SquidDev:
@Wilma456 How did you run Minecraft? It'll only substitute everything if run via Gradle.

Wilma456:
With IntelJ Run>Minecraft Client

Restioson:
You could run it via a gradle run configuration like `gradlew runClient`

apemanzilla:
Why don't you just use `"ComputerCraft/" + ComputerCraft.getVersion()`?

Wojbie:
I think it makes more sense to hardcore something that will never change in same jar file? it does skip one concatenation and a function call per http request that way.

Lignum:
> Why don't you just use "ComputerCraft/" + ComputerCraft.getVersion()?

That just returns `"${version}"`, so it's the same thing.

apemanzilla:
I'd argue that using a method call is better practice than using Gradle to replace strings at compile-time.

SquidDev:
It's worth noting that template substitution is usei elsewhere:

https://github.com/dan200/ComputerCraft/blob/579f7443a8c47d71efd71da41744239b5019f320/src/main/java/dan200/computercraft/shared/computer/core/ServerComputer.java#L327

A couple of alternatives:
 - Have `public static final VERSION = ...` and use that in strings. This does introduce another way of getting the version though.
 - Use the host string as the user agent. This does mean the user agent doesn't follow the standald format however.

Wilma456:
I had changed it to ${version} because that's a little bit faster than call a function.

 I think _HOST as User Agent is a bad Idea, because it's not standard.

apemanzilla:
The overhead of calling `ComputerCraft.getVersion` would be negligible. IMO we should do something like what Squid suggested, with a constant for the version somewhere, to keep relying on text replacements to a minimum.

Guard13007:
I think the overhead argument is meaningless here relative to consistency
with the codebase, which someone else pointed out substitutions of that
nature are already everywhere.

Lignum:
> I think the overhead argument is meaningless here relative to consistency with the codebase, which someone else pointed out substitutions of that nature are already everywhere.

Yes, those kinds of microoptimisations shouldn't even be argued about in 2017. The JVM would undoubtedly inline the function call, making it boil down to the same thing anyway.

Restioson:
Although @SquidDev's alternative suggestions do make sense, I feel like they are out of scope for this PR. The current standard should be used

dan200:
This seems like it has the potential to break things, and CC has worked perfectly fine with it's current user agent for 5 years. Too risky and no benefit. Users are free to set their own user agent as a request property.

PR #416, opened 2017-08-13
Remove term.isColor() check from hello.lua

Wilma456:
Now that normal Computer had term.setTextColor() and can change the colour to gray, we don't need this check anymore.


BombBloke:
While you're at it, you might like to prevent the chance of the text colour matching the background colour.

dan200:
No thanks. Most of the shades of gray don't look good.

PR #417, opened 2017-08-13
Enable Syntaxhighlight on normal Computers in edit

Wilma456:
edit

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -40,10 +40,10 @@ if term.isColour() then
 else
     bgColour = colours.black
     textColour = colours.white
-    highlightColour = colours.white
-    keywordColour = colours.white
-    commentColour = colours.white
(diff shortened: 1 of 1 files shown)


SquidDev:
I like this in principle, but I'm really not sure this works with the default palette. I wonder if #307 is sufficient, that way people can configure the palette and colours themselves.

Wilma456:
I don't think that #307 will be merged in the next time. What do you mean with "default palette"? This are the default colours. I don't run  setPalleteColor().

SquidDev:
> I don't think that #307 will be merged in the next time.

It's not really a measure of when it'll be merged, rather if it'll be merged. There's no point duplicating work just because one PR will be merged sooner than another.

> What do you mean with "default palette"? This are the default colours. I don't run setPalleteColor().

I know, and I don't think the colours in the screenshot work. Strings are rather hard to read for instance. *However* if someone changed the palette a little, they might find a colour/palette combination that worked for them.

dan200:
No thanks.

PR #422, opened 2017-08-20
Switch lua.lua to use extutils.pagedPrint() for Tables

JakobDev:
With this PR lua.lua can how the full table, if the table is bigger than your Screen.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/lua.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/lua.lua
@@ -72,13 +72,13 @@ while bRunning do
                 if type( value ) == "table" then
                     local metatable = getmetatable( value )
                     if type(metatable) == "table" and type(metatable.__tostring) == "function" then
-                        print( tostring( value ) )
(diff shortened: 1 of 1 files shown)


SquidDev:
I feel we'd have to improve `pagedPrint` before doing this, maybe [something like this](https://github.com/dan200/ComputerCraft/issues/212#issuecomment-299712364). Otherwise you're going to be pressing keys a long time for big tables.

It might also be worth creating a modified serialise function, which can handle recursive structures and functions.

Wilma456:
Sorry for closing

apemanzilla:
Eh, I'm not really sure if this should be the default behavior. This is a REPL, you can always call `textutils.pagedPrint` yourself.

lupus590:
Perhaps have this as an option in settings?

BombBloke:
I'd rather it be the default behaviour, personally. It's hard to think of cases where I'd ever want part of a displayed table to be immediately hidden out the top of my screen (it's not like we can scroll back up), and if I *do* decide it'll take too long to scroll through a complete table, I can always just terminate the textutils call.

PR #424, opened 2017-08-21
Add more Checks

Wilma456:


src/main/resources/assets/computercraft/lua/rom/apis/rednet.lua
```diff
@@ -93,10 +99,19 @@ function send( nRecipient, message, sProtocol )
...
 
 function receive( sProtocolFilter, nTimeout )
+    if sProtocolFilter ~= nil and type( sProtocolFilter ) ~= "string" then
+        error( "bad argument #1 (expected string, got " .. type( sProtocolFilter ) .. ")", 2 )
+    end
+    if nTimeout ~= nil and type( nTimeoutl ) ~= "number" then
+        error( "bad argument #2 (expected number, got " .. type( nTimeout ) .. ")", 2 )
+    end
     -- The parameters used to be ( nTimeout ), detect this case for backwards compatibility
     if type(sProtocolFilter) == "number" and nTimeout == nil then
```
Wojbie: The check above makes this line pointless. It would make more sense to move checks after this if to ensure backwards compatibility.

src/main/resources/assets/computercraft/lua/rom/apis/rednet.lua
```diff
@@ -147,14 +165,17 @@ end
...
     end
     tHostnames[ sProtocol ] = nil
 end
 
 function lookup( sProtocol, sHostname )
     if type( sProtocol ) ~= "string" then
-        error( "expected string", 2 )
+        error( "bad argument #1 (expected string, got " .. type( sProtocol ) .. ")", 2 )
+    end
+    if type( sHostname ) ~= "string" then
```
Wojbie: Hostname can be nil in this function.

Wilma456:
The Checks are now fixed

@Wojbie Thank you for looking after that.

SquidDev:
@Wilma456 Could you create a new PR with fixing the above issue? Otherwise this could cause problems.

PR #426, opened 2017-08-23
Fix shell.setCompletionFunction()

Wilma456:
f the Path start with /, what is valid, shell.setCompletionFunction() doesn't works. this PR fix it.


Wilma456:
It uses now fs.combine()

dan200:
No thanks. The intended use of this command is that you pass in the absolute, canonical path to the file.

PR #427, opened 2017-08-23
Prevent shell from crashing with wrong completion function

Wilma456:


dan200:
No. It's better for a buggy completion function to crash and show it's error message than just silently fail.

PR #430, opened 2017-08-28
Add term.setPixel() - Manipulate single Pixel

Wilma456:
This is a very powerful function that is designed for Games - It allows you to manipulate a singe Pixel. Of course, you can do term.setCursorPos(x,y,) term.blit(text,col,col), but this function is faster, what is needed for Games. The Cursor Stay at his Position, if you call this function. Another Thing is, that you can change the Char/TextColour/BackgroundColour of a Pixel without changing the rest.
```
term.setPixel(10,10,"A","f","e")
--Give the Pixel 10/10 the Char A, with the Color balck and a red Background
term.setPixel(1,1,nil,nil,"b")
--Give the Pixel a blue Background, Char and Text Color will stay
```

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/TermAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/TermAPI.java
@@ -78,7 +78,8 @@ public class TermAPI implements ILuaAPI
             "setPaletteColour",
             "setPaletteColor",
             "getPaletteColour",
-            "getPaletteColor"
+            "getPaletteColor",
+            "setPixel"
         };
     }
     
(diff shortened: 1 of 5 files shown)


lupus590:
People could just make their own buffer which provides this kind of function.

Wilma456:
@lupus590 But it will not run as fast as the Java Code

src/main/java/dan200/computercraft/core/apis/TermAPI.java
```diff
@@ -298,6 +299,31 @@ public static void setColour( Terminal terminal, int colour, double r, double g,
...
+            {
+                // setPixel
+                int x = getInt( args, 0 ) - 1;
+                int y = getInt( args, 1 ) - 1;
+                String text = optString( args, 2, null );
+                String textColour = optString( args, 3, null );
+                String backgroundColour = optString( args, 4, null );
+                synchronized( m_terminal )
+                {
+                    if ( x <= 0 || x >= m_terminal.getWidth() || y <= 0 || y >= m_terminal.getHeight() ) {
```
SquidDev: The other terminal functions fail silently when out of range. This should probably do the same.

src/main/java/dan200/computercraft/core/apis/TermAPI.java
```diff
@@ -298,6 +299,31 @@ public static void setColour( Terminal terminal, int colour, double r, double g,
...
+                        throw new LuaException( "Position out of range" );
+                    }
+                    if ( text != null ) {
+                        m_terminal.setPixelChar( x, y, text.substring( 0, 1) );
```
SquidDev: Will these not fail if the string is empty? Also, seeing as we're just writing a single character, could the `Terminal.*` functions accept a `char` instead?

SquidDev:
Most of the times you're drawing to the screen, you're writing multiple pixels rather than a single one - and so `blit` is a superior function. Can you provide an example where you're only needing to change a few pixels each fame?

CrazedProgrammer:
This does not seem useful, considering term.blit exists. If you are going to draw pixel by pixel in games you're better off making your own buffer than pushing thousands of draw calls down the term stack.

BombBloke:
> Can you provide an example where you're only needing to change a few pixels each frame?

To my limited understanding, ComputerCraft redraws entire lines every time you output something anyway. Does this mean that it's just as fast to blit out entire lines regardless?

dan200:
This does not seem useful. Also, characters are not "pixels"

PR #431, opened 2017-08-30
Fix fatal Bug with coroutine.yield

Wilma456:
It is possible to froze all Computers in the world by running this code:
```
coroutine.yield = function() end
while true do
end
```
To only Way to break the freeze is to rejoin the world. With this fix, the Computer who run this code will crash after a few seconds and everything works normal.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -177,13 +177,18 @@ if string.find( _HOST, "ComputerCraft" ) == 1 then
     end
 end
 
+local nativeyield = coroutine.yield
+function coroutine.yield( sFilter )
+    return nativeyield( sFilter )
+end
+
 -- Install lua parts of the os api
 function os.version()
     return "CraftOS 1.8"
 end
 
 function os.pullEventRaw( sFilter )
-    return coroutine.yield( sFilter )
+    return nativeyield( sFilter )
 end
 
 function os.pullEvent( sFilter )
@@ -683,7 +688,7 @@ local nativeShutdown = os.shutdown
 function os.shutdown()
     nativeShutdown()
     while true do
-        coroutine.yield()
+        nativeyield()
     end
 end
 
@@ -691,7 +696,7 @@ local nativeReboot = os.reboot
 function os.reboot()
     nativeReboot()
     while true do
-        coroutine.yield()
+        nativeyield()
     end
 end
 


SquidDev:
This is one of those bugs I've never been able to reproduce, so I'd appreciate someone else chiming in here:

Has anyone been able to diagnose the root cause of this issue? Whilst infinite loops are inevitably going to block other computers from running, there shouldn't ever be a time when they stop them executing *forever*.

On a side note, does this occur with #163? It fixed some issues with yield prevention, so it might also fix this.

Wilma456:
@SquidDev It does not occur with #163. It has there the same effect as my fix on LUA Side.

KingofGamesYami:
@SquidDev the infinite loop is not actually required to do this.  Simplest reproduction steps:
```
>lua
lua>coroutine.yield = function() end
lua>_
```
Symptoms:
* Cursor will continue to blink
* No input will be accepted by the computer (ei typing or clicking has no effect)
* Other computers in the world will not turn on when clicked
* Other computers will not execute code or receive input
* On other computers, if the cursor was blinking when this was executed, it will continue to blink
* If you wait long enough, the computer this was executed on will error with something like:
`bios.lua:271: Too long without yielding` ... this leads to the computer printing "Goodbye" and presumably attempting to shut off... which doesn't work, and the computer cursor continues to blink.
image Other symptoms listed continue to occur.

Conclusion: The infinite loop has nothing to do with it.  Something about replacing coroutine.yield with a blank function does.

SquidDev:
Sadly this fix isn't going to fix everything - overriding `os.pullEventRaw = function() end` has the same effect. I haven't looked into this a lot, but here's a tiny bit more useful information:

One problem seems to be the lock on `m_machine` is never released, as mentioned [here](https://github.com/dan200/ComputerCraft/pull/218#issuecomment-300451639). Taking a selective thread dump backs this up [full dump here](https://gist.github.com/SquidDev/d406ea2aa7f5cde46e8734d1be5320d1).

```
java.lang.Thread.State: BLOCKED (on object monitor)
	at dan200.computercraft.core.computer.Computer$3.execute(Computer.java:871)
	- waiting to lock <0x00000000f7a4f7f8> (a dan200.computercraft.core.lua.LuaJLuaMachine)
	at dan200.computercraft.core.computer.ComputerThread$1$1.run(ComputerThread.java:109)
	at java.lang.Thread.run(Thread.java:745)

"Thread-250" #303 daemon prio=5 os_prio=0 tid=0x0000000019d74800 nid=0x2190 waiting for monitor entry [0x000000005e4ae000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at java.lang.Object.wait(Native Method)
	at java.lang.Object.wait(Object.java:502)
	at org.luaj.vm2.LuaThread$State.lua_resume(Unknown Source)
	- locked <0x00000000f7a4fc08> (a org.luaj.vm2.LuaThread$State)
	at org.luaj.vm2.LuaThread.resume(Unknown Source)
	at org.luaj.vm2.lib.CoroutineLib.invoke(Unknown Source)
	at dan200.computercraft.core.lua.LuaJLuaMachine.handleEvent(LuaJLuaMachine.java:225)
	at dan200.computercraft.core.computer.Computer$3.execute(Computer.java:871)
	- locked <0x00000000f7a4f7f8> (a dan200.computercraft.core.lua.LuaJLuaMachine)
	at dan200.computercraft.core.computer.ComputerThread$1$1.run(ComputerThread.java:109)
	at java.lang.Thread.run(Thread.java:745)
```

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -177,13 +177,18 @@ if string.find( _HOST, "ComputerCraft" ) == 1 then
     end
 end
 
+local nativeyield = coroutine.yield
+function coroutine.yield( sFilter )
+    return nativeyield( sFilter )
+end
+
```
BombBloke: Why override coroutine.yield itself??? Is that not simply overhead without any benefit?

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -177,13 +177,18 @@ if string.find( _HOST, "ComputerCraft" ) == 1 then
...
 -- Install lua parts of the os api
 function os.version()
     return "CraftOS 1.8"
 end
 
 function os.pullEventRaw( sFilter )
-    return coroutine.yield( sFilter )
+    return nativeyield( sFilter )
 end
```
BombBloke: `os.pullEventRaw = nativeyield` maybe

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -683,15 +688,15 @@ local nativeShutdown = os.shutdown
 function os.shutdown()
     nativeShutdown()
     while true do
-        coroutine.yield()
+        nativeyield()
```
BombBloke: Could consider setting a "neverResume" filter, or somesuch.

SquidDev: Ideally `os.shutdown` would just terminate/suspend the VM there, meaning we didn't need this code at all.

SquidDev:
I've done a little bit more debugging, and come up with the following reproduction case:

```lua
local co = coroutine.create(function()
    while true do end
end)
while coroutine.status(co) ~= "dead" do
    coroutine.resume(co)
end
```

As we're not calling any CC methods and we never yield, we won't ever get a timeout error, instead we'll just hit the debug hook:

https://github.com/dan200/ComputerCraft/blob/fd8837c631dfe7d0d78dace6bcc2c3c19416a7bd/src/main/java/dan200/computercraft/core/lua/LuaJLuaMachine.java#L76-L80

This'll cause our coroutine to yield into the parent loop, which just resumes the thread again - thus nothing is terminated. Changing the hook to execute every instruction [by changing this line to be 1](https://github.com/dan200/ComputerCraft/blob/fd8837c631dfe7d0d78dace6bcc2c3c19416a7bd/src/main/java/dan200/computercraft/core/lua/LuaJLuaMachine.java#L85) seems to fix this particular instance, but may cause problems elsewhere.

It'd be nice if @dan200 could chime in on this, as I'm not entirely sure why the timeout code does what it does.

dan200:
This is a bad fix and doesn't resolve the *actual* problem (computers that infinite loop and timeout aren't allowing other computers to resume afterwards). Please fix *that* instead.

Issue #343, opened 2017-06-29
Palette Control


BombBloke:
Three gripes here, though I'm not entirely certain about the best resolution to any of them, so I figure a discussion is in order following on from #193 / #197.

1) Lack of a "term.resetPalette()", or somesuch. Once the palette of a given display is changed, the only way to switch it back via code is to do it manually. Having to store CC's "original" palette in user-scripts (eg so they can ensure correct rendering or clean up on exit) feels messy, especially since it's prone to change from version to version, hence I feel this would be worth adding.

Catch is that buffers (eg window) *will* need CC's default palette encoded into them.

Another existing workaround *is* obviously for the user to reboot their system, but that's a) messy b) not an option when the display is an external monitor (those require either replacing or a chunk reload).

2) [colours.rgb8()](https://github.com/dan200/ComputerCraft/blob/fb5ba01e5a551a36d0366a126630d49a31c29785/src/main/resources/assets/computercraft/lua/rom/apis/colors.lua#L54) converts three 0-1 values to a single RGB24 representation (eg 1,1,1 to 0xFFFFFF; you can alternatively have it perform the reverse, but since term.setPaletteColour() does that automatically it's somewhat redundant).

RGB8 conventionally refers to a mode where one byte is used per *pixel* (as opposed to *intensity*), so the function should hence be renamed - colours.rgb24() seems obvious, but maybe someone has a better title?

It could also stand to be range-limited, as it currently just checks whether the values it's passed are numeric. term.setPaletteColour() is also missing limit checks - this makes it appear to be broken at a glance, as passing in something like 200, 100, 50 (which users *will* do, though they're currently invalid) gets you 56, 156, 206 (effectively inverting the values).

3) Use of 0-1 values to define intensities (versus 0-255) deserves re-raising; I've searched and searched but still cannot find a single implementation of an indexed display mode which does such a thing (and the *non*-indexed modes I *can* find certainly aren't from the era I've been assuming ComputerCraft to be themed upon).

If the goal is simply to "be different" then fine, but I believe coders and non-coders alike will be expecting standard triple-byte values (typically in the form of R,G,B), and I feel the current path will simply confuse people.

On the flip side, the only benefit I can see is marginally higher colour selection range - but we're talking about a 16 colour palette, so as benefits go it seems pretty slight.

Lignum:
[...]

PR #433, opened 2017-09-04
Palette improvements

Lignum:
After heavy debate, as seen in #343 and #197, I feel it's time for a new PR to fix some of the things that people were complaining about.

This includes:
- Functions like `term.(g|s)etPaletteColou?r` now deal with colours encoded as RGB8 instead of 0-1 range values.
- There is now a `term.nativePaletteColo(u)r( colour )`, as suggested by @SquidDev, which returns the default RGB8 values for a certain colour. @BombBloke I'd love to put this in `colours`, but writing this function in Java is a lot cleaner, since all the colours are stored in the `Colours` class.
- `colours.rgb8` is now neatly divided into `colours.packRGB` and `colours.unpackRGB`.

This is going to break a lot of stuff, but that's why we haven't released yet, after all.

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/TermAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/TermAPI.java
@@ -10,6 +10,7 @@ import dan200.computercraft.api.lua.ILuaContext;
 import dan200.computercraft.api.lua.LuaException;
 import dan200.computercraft.core.computer.IComputerEnvironment;
 import dan200.computercraft.core.terminal.Terminal;
+import dan200.computercraft.shared.util.Colour;
 import dan200.computercraft.shared.util.Palette;
 import org.apache.commons.lang3.ArrayUtils;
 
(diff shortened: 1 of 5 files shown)


viluon:
Thank you @Lignum.

Since the API is being changed again, I'd like to suggest adding a means of changing multiple palette entries at once. A couple of `term.(s|g)Palette` functions accepting/returning a table would enable for effortless palette swaps.

Lignum:
@viluon No problem! Passing tables was a feature in the original PR for some time, [but we decided we didn't need it](https://github.com/dan200/ComputerCraft/pull/197#discussion_r115137184). As such, I won't bother putting any effort into adding it back unless @dan200 has changed his mind.

BombBloke:
> @BombBloke I'd love to put this in `colours`, but writing this function in Java is a lot cleaner, since all the colours are stored in the `Colours` class.

Fair 'nuff. At the end of the day the location is less important than the availability. Thanks for your efforts!  :)

PR #434, opened 2017-09-05
Fix help.lua

Wilma456:


SquidDev:
This is a bug in the `io` library, rather than `help.lua`. You can see that `_sMode` is meant to default to `r`:

https://github.com/dan200/ComputerCraft/blob/fd8837c631dfe7d0d78dace6bcc2c3c19416a7bd/src/main/resources/assets/computercraft/lua/rom/apis/io.lua#L74

You should probably push an additional fix to #429 instead.

Wilma456:
The other PR has been updated

PR #436, opened 2017-09-06
Add autocompletion for Keywords in textutils.complete()

JakobDev:
At the moment, textutils.complete will not complete Keywords. But this PR add a 3rd Argument with allows to enable autocompletion with Keywords.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
@@ -366,15 +366,19 @@ function urlEncode( str )
 end
 
 local tEmpty = {}
-function complete( sSearchText, tSearchTable )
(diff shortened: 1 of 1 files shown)


Wojbie:
Could you provide use case for this addition?

Wilma456:
Lua shell or edit

SquidDev:
@Wilma456 This doesn't appear to do what you want it to do. There are a couple of problems:

 - This reintroduces the `do` completing to `dofile` issue (which was previously fixed by https://github.com/dan200/ComputerCraft/commit/76a3562d58803c4a1d5d31a851797c23a68c0c53).
 - This won't complete keywords on the "global" level (so `i` will not complete to `if`). 
 - If you have a field with a keyword (such as `x = { ["if"] = 1 }`) then `x.` will complete to `x.if`. However, this is a syntax error and so is probably not intentional behaviour.

PR #437, opened 2017-09-07
Make lua.lua display all tables

JakobDev:
At the moment, lua.lua can only display who are work with textutils.serialise(). With this PR, all table will be displayed.

Here's a screen how it looks, if textutils.serialise() not working:
cclite_1504794266

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/lua.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/lua.lua
@@ -19,15 +19,25 @@ local tEnv = {
 setmetatable( tEnv, { __index = _ENV } )
 
 local function displayTable( tDisplay )
+    local tSeen = {}
     local sShow = "{\n"
-    for k, v in pairs( tDisplay ) do
+    for k, v in ipairs( tDisplay ) do
         sShow = sShow .. "  "
-        if type( v ) == "number" or type( v ) == "boolean" then
-            sShow = sShow .. k .. " = " .. tostring( v ) .. ",\n"
-        elseif type( v ) == "string" then
-            sShow = sShow ..k .. ' = "' .. v .. '",\n'
+        if type( v ) == "string" then
+            sShow = sShow .. '"' .. v .. '",\n'
         else
-            sShow = sShow .. k .. " = ".. type( v ) .. ",\n"
+            sShow = sShow .. tostring( v ) .. ",\n"
+        end
+        tSeen[k] = true
+    end
+    for k, v in pairs( tDisplay ) do
+        if not tSeen[k] then
+            sShow = sShow .. "  "
+            if type( v ) == "string" then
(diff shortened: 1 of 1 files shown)


SquidDev:
I've been meaning to PR something like this for a while, so 👍 for getting on with it :). A couple of thoughts though:

 - IMO it'd be nicer to print the result of `tostring` instead of just the type. This means you can tell when two functions/coroutines/etc... are the same.
 - Ideally the above table would have the array-part entries be continuously, without their key. For instance, in your example you'd have: `{ "Hello", function, bol = true, tab = table, num = 1 }`. This is the [current behaviour of `textutils.serialize`](https://github.com/dan200/ComputerCraft/blob/master/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua#L213-L232).
 - It might be nice to use `__tostring` on child tables too, like done [here](https://github.com/dan200/ComputerCraft/blob/master/src/main/resources/assets/computercraft/lua/rom/programs/lua.lua#L74-L75).
 - This will fail on recursive structures, for instance `_G`. If one encounters a node you've already hit, then you should probably just use `tostring( tDisplay )`.

Wilma456:
@SquidDev You mean, if __tostring is a function in a child table then call it and print the result? btw. it fails not with _G.

SquidDev:
@Wilma456 No, it should look in the metatable for the `__tostring` implementation.  See the linked code on how to implement that.

Ahh, you're not even recursing over nested tables, so you won't get any issues. Ideally you'd recur until you find something you've hit already, like textutils.serialize does. In fact, I'd use that as the basis for the implementation, and build from there.

Wojbie:
Is this a good idea to not show contents of nested tables?

lupus590:
@Wojbie has a good point, `_G` will take a while to scroll through, and if I want to see what is in a subtable I could always type `table.subtable` into the Lua console

Wilma456:
@lupus590 So you think, I should not show the content of subtables?

Wilma456:
What did you think, what's the best way, to show, that it is a table who provide a __tostring() and not a string?

btw:
Should I move this function to textutils to give other programs access to it?

lupus590:
Part of the point of the metamethod is that it's discreet about what it's doing.

As for moving to textutils, maybe. I'll have a look at it properly to make up my mind.

Wilma456:
Changes are now done!

src/main/resources/assets/computercraft/lua/rom/programs/lua.lua
```diff
@@ -19,15 +19,25 @@ local tEnv = {
 setmetatable( tEnv, { __index = _ENV } )
 
 local function displayTable( tDisplay )
```
lupus590: The function is called displayTable but returns a string instead of doing its name sake.

If the only use of the resulting string is to print it (for this program at least) then in my mind the function should do the actual printing. If the output string does get used for something else then this function should be renamed.

PR #438, opened 2017-09-10
Create uiutils.lua and add uiutils.read

hugeblank:
As mentioned in the commit:

This API is planned to be the home for slightly advanced, but still low level functions that assist the user in making graphically pleasing programs. For example, a button API would be perfect here.

This was (and still is being) discussed on #387 , and I decided to put it into effect.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/uiutils.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/uiutils.lua
@@ -1,0 +1,282 @@
+function read( _sReplaceChar, _tHistory, _fnComplete, _sDefault, _nLimit )
+    if _sReplaceChar ~= nil and type( _sReplaceChar ) ~= "string" then
+        error( "bad argument #1 (expected string, got " .. type( _sReplaceChar ) .. ")", 2 ) 
+    end
+    if _tHistory ~= nil and type( _tHistory ) ~= "table" then
+        error( "bad argument #2 (expected table, got " .. type( _tHistory ) .. ")", 2 ) 
+    end
+    if _fnComplete ~= nil and type( _fnComplete ) ~= "function" then
+        error( "bad argument #3 (expected function, got " .. type( _fnComplete ) .. ")", 2 ) 
+    end
+    if _sDefault ~= nil and type( _sDefault ) ~= "string" then
+        error( "bad argument #4 (expected string, got " .. type( _sDefault ) .. ")", 2 ) 
+    end
+    if _nLimit ~= nil and type( _nLimit ) ~= "number" then
+        error( "bad argument #5 (expected number, got " .. type( _nLimit ) .. ")", 2 ) 
+    end
+    term.setCursorBlink( true )
+
+    local sLine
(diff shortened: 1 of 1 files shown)


SquidDev:
> Perhaps a XXX function which takes a table to allow for a wide range of customisations would satisfy these needs.

As @Lignum says, I think a table would be better idea here. This way we can extend its functionality without having argument spaghetti.

Wilma456:
Just a suggestion: Why not create a folder like /rom/moduls for for the new require() and put the API there? The other APIs must be compatible to os.loadAPI() for backwards compatibility, but new APIs should be compatible to require().

Lignum:
Creating a `uiutils` API would be an open invitation for more PRs adding in UI features which should really be third party APIs. If we do add functions like this, they should go in `textutils`, since that's home to existing UI-related functions like `pagedPrint`, and even then, it should not contain a copy of `read`, but a completely new, more flexible function.

@Wilma456 The APIs in `/rom/apis` are automatically loaded anyway, so it doesn't really matter.

hugeblank:
@Lignum The UI features must be low level in order to fit what the purpose of this API is going to be.

Lignum:
A button API is a rather big abstraction to be considered low level, if that's your point of reference. In fact, I find that ComputerCraft is already abundant enough in terms of "low level" UI functions.

Either way, there should not be an exact copy of `read` in here.

hugeblank:
Personally I don't find this to be that big of an abstraction, it's very limited in what it can do, and still allows better button API's (like touchpoint) to exist. 
```
local buttons = {}
button = {
	add = function(x1, y1, x2, y2, name)
		buttons[#buttons+1] = {x1, y1, x2, y2, name}
	end,
	check = function()
		_, _, cx, cy = os.pullEvent("mouse_click")
		for i = 1, #buttons do
			if buttons[i][1] <= cx and buttons[i][2] <= cy and buttons[i][3] >= cx and buttons[i][4] >= cy then
				return buttons[i][5]
			end
		end
	end,
        clear = function()
		buttons = {}
	end
}
```

Lignum:
> Personally I don't find this to be that big of an abstraction, it's very limited in what it can do

Arguably, it is up to personal taste, but for an OS, I feel this is almost certainly too much. And if it is so limited that you could write it up in 5 minutes, then why have it at all?

> and still allows better button API's (like touchpoint) to exist.

Exactly, those exist, meaning we don't have to clutter CraftOS.

> If anything, the addition of uiutils will up the standard quality of APIs.

That's rather far fetched.

> so if what lignum and now you are saying
to:
-hide all characters with "*"
-output "test" ("****") before allowing typing
-start the scrolling point at 20
the parameters would look like:
read({_sReplaceChar="*", _sDefault = "test", _nLimit=20})

Yes, but as a completely redesigned function.

dan200:
I'm not sure this is a good fit for a builtin library.

hugeblank:
@dan200 Would you be okay with this function being put into `textutils`? I plan on adding @Lignum's  idea in place as well if it would be okay in `textutils`.

PR #440, opened 2017-09-11
Make io.write() accept multiple args

Wilma456:
This is just to bring the io API from CC close the the io API from normal lua, which accept multiple args for io.write().


SquidDev:
You should probably use `select('#', ...)` instead of wrapping it in a table and calling `ipairs`. This way it won't break when you pass `nil` to it. This is what the `print` function does:

https://github.com/dan200/ComputerCraft/blob/1fdfcdb5f2fdaef5c4039f53098e07c16fd9dd3b/src/main/resources/assets/computercraft/lua/bios.lua#L274-L276

Wilma456:
@dan200 It uses now select()

PR #441, opened 2017-09-12
Fix a non-existent method being used in paintutils

SquidDev:
`:readAll()` was being used, but the handle came from the `io` library, which doesn't have such a thing.


dan200:
This certainly fills be me with confident that people are actually testing their PRs! Merging.

PR #444, opened 2017-09-12
Simple fix to os.sleep()

Wilma456:
If you call os.sleep() with a wrong argument, you will get a error in bios.lua line 679. By adding return, the error will show the line in the program, that calls os.sleep() with the wrong argument.


BombBloke:
Cleaner to just do "os.sleep = sleep".

PR #446, opened 2017-09-12
Add folder /rom/modules

Wilma456:
This folder is made, to give Resourcepack Authors a place to add modules like the autorun folder is a place, where Resourcepack authors can place their autorun files. And of course, this folder can be used in further PRs.


lupus590:
If I recall correctly, the API system has folders for pocket and advanced computers.

PR #447, opened 2017-09-12
Correct minor typo in rednet.receive

BombBloke:
Caused attempts to set a time-out value to throw "expected number, got number".


lupus590:
how long has that been in there?

Wilma456:
@lupus590 Since #424. Sorry for that typo.

PR #448, opened 2017-09-12
Fix check of write()

Wilma456:
if you call write(nil), you will get the error "bios.lua:229: bad argument: string expected, got nil", so nil is not a valid argument for write() and should be removed.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -206,8 +206,8 @@ function sleep( nTime )
 end
 
 function write( sText )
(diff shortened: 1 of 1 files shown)


lupus590:
In my opinion, trying to write nil should error. It gives an extra chance for users to notice that they have mistyped a variable name or miss assigned it.

lupus590:
having the 'error blame' go to the calling program would be better though.

Wilma456:
@lupus590 I have removed nil from the allowed args, so you will get a bad argument error when you call it with nil.

lupus590:
I see now, the arg check uses backwards logic to me.

PR #449, opened 2017-09-12
Advanced autocompletion logic additions and handling for paths with spaces.

Wojbie:
This PR is result of few hours of brainstorming how to add some kind of system that would allow to add quotes to file paths if one of elements contains space.
In kinda sprawled into advanced autocompletion mode that allows above and could possible be used by more advanced users to add complicated and interesting autocompletion logic to theirs programs while still keeping backwards compatibility with current system.

It has basic 5 parts:

1. Addition of ability for read to accept table instead of string as element of return table from autocompletion function. Said table is defined as {"line","suggestion","line","suggestion","line", ect}.
This table is treated as new contents of sLine - this forced addition of sLine as one af arguments of autoc funcion itself. End results look like this: (in case of this image generated table was ('cp ','"','stup','id file.lua"'}
file
2. Addition to fs.complete ability to detect elements with spaces or existence of quote at begining of path already and handling it correctly. This allows results like this:
file
folder
3. Fix to current bug in read function when it attempted to autocomplete a word that was already ended by closing quote:
rom
Now it will not attempt to autocomplete in this case.
4. Change in  logic of autocompletion when in _sReplaceChar is set. This makes it so the suggestion is visible to end user. This might make it somehow usefull in edge cases? More useful then current way of make it all stars anyways and let people guess what it says. 
secret
secret2
5. Adds copy of completeMultipleChoice function from startup to textutils API. It kinda makes sense to make it one of default functions considering that everyone making autocomplete is coping it in their code.

Anyhow that was a wall of text (and images). Thought/opinions/suggestions?

EDIT: Almost forgot - Thanks to @SquidDev and @BombBloke for putting up with me and giving suggestions along the way.

EDIT2: Code changes are kinda messed up due to few indentation changes. Sorry.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -814,7 +814,7 @@ end
 
 -- Install the lua part of the FS api
 local tEmpty = {}
-function fs.complete( sPath, sLocation, bIncludeFiles, bIncludeDirs )
+function fs.complete( sPath, sLocation, bIncludeFiles, bIncludeDirs, sLine )
     if type( sPath ) ~= "string" then
         error( "bad argument #1 (expected string, got " .. type( sPath ) .. ")", 2 ) 
     end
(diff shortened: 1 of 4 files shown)


Wojbie:
Like i like this code but i realize its kinda messy and overcomplicated things for now.
Closing until i find time to make it better.

PR #455, opened 2017-09-16
Add read() to Filehandle

Wilma456:
This PR allows you to just read single Bytes of a file like in real LUA. At the moment you can only read a full line or the full file. And this help to make the io API more complete.

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/handles/EncodedInputHandle.java
+++ b/src/main/java/dan200/computercraft/core/apis/handles/EncodedInputHandle.java
@@ -6,6 +6,8 @@ import dan200.computercraft.api.lua.LuaException;
 import javax.annotation.Nonnull;
 import java.io.*;
 
+import static dan200.computercraft.core.apis.ArgumentHelper.*;
+
 public class EncodedInputHandle extends HandleGeneric
 {
     private final BufferedReader m_reader;
@@ -49,6 +51,7 @@ public class EncodedInputHandle extends HandleGeneric
             "readLine",
             "readAll",
             "close",
+            "read",
         };
     }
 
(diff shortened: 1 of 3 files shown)


SquidDev:
It'd probably be neater to use [`Reader.read(char[], int, int)`](https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html#read-char:A-int-int-) and then create [a new string with a given length](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#String-char:A-int-int-), instead of looping a given number of times. See how `BinaryInputHandle` does it:

https://github.com/dan200/ComputerCraft/blob/1c8480a3291d2a6ac28fb8f633eab5edb130688d/src/main/java/dan200/computercraft/core/apis/handles/BinaryInputHandle.java#L53-L57

Wilma456:
@SquidDev Now done

src/main/java/dan200/computercraft/core/apis/handles/EncodedInputHandle.java
```diff
@@ -5,6 +5,7 @@
 
 import javax.annotation.Nonnull;
 import java.io.*;
+import java.util.Arrays;
```
SquidDev: I don't think this import is being used at all.

src/main/java/dan200/computercraft/core/apis/handles/EncodedInputHandle.java
```diff
@@ -110,22 +111,16 @@ private static BufferedReader makeReader( InputStream stream, String encoding )
...
+                    int count = optInt( args, 0, 1 );
+                    if( count <= 0 || count >= 1024 * 16 )
+                    {
+                        throw new LuaException( "Count out of range" );
                     }
-                    return new Object[] { result.toString() };
+                    char[] bytes = new char[ count ];
+                    count = m_reader.read( bytes );
+                    if( count < 0 ) return null;
+                    String str = new String( bytes );
```
SquidDev: This should be `new String(bytes, 0, count)` instead. Otherwise you may end up with lots of `\0`s at the end of the string.

src/main/resources/assets/computercraft/lua/rom/apis/io.lua
```diff
@@ -91,6 +91,8 @@ function open( _sPath, _sMode )
 					return file.readLine()
 				elseif sFormat == "*a" then
 					return file.readAll()
+                elseif _G.type( sFormat ) == "number" then
+                    return file.read( sFormat )
```
Lignum: Your code seems to be caught in the wind here

PR #456, opened 2017-09-16
Set errorlevel for "Unsupported format" to 2

Wilma456:


SquidDev:
There seems to be a fair few places where the error level is wrong. I wonder if it's worth creating one "mega PR" with all the fixes in, rather half a dozen separate ones.

PR #459, opened 2017-09-18
Better String for optional Args in ArgumentHelper

Wilma456:
If a Arg optional, you get now the error "type or nil expected" and not only "type expected"

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/ArgumentHelper.java
+++ b/src/main/java/dan200/computercraft/core/apis/ArgumentHelper.java
@@ -146,7 +146,7 @@ public final class ArgumentHelper
         }
         else
         {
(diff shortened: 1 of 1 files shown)


SquidDev:
It's worth noting that PUC Lua and LuaJ do `X expected, got Y`, rather than `X or nil expected, got Y` (see `table.concat({}, {})` for instance). I don't know how much we value being consistent with the built-in Lua libraries.

BombBloke:
Going right through the ROM files is probably best left until some sort of decision is reached regarding #397 (even if that decision ends up being "let's forget about it"). While I believe the current gmatch'ing implementation it offers is... suboptimal, I do see some benefit from the idea of being able to reformat most type-related errors from a single point in the code.

PR #460, opened 2017-09-18
Add ability to remove first line in edit.lue

Wilma456:


lupus590:
What's the point of this?

SquidDev:
Wilma, just a comment on this and a couple of your other PRs: it'd be good to explain *why* this change is a good idea. Whilst the description and code explain what changes you've made, it isn't always clear why those changes are needed. 

Maybe just a couple of sentences on where this feature would be useful wouldn't go amis.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -656,6 +656,13 @@ while bRunning do
                     table.remove( tLines, y )
                     setCursor( sPrevLen + 1, y - 1 )
                     redrawText()
+                elseif y == 1 and type( tLines[2] ) == "string" then
+                    -- Remove first line
+                    local sPrevLen = string.len( tLines[2] )
```
1lann: Shouldn't this be `nPrevLen`?

BombBloke:
"If cursor is at 1,1 and user hits backspace, move first line's content to the end of the second line."

Is this actually what's intended?? I can't say I've come across such a behaviour before.

Wilma456:
If you want to remove the first line of a file, you have to open the file in a editor with a clipboard, copy second line, paste it's content at the first line and delete the second line. With this PR you no longer need this Workaround.

SquidDev:
@Wilma456 But why only the first line? This PR doesn't help if you want to delete the second line, or the third line, or the last line. I think it would be better to have some generalised way of highlighting text and deleting it, rather than one very specific action.

BombBloke:
If you want to remove the *first* line of a file, all you have to do is use the *Delete* button instead of *Backspace* - same as with most any text editor.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -656,6 +656,13 @@ while bRunning do
...
+                    tLines[2] = tLines[2] .. tLines[1]
```
Wojbie: Soo if i read this operation right then by pressing backspace at beginning of the first line you would cause contents of first line to be added to end of second line and then first line is removed. Is that something that exists in some text editor? Personally i never saw one acting like that.

dan200:
What a weird PR. You can delete the first line of a text file just fine by any of the following methods:

1. Go to the end of the first line and hold backspace
2. Go to the beginning of the second line and hold backspace
3. Go to the beginning of the first line and hold delete

Do any of those approaches not work for you @Wilma456 ?

PR #461, opened 2017-09-21
Add more escape sequences

Wilma456:
For Details see https://www.lua.org/pil/2.4.html

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -242,7 +242,41 @@ function write( sText )
             sText = string.sub( sText, 2 )
         end
         
-        local text = string.match( sText, "^[^ \t\n]+" )
+        local verticaltab = string.match( sText, "^\v" ) or string.match( sText, "^\f" )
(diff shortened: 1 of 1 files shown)


SquidDev:
The trouble with this is that most of these characters have a different meaning under CC's charset. `\f` and `\v` are [gender symbols](https://en.wikipedia.org/wiki/Gender_symbol), `\a` and `\b` are various dots. It seems inconvenient to require people to use `term.write` just to use those characters. We'd also have to consider the matter of backwards compatibility, as various programs _may_ already be using such symbols in a call to `write`. 

All in all, I'm really not sure whether the added compatibility is worth the potential issues. These sorts of things are probably better left to projects like BSRocks and its ANSI emulation layer.

lupus590:
While conforming to the vanilla Lua standard is nice, CC is its own thing and backwards compatibility is more important in my opinion.

Perhaps have this as a config option? Per CC computer config option?

SquidDev:
@lupus590 If anything, this should be a separate function - it feels rather dirty to be putting this behaviour behind a config option.

PR #462, opened 2017-09-23
Add turtle.getEquipLeft() and turtle.getEquipRight()

JakobDev:
This functions allows you to get to equipped Item.

I know, this is already possible with unequip the Item, get Item Detail and reequip it. But the current methode has 2 Problems in my Opinion:
1. The User see, that the Item is unequip and reequip.
2. It will not work, if the Turtle had a full Inventory.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
@@ -11,6 +11,7 @@ import dan200.computercraft.api.lua.LuaException;
 import dan200.computercraft.api.turtle.ITurtleAccess;
 import dan200.computercraft.api.turtle.ITurtleCommand;
 import dan200.computercraft.api.turtle.TurtleSide;
+import dan200.computercraft.api.turtle.ITurtleUpgrade;
 import dan200.computercraft.core.apis.IAPIEnvironment;
 import dan200.computercraft.core.apis.ILuaAPI;
 import dan200.computercraft.shared.turtle.core.*;
@@ -107,6 +108,8 @@ public class TurtleAPI implements ILuaAPI
             "inspectUp",
             "inspectDown",
             "getItemDetail",
+            "getEquipLeft",
+            "getEquipRight",
         };
     }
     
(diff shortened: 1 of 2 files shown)


src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
```diff
@@ -446,6 +449,40 @@ else if( side.equalsIgnoreCase( "right" ) )
                     return new Object[] { null };
                 }
             }
+            case 42:
```
SquidDev: It'd probably be nicer to split this into a separate function so one is not duplicating functionality. Maybe do something like [the main `equip*` functions](https://github.com/Wilma456/ComputerCraft-1/blob/f5812c5d29461080fed1d4044939e23e88fd5063/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java#L404-L413), and create a separate `TurtleGetEquipCommand` class.

src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
```diff
@@ -446,6 +449,40 @@ else if( side.equalsIgnoreCase( "right" ) )
...
+                if( upgrade != null ) 
+                {
+                    ItemStack stack = upgrade.getCraftingItem();
+                    Item item = stack.getItem();
+                    String name = Item.REGISTRY.getNameForObject( item ).toString();
+                    return new Object[] { name };
+                }
+                else 
+                {
+                    return new Object[] { null };
```
Lignum: It's probably nicer to just do `return null;`.

SquidDev:
There's a bit of me which doesn't like leaking "internal" data such as upgrade IDs. I guess we already do this with `speaker.playSound()` and `turtle.inspect()` though so it's not the end of the world.

If we are adding this, would it also be worth adding such a method to the `pocket` API? This way we can keep the two libraries mostly in sync.

It might also be nice to have a program called something like `upgrades` which acts similarly to the existing `peripherals`: printing out the currently available upgrades.

PR #468, opened 2017-09-26
Add commands.complete()

JakobDev:
This function complete a Command like the Chat. e.g. commands.complete("time ") will return the table {"set","add","query"}. 

Another Usage of this function is getting information from MC. e.g. commands.complete("playsound ") will return a list of all available sounds and commands.complete("setblock ~ ~ ~  ") will return a list of all blocks.


SquidDev:
Whilst this'd be a really neat feature to have, I'm slightly worried about thread safety. The completion generator will need to run on the server thread, so we'll probably want `complete` and `completeAsync` variants. 

However, this'd then make it unusable as a completion provider to `read` as it'd introduce a delay in typing. One could probably adjust the completion provider to rus in parallel to reduce the latency, but I'm not sure how much it's worth it.

thatcraniumguy:

PR #469, opened 2017-09-27
Add more Recipes to Recipebook

### LINKED ISSUE #467: Upgrade Recipes are not listed
Since 1.12 you can see the recipes. Normal Recipes of CC are listed, but not upgrade recipes. A Example for  Upgrade Recipe is to craft a mining turtle with a turtle and a Pickaxe.

JakobDev:
With a little help of @SquidDev . Just write a comment, if I had forget one. Closes #467.


SquidDev:
Looks good. A couple of things worth changing though:

 - Remove the player head recipes. I'm pretty sure these are an Easter egg, and so should be hidden.
 - Drop all the junk in advancement names. For instance, `_1`, `computercraft` and `minecraft`. It makes sense for the recipes to have those names as they are auto-generated, but it's nicer if the advancements have "cleaner" names.
 - I only really thought about it, but can merge all the recipes for one item into a single advancement. For instance, [here's all the pocket upgrades](https://gist.github.com/SquidDev/05bdf7b8e11e29c7671c9474398feea5). This is great for the dye recipes, as they are all identical.
 - For upgrades, it might be worth having the "upgrade item" (so pickaxe) as an additional criteria. Obviously this doesn't work if you have multiple recipes in an advancement.

There's also a couple of issues which will result in the advancements not loading correctly:

 - You reference a `computercraft:bow` upgrade, but there's no such thing.

If you enter a world, you should get some diagnostic error messages detailing what problems there are. Fix those errors and rinse and repeat until everything's fixed. It's a painful business but sadly only the first error is displayed.

---

On a totally unrelated note: `advanved_modem`. How did that get through code review? This is seriously embarrassing :/.

Wilma456:
The Recipes are autogenerated with the completion of /recipe. I don't know, if It works without _1, computercraft or minecraft. I think, I will wait for a comment of @dan200, if I should put everything in one  advancement or add the upgrade Items as criteria.

I had worked on a shooting turtle, so my script generated a Advancement for it and had forget to remove it, because it exits on my Computer and I get no error.

I will look after he recipes tomorrow and then I will remove the Shooting Turtle Recipe.

Wilma456:
I had now removed the bow upgrade recipe. For more steps I will wait for the opinion of @dan200.

thatcraniumguy:

PR #470, opened 2017-09-28
Add textutils.prompt()

hugeblank:
textutils.prompt is the brainchild of numerous people that have attempted to contribute to ComputerCraft. SquidDev has mentioned this in various PR's as "super-duper READ-O-MATIC 4000 with turbo text entry", and to be quite honest, that's what it is. It is the read function on steroids, if you will.
 
Features:
- All arguments are put into a table, so that they don't have to be put in any particular order, like how it is in read().
- Type error protection has been condensed into a more understandable form.
- Adds a few arguments, including those from read(), and others that would never make it into read():
  - replaceChar (_sReplaceChar), history (_tHistory), complete (_fnComplete), prefix (_sDefault); all arguments from read
  - limit (number): As defined in #387
  - newline (boolean): As defined in #386
  - completeBGColor (number): Changes the Background color of the suggested completion values
  - completeTextColor (number): Changes the Text color of the suggested completion values

What's really great about this function being used over read is that it allows for both command lines, and applications with a GUI alike to have greater customizability, over read(). Hopefully in the future this may be used in replacement of read(). 
Along with this, my hope with prompt is to have people that have wanted a specific feature implemented into read(), be put here. As there is no particular order to the arguments, it's all subject to the users need of it.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
@@ -1,3 +1,290 @@
+function prompt( _tOptions )
+    local tVerify = {replaceChar = "string", history = "table", complete = "function", prefix = "string", limit = "number", newline = "boolean", completeBGColor = "number", completeTextColor = "number"}
+    if not _tOptions then
+        _tOptions = {}
+    end
+    for k, v in pairs(tVerify) do
+        if _tOptions[k] ~= nil and type( _tOptions[k] ) ~= v then
+            error( "bad argument " .. k .. " (expected " .. v .. " got ".. type( _tOptions[k] ) .. ")", 2)
+        end
+    end
+    
+    term.setCursorBlink( true )
+
+    local sLine
+    if type( _tOptions.prefix ) == "string" then
+        sLine = _tOptions.prefix
+    else
+        sLine = ""
+    end
+    local nhistoryPos
+    local nPos = #sLine
+    if _tOptions.replaceChar then
+        _tOptions.replaceChar = string.sub( _tOptions.replaceChar, 1, 1 )
+    end
+
+    local tCompletions
+    local nCompletion
+    local function recomplete()
+        if _tOptions.complete and nPos == string.len(sLine) then
+            tCompletions = _tOptions.complete( sLine )
+            if tCompletions and #tCompletions > 0 then
+                nCompletion = 1
+            else
+                nCompletion = nil
+            end
+        else
+            tCompletions = nil
+            nCompletion = nil
+        end
+    end
+
+    local function uncomplete()
+        tCompletions = nil
+        nCompletion = nil
+    end
+    
+    local w
+    if not _tOptions.limit then
+        w = term.getSize()
+    else
+        w = _tOptions.limit
+    end
+    
+    local sx = term.getCursorPos()
+
+    local function redraw( _bClear )
+        local nScroll = 0
+        if sx + nPos >= w then
+            nScroll = (sx + nPos) - w
+        end
+
+        local cx,cy = term.getCursorPos()
+        term.setCursorPos( sx, cy )
+        local sReplace = (_bClear and " ") or _tOptions.replaceChar
+        if sReplace then
+            term.write( string.sub( string.rep( sReplace, math.max( string.len(sLine) + nScroll, 0 ) ),  nScroll + 1, nScroll + w ) )
+        else
+            term.write( string.sub( sLine, nScroll + 1, nScroll + w ) )
+        end
+
+        if nCompletion then
+            local sCompletion = tCompletions[ nCompletion ]
+            local oldText, oldBg
+            if not _bClear then
+                oldText = term.getTextColor()
+                oldBg = term.getBackgroundColor()
+                if not _tOptions.completeTextColor then
+                    term.setTextColor( colors.white )
+                else
+                    term.setTextColor( _tOptions.completeTextColor )
+                end
+                if not _tOptions.completeBGColor then
+                    term.setBackgroundColor( colors.gray )
+                else
+                    term.setBackgroundColor( _tOptions.completeBGColor )
+                end
+            end
+            if sReplace then
+                term.write( string.rep( sReplace, string.len( sCompletion ) ) )
+            else
+                term.write( sCompletion )
+            end
+            if not _bClear then
+                term.setTextColor( oldText )
+                term.setBackgroundColor( oldBg )
+            end
+        end
+
+        term.setCursorPos( sx + nPos - nScroll, cy )
+    end
+    
+    local function clear()
+        redraw( true )
+    end
+
+    recomplete()
+    redraw()
+
+    local function acceptCompletion()
+        if nCompletion then
+            -- Clear
+            clear()
+
+            -- Find the common prefix of all the other suggestions which start with the same letter as the current one
+            local sCompletion = tCompletions[ nCompletion ]
+            sLine = sLine .. sCompletion
+            nPos = string.len( sLine )
+
+            -- Redraw
+            recomplete()
+            redraw()
+        end
+    end
+    while true do
+        local sEvent, param = os.pullEvent()
+        if sEvent == "char" then
+            -- Typed key
+            clear()
+            sLine = string.sub( sLine, 1, nPos ) .. param .. string.sub( sLine, nPos + 1 )
+            nPos = nPos + 1
+            recomplete()
+            redraw()
+
+        elseif sEvent == "paste" then
+            -- Pasted text
+            clear()
+            sLine = string.sub( sLine, 1, nPos ) .. param .. string.sub( sLine, nPos + 1 )
+            nPos = nPos + string.len( param )
+            recomplete()
+            redraw()
+
+        elseif sEvent == "key" then
+            if param == keys.enter then
+                -- Enter
+                if nCompletion then
+                    clear()
+                    uncomplete()
+                    redraw()
+                end
+                break
+                
+            elseif param == keys.left then
+                -- Left
+                if nPos > 0 then
+                    clear()
+                    nPos = nPos - 1
+                    recomplete()
+                    redraw()
+                end
+                
+            elseif param == keys.right then
+                -- Right                
+                if nPos < string.len(sLine) then
+                    -- Move right
+                    clear()
+                    nPos = nPos + 1
+                    recomplete()
+                    redraw()
+                else
+                    -- Accept autocomplete
+                    acceptCompletion()
+                end
+
+            elseif param == keys.up or param == keys.down then
+                -- Up or down
+                if nCompletion then
+                    -- Cycle completions
+                    clear()
+                    if param == keys.up then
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+        local nScroll = 0
+        if sx + nPos >= w then
+            nScroll = (sx + nPos) - w
+        end
+
+        local cx,cy = term.getCursorPos()
+        term.setCursorPos( sx, cy )
+        local sReplace = (_bClear and " ") or _tOptions.replaceChar
+        if sReplace then
+            term.write( string.sub( string.rep( sReplace, math.max( string.len(sLine) + nScroll, 0 ) ),  nScroll + 1, nScroll + w ) )
```
lurr: Can be shortened to:
```term.write( sReplace:rep(math.max( #sLine + nScroll, 0 ):sub( nScroll + 1, nScroll + w ) )```

File: src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua [hunk 2]
```diff
@@ -1,3 +1,290 @@
...
+        else
+            term.write( string.sub( sLine, nScroll + 1, nScroll + w ) )
```
lurr: Can be shortened to
```term.write( sLine:sub( nScroll + 1, nScroll + w) )```

File: src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua [hunk 3]
```diff
@@ -1,3 +1,290 @@
...
+
+    local function acceptCompletion()
+        if nCompletion then
+            -- Clear
+            clear()
+
+            -- Find the common prefix of all the other suggestions which start with the same letter as the current one
+            local sCompletion = tCompletions[ nCompletion ]
+            sLine = sLine .. sCompletion
+            nPos = string.len( sLine )
```
lurr: Can be shortened to
```nPos = #sLine```

File: src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua [hunk 4]
```diff
@@ -1,3 +1,290 @@
...
+                if nPos > 0 then
+                    clear()
+                    nPos = nPos - 1
+                    recomplete()
+                    redraw()
+                end
+                
+            elseif param == keys.right then
+                -- Right                
+                if nPos < string.len(sLine) then
```
lurr: ```if nPos < #sLine then```

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+                    nPos = 0
+                    recomplete()
+                    redraw()
+                end
+
+            elseif param == keys.delete then
+                -- Delete
+                if nPos < string.len(sLine) then
+                    clear()
+                    sLine = string.sub( sLine, 1, nPos ) .. string.sub( sLine, nPos + 2 )
```
lurr: ```sLine = sLine:sub(1, nPos) .. sLine:sub( nPos, 2 )```

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+                if nPos > 0 then
+                    clear()
+                    nPos = 0
+                    recomplete()
+                    redraw()
+                end
+
+            elseif param == keys.delete then
+                -- Delete
+                if nPos < string.len(sLine) then
```
lurr: ```if nPos < #sLine then```

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+                    uncomplete()
+                    redraw()
+
+                end
+
+            elseif param == keys.backspace then
+                -- Backspace
+                if nPos > 0 then
+                    clear()
+                    sLine = string.sub( sLine, 1, nPos - 1 ) .. string.sub( sLine, nPos + 1 )
```
lurr: ```sLine = sLine:sub(1, nPos - 1) .. sLine:sub( nPos + 1 )```

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+                        -- Down
+                        if nhistoryPos == #_tOptions.history then
+                            nhistoryPos = nil
+                        elseif nhistoryPos ~= nil then
+                            nhistoryPos = nhistoryPos + 1
+                        end                        
+                    end
+                    if nhistoryPos then
+                        sLine = _tOptions.history[nhistoryPos]
+                        nPos = string.len( sLine )
```
lurr: ```nPos = #sLine```

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+            clear()
+            sLine = string.sub( sLine, 1, nPos ) .. param .. string.sub( sLine, nPos + 1 )
+            nPos = nPos + 1
+            recomplete()
+            redraw()
+
+        elseif sEvent == "paste" then
+            -- Pasted text
+            clear()
+            sLine = string.sub( sLine, 1, nPos ) .. param .. string.sub( sLine, nPos + 1 )
```
lurr: ```sLine = sLine:sub( 1, nPos ) .. param .. sLine:sub( nPos + 1 )```

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+            recomplete()
+            redraw()
+        end
+    end
+    while true do
+        local sEvent, param = os.pullEvent()
+        if sEvent == "char" then
+            -- Typed key
+            clear()
+            sLine = string.sub( sLine, 1, nPos ) .. param .. string.sub( sLine, nPos + 1 )
```
lurr: ```sLine = sLine:sub( 1, nPos ) .. param .. sLine:sub( nPos + 1 )```

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+                    term.setTextColor( _tOptions.completeTextColor )
+                end
+                if not _tOptions.completeBGColor then
+                    term.setBackgroundColor( colors.gray )
+                else
+                    term.setBackgroundColor( _tOptions.completeBGColor )
+                end
+            end
+            if sReplace then
+                term.write( string.rep( sReplace, string.len( sCompletion ) ) )
```
lurr: ```term.write( sReplace:rep( #sCompletion ) )```

src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -1,3 +1,290 @@
...
+            sLine = string.sub( sLine, 1, nPos ) .. param .. string.sub( sLine, nPos + 1 )
+            nPos = nPos + 1
+            recomplete()
+            redraw()
+
+        elseif sEvent == "paste" then
+            -- Pasted text
+            clear()
+            sLine = string.sub( sLine, 1, nPos ) .. param .. string.sub( sLine, nPos + 1 )
+            nPos = nPos + string.len( param )
```
lurr: ```nPos = nPos + #param```

hugeblank:
@lurr Most everything you nitpicked was stuff that is in the original read function. I tried to keep everything as close as possible to the style that read() was written. While your proposed changes may be nice, the goal of this PR is not to redefine the style of ComputerCraft scripts.

apemanzilla:
@hugeblank I'd argue that keeping the exact code style in line with CC's original style isn't a good reason to avoid this kind of thing. Using colon notation and the `#` operator make it much shorter and easier to read IMO.

apemanzilla:
Also, I have a few suggestions:

- Help file. It's hard to find documentation for a lot of new features, adding a help file would make it much easier for others to use this.
- A filter function. Called whenever the user types input, returning either true or false to determine whether the input should be ignored or allowed. Something like `function(currentText, input) return ... end`, for example. Special care might need to be taken with the `paste` event.
- Configurable keys, pretty self explanatory.

BombBloke:
And I'd argue that "keeping things in line with CC's style" would be more of a reason to ditch the "arguments via a table" thing, if you really want to play that card.

hugeblank:
>.< conflicting opinions, If the general consensus is to go through with lurs changes then I will... unless by divine intervention (Dan), things go the other way.

@apemanzilla I'm going to need more explanation on what you mean by points 2 and 3... even though point 3 is supposedly self explanatory, 
Although I think I get what you're saying for the filter function, for example when pasting, the 'v' gets input, even though it shouldn't be? Maybe?

BombBloke:
Re the filter function, the idea is that the user presses a key, the char / paste gets sent to the function, and its return value is then treated as the actual input. For example, for pincode entry, you might write a filter function that returns "" for anything non-numeric.

apemanzilla:
@hugeblank BB pretty much got it right with regards to the filter function. A function that's called for all input, and can change or negate it accordingly.

By configurable keys, I meant adding options to change which keys (or key combos) did what - e.g. change the history forward/backward keys from up/down to page up/page down.

hugeblank:
Okay, I understand the idea of the filter function, and I think I could make it happen.
And, I can easily do the configurable keys, that won't be too difficult, just a tad bit tedious :P

hugeblank:
@apemanzilla @BombBloke Added both the filter function and the custom characters, killing and reviving slowWrite in the process. 

...and I forgot the freaking help file, let me implement that

apemanzilla:
I never liked slowWrite anyways.

hugeblank:
Unfortunately, it must live on...

apemanzilla:
Well, it's better than nothing, but given how complex the function is it would be a good idea to expand on it some - maybe change it to `textutils.prompt(table) - run "help prompt" for more info` and then add a separate help file that goes into specifics about options?

hugeblank:
My concern is that it isn't a big enough implementation to include an entire help article about it. I'll let this sit for a while and see what Bomb and others think.

apemanzilla:
I mean, there's a lot of options already - enough to easily fill a screen. I don't think it would be a good idea to try and fit it all into the textutils page.

hugeblank:
I'm saying to include everything it implements _at all_, not just adding it to the textutils help page.

hugeblank:
[this](https://github.com/hugeblank/ComputerCraft/blob/master/src/main/resources/assets/computercraft/lua/rom/help/prompt.txt) is what I would want the prompt.txt file to look something like if this it may be generally accepted.

thatcraniumguy:

Issue #466, opened 2017-09-26
Publish the API to a Maven repo


josephcsible:
We should publish the ComputerCraft API to a Maven repository to make it easier for other mods to use it.

thiakil:
For reference, to make an api jar, you can use something like this (just need to change the `include` line)

```
task apiJar(type: Jar) {
    classifier "api"
    dependsOn "sourceMainJava"
    include 'path/to/api/**'
    from sourceSets.main.java.srcDirs
    from sourceSets.main.output
}

artifacts {
    archives apiJar
}
```

SquidDev:
It's worth noting that https://cc.crzd.me has a sort-of-Maven repository running, which one is free to depend on. That's what OpenComputers and Plethora do:

```gradle
repositories {
  maven { url "https://cc.crzd.me/maven/" }
}
dependencies {
  compile "dan200.computercraft:ComputerCraft:1.81-build3"
}
```

Obviously having an official repository would be better, but this is a start.

thiakil:
Even just having the `api` artifact in the `build.gradle` would help; I have no issues hosting a copy of it for projects I'm a part of

PR #472, opened 2017-10-01
add basic api jar support

thiakil:
Adds an API jar artifact and a way to publish it to a maven directory (set environment variable `LOCAL_MAVEN`)

Solves the code side of #466


thatcraniumguy:

Issue #383, opened 2017-07-20
[Suggestion] turtle.getFacing();


MattB52:
Add a function to get the direction a turtle is facing such that it would return 0 for north, 1 for east, 2 for south, and 3 for west.

SquidDev:
It's worth noting you can already emulate this with [`turtle.inspect()` and placing torches against a wall](http://www.computercraft.info/forums2/index.php?/topic/24098-). Whilst this is a clever work around, it isn't quite the same as having functionality baked in.

I'm not sure using 0-3 as return values would be the best solution, especially as Lua tables start at 1.

KnightMiner:
I've seen this done as a peripheral before (using a compass), not sure how well it fits as its own feature.

I actually think 0-3 is a good idea as then you can do something like `(direction + x) % 4` and still have a valid direction. Might be weird for tables, though its a little hard to handle as typically 1 index is not normal

xAnavrins:
Turtle orientation pretty much fall in the same class than GPS coordinates, and with some cleverness, you can figure out your orientation using only two GPS.locate, I don't think it would be nessesary to add a function or a peripheral for it.

MineRobber9000:
I believe that this would be a good function to add. It wouldn't be hard to implement.

BombBloke:
It's less a question of "how it should be implemented" and more a question as to "whether it should be implemented".

Assuming the point of the mod is to get users to write code, this runs counter to that by removing the need for them to figure out their turtle's facing on their own.

There's also the "realism" aspect - using a peripheral for it is one thing (and as things stand, GPS already makes it trivial to get facing), but having turtles simply know their direction without an appropriate sensor is a different story.

Exerro:
I think turtles should be able to do most of the things a player can do to improve the potential for automation. A player can read a compass, why can't a turtle? I don't think it should be built into the turtle without a peripheral however.

@BombBloke if the point of the mod is to get users to write code, then why is GPS code built into computers by default such that it's trivial to get a turtle's location with a minor physical setup?

@SquidDev and @xAnavrins, while it is possible to do already, it's just awkward. Why not just add the peripheral? It's hardly bloat, it doesn't unbalance anything, it just makes it easier to control turtles, and might even encourage *more* programming because the work required for a seemingly simple action is proportional to the action itself.

BombBloke:
[...]

PR #473, opened 2017-10-04
Add turtle.getTurn()

JakobDev:
With this function you can get the Direction of a Turtle. The return value is a number, because it can be used in Loops. Here are the return value:
 1
4 2
 3

Note:
This is not compatible to #462. If it get merged, I will change case 42 to case 44.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
@@ -16,6 +16,7 @@ import dan200.computercraft.core.apis.ILuaAPI;
 import dan200.computercraft.shared.turtle.core.*;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
+import net.minecraft.util.EnumFacing;
 
 import javax.annotation.Nonnull;
 import java.util.HashMap;
@@ -107,6 +108,7 @@ public class TurtleAPI implements ILuaAPI
             "inspectUp",
             "inspectDown",
             "getItemDetail",
+            "getTurn",
         };
     }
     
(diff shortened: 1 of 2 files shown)


SquidDev:
It might be worth looking at the discussion in #383. I'm not going to re-iterate what was said there, merely state that I'm not a fan.

hugeblank:
What essentially was said in that PR was turtles having the ability to get their direction is giving them "magic" knowledge. And I have to agree with this, this doesn't seem like something that belongs in a turtle, as they don't have any sort of "compass" built in. 

Think of it this way: a player get's teleported randomly in a radius 10000 blocks away from his base, and has to find his way back. Assuming the player is unfamiliar with the land around him, how is he supposed to know which way to start walking? Players aren't given a direction unless they open the debug screen, which is comparable to a GPS. And even then, the player has to do some small movements to figure out which way is +X, and +Z (the rest from there can be assumed). 

One of the things brought up in #383 was turtles should be able to do what players can do, I.E. read a compass. And I absolutely agree with this. If a player can get their direction just by looking at a compass, then so should a turtle. 

So what I would recommend is turning a compass into a peripheral, and _THEN_ if it is equipped, provide this as a method (and maybe a `radiusFromSpawn` method too? ;D).

xAnavrins:
I don't support this, getting a turtle orientation pretty much falls in the same category as getting a turtle coordinate, which you can only get with GPS.
You can get turtle orientation using two GPS calls and some maths.

absolument:
Without GPS and since ComputerCraft 1.76 you can get Direction of a Turtle, by placing a block with facing block.state and then inspect it :
`pastebin get KE8cUP8x compas`
[https://pastebin.com/KE8cUP8x](url)

ex:
Put a chest in the first slot of Turtle inventory, then run this command
`compas 1`

SuperDyl:
Two parts:

1. Turtles could have knowledge of world direction because blocks in Minecraft all have world direction knowledge. The cracks of breaking blocks and the textures of the tops of blocks can be used to find cardinal directions.

2. A `turtle.getTurn()` function ought to use the same system as math for angle measure. That means either 1, 2, 3, and 4 (for the four cardinal directions) or 0, 90, 180, and 270 (for angle measures) would be returned, but most to follow conventions, rotations in math are done counterclockwise. That means the output should be:

  1
2  4
  3

thatcraniumguy:

PR #474, opened 2017-10-04
Improved autocompletion for set.lua

Wilma456:
If a setting is boolean, set.lua will now complete true or false.


KingofGamesYami:
What if a setting is false or non-boolean?  Eg. something like "maximum number of ___" where false would indicate unlimited.

Wilma456:
If a setting is not a boolean, it will not be autocompletet to true or false.

thatcraniumguy:

PR #479, opened 2017-10-12
Only send terminal state to interacting players

SquidDev:
This splits the computer state (blinking, label, etc...) and terminal state into two separate packets. When a computer changes, the computer state is sent to all players and the terminal state is sent to players who are currently using the computer. This reduces network usage by a substantial amount. 

CCTweaks has employed a similar optimisation for a while now, and not seen any adverse effects.


thatcraniumguy:

PR #480, opened 2017-10-13
Add getTextScale() to Monitor

JakobDev:
We had for almost all set Function also a get Function, but not for setTextScale(). This PR add one.


SquidDev:
This appears to not work for monitors made of multiple blocks, as it doesn't use the "main" terminal. I've put together a patch here: https://github.com/SquidDev-CC/CC-Tweaked/commit/aa447ec1016126c9bf3bd921d4f4edf41ab1b5a5 which appears to fix the issue.

thatcraniumguy:

PR #481, opened 2017-10-19
Make all Lua APIs comatible with require()

Wilma456:
With this PR, all APIs will be compatible with os.loadAPI() and require().

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/colors.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/colors.lua
@@ -1,22 +1,29 @@
 -- Colors
-white = 1
-orange = 2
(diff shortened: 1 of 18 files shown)


SquidDev:
I'd say that _generally_ this is a good idea,  but there's a couple of issues:

 - Some libraries have to be global and so won't work when loaded with `require`. Off the top of my head, these are `rednet`, `peripheral`, `term`, `turtle`.
 - Instead of having `colors`, `redstone`, etc... as the local, use `M` (or `_M`). This is pretty [standard amongst Lua modules](https://github.com/Olivine-Labs/busted/blob/master/busted/done.lua#L1) and means you don't end up with ugly variable names like `windowapi`.

PR #482, opened 2017-10-25
Add simple JEI integration

SquidDev:
I know that mod integration is something that ComputerCraft normally steers clear of, but JEI is sufficiently ubiquitous that this is probably worth it.

JEI compares items simply by item name and damage value, ignoring NBT. This means many recipes do not function correctly. For instance, viewing the recipe of a Melee Turtle will display all upgrade recipes. This PR fixes this:

 - Pocket computers and turtles are distinguished by upgrades and computer family.
 - Disks are distinguished by colour.
 - Hide treasure disks from the item list

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,28 +41,16 @@ minecraft {
     // makeObfSourceJar = false // an Srg named sources jar is made by default. uncomment this to disable.
 }
 
+repositories {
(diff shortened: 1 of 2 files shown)


thatcraniumguy:

PR #490, opened 2017-11-15
Propagate errors arising from API loading

zardyh:
This (admittedly backwards-incompatible commit, but read on) changes the behaviour of `os.loadAPI` to propagate errors that happen during the loading of APIs. The old behaviour would lead to objectively worse error messages, by delaying the reporting of errors. As an example, consider the following code.

```
os.loadAPI("foo") -- foo doesn't exist
foo.bar()
```
In this case, the code will end up printing something like
```
File not found
file:2: attempt to index global 'foo' (a nil value)
```

An inexperienced programmer could easily be stumped by the fact their code is failing in the wrong place - what would be expected is that the error raised by `loadfile` would be propagated upwards from `os.loadAPI`, thus halting in the first line with an useful error message.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -644,14 +644,12 @@ function os.loadAPI( _sPath )
     if fnAPI then
         local ok, err = pcall( fnAPI )
         if not ok then
-            printError( err )
             tAPIsLoading[sName] = nil
-            return false
+            return error( "Failed to load API " .. sName .. " due to " .. err, 1 )
         end
     else
-        printError( err )
         tAPIsLoading[sName] = nil
-        return false
+        return error( "Failed to load API " .. sName .. " due to " .. err, 1 )
     end
     
     local tAPI = {}


lupus590:
I wouldn't say that this is backwards incompatible, in my mind that means that something which could be done can no longer be done or has to be done another way.

Although, I suppose it would break programs which attempt to load and then test the table to see if the load was successful.

zardyh:
Indeed, it can break programs that rely in the (in my opinion) improper behaviour. Of course, a fix is easy:

```lua
-- old
os.loadAPI("foo")
if not foo then
  ...
end
```

becomes

```lua
local ok, err = pcall(os.loadAPI, "foo")
if not ok then
  ...
end
```

lupus590:
perhaps you should change os.loadAPI so that instead of propagating the error it 'catches' and returns it,  users then don't need to use `pcall`.

zardyh:
That'd have the same problem as before: Errors would be silently ignored, leading to confusing interactions.

lupus590:
but existing programs would not break

zardyh:
The point _is_ to break programs. To break **broken programs**, that is, instead of allowing them to crash elsewhere with confusing error messages.

I have _personally_ (and @SquidDev can attest to this) spent over 2 hours helping someone on Discord to fix a problem that boiled down to "the API wasn't found, and the program only stopped a lot later". Debugging this kind of issue is not fun - so why even let it happen?

apemanzilla:
IMO "breaking" `os.loadAPI` with this change is fine given that people should really be moving to `require` anyways

apemanzilla:
Also, at the very least, if this PR doesn't get merged, because it's introducing breaking changes, the error message should also be returned, so that `assert(os.loadAPI("thing"))` can be used to "safely" error out.

dan200:
It's not really worth "fixing" os.loadAPI when as ape says, we have "require" now. It's better to keep the old behaviour for compatibility.

PR #491, opened 2017-11-18
Provide an API for registering custom APIs

### LINKED ISSUE #346: Improved methods for mounting and such
As far as I can tell, there isn't a way to mount and automatically run a lua file (from the Java side of things).  Specifically, I would like to run a short program adding `mymount/programs` to the shell path, and adding auto-completion support for the program(s) that are mounted.

I would also like a method for determining the absolute directory of a computer (ei C:/somebody/...) as I need that for another API.  I've been working around this without much trouble, but it should be trivial to implement.

It would also be nice if it was possible to add Java-side APIs to computers (without peripherals / upgrades / etc.).  I've been using CCTweaks to get around this, perhaps @SquidDev could provide some idea of what would need to be done for this.  I would honestly just PR the extra API, except for the "no additional dependencies" rule, and this is one _huge_ frigging dependency.

These are just a couple things I've run into during the development of [CCGit](https://github.com/KingofGamesYami/CCGit-Rewrite)

SquidDev:
`ILuaAPI` has been moved to `dan200.computercraft.api.lua`. One creates a new API by registering an instance of `ILuaAPIFactory`. This takes an instance of `IComputerSystem` and returns such an API.

`IComputerSystem` is an extension of `IComputerAccess`, with methods to access additional information about the the computer, such as its label and filesystem.

This closes #346, which also explains some of the rational behind this functionality.


src/main/java/dan200/computercraft/ComputerCraft.java
```diff
@@ -644,6 +646,14 @@ public static void registerMediaProvider( IMediaProvider provider )
         }
     }
 
+    public static void registerAPIFactory( ILuaAPIFactory provider )
```
Lignum: Perhaps this should return a `boolean` specifying whether the API was actually registered?

SquidDev: I'm only really doing it this way because that's what all the other methods do. The only way it could thrown an exception is if the reflection fails, which shouldn't ever happen anyway.

Lignum: Oh, I was referring to `if( provider != null && !apiFactories.contains( provider ) )` failing.

src/main/java/dan200/computercraft/ComputerCraft.java
```diff
@@ -244,6 +245,7 @@
     private static List<IMediaProvider> mediaProviders = new ArrayList<>();
     private static List<ITurtlePermissionProvider> permissionProviders = new ArrayList<>();
     private static final Map<String, IPocketUpgrade> pocketUpgrades = new HashMap<>();
+    private static final List<ILuaAPIFactory> apiFactories = new ArrayList<>();
```
apemanzilla: By using a `Set` you can avoid explicitly checking if it's been registered later

SquidDev: There's an argument that this allows for some level of priority ordering. I'm not entirely sure whether such an ordering is needed though - *ideally* mods shouldn't be overwriting each other's APIs, but it's something worth discussing.

apemanzilla: I don't think there's any particular reason order is important, but if it is, `LinkedHashSet` maintains order

PR #492, opened 2017-11-21
 Improve vertex transformation system

SquidDev:
This migrates `TurtleMultiModel`'s current vertex transformation system into something more powerful and "correct". Namely, it has the following improvements:

 - Handles all position formats (float, byte, etc...)
 - Correctly translates normals of quads
 - Reorders faces if the winding order is reversed

This also improves how particular turtles are rendered, as everyone needs some joy in their life.


thatcraniumguy:

PR #495, opened 2017-11-22
Show Filenames in Multishell Title

JakobDev:
With this PR, you can directly see, which File is opened is which tab, if you have multiple files open.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -21,6 +21,10 @@ if not fs.exists( sPath ) and not string.find( sPath, "%." ) then
     end
 end
 
+if multishell and settings.get( "edit.multishell_title" ) == true then
+    multishell.setTitle( multishell.getCurrent(), "edit (" .. fs.getName( sPath ) .. ")" )
+end
+
 local x,y = 1,1
 local w,h = term.getSize()
 local scrollX, scrollY = 0,0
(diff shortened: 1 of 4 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -21,6 +21,10 @@ if not fs.exists( sPath ) and not string.find( sPath, "%." ) then
     end
 end
 
+if multishell and settings.get( "edit.multishell_title" ) == true then
```
SquidDev: There isn't any need for `== true`. I know you prefer it, but let's keep it consistent with the rest of the code :).

SquidDev:
I wonder if it'd be worth limiting file names to 20 characters or something? Maybe something like:

```lua
local name = fs.getName( path )
if #name > 20 then name = name:sub(1, 17) .. "..." end
if multishell and settings.get( "paint.multishell_title" ) == true then
  multishell.setTitle( multishell.getCurrent(), "paint (" .. name .. ")" )
end
```

This way really long file names won't fill the tab bar, but most file names will still display fine.

lupus590:
Max length as a settings option?

Wilma456:
@lupus590 Whiyh is a good name for this settings? Maybe multishell.max_filename? What do you think?

lupus590:
`multishell.tabTitleMaxLength` (or is underscore the established convention?)

thatcraniumguy:

PR #497, opened 2017-12-01
Add map-like rendering for pocket computers

SquidDev:
This has been a suggested feature for as long as pocket computers have been a thing, so I thought it'd be worth seeing what other people think.

All in all, it works pretty well though there are a couple of issues:
 - Text is near illegible when in the off-hand. I don't think this is too big an issue as I see this being used for more graphical readouts.
 - Pocket upgrades have the ability to interact with right clicks. This may be less intuitive when rendering items like this.
 - The aspect ratio of the item's screen are different to the actual terminal's ratio. This means there is a wide margin to the left/right of the terminal. Rendering the background colour to the edge is even uglier, so I think we'd have to leave as is.

## Screenshots
Shows this in action

Rendering in the offhand

The image was mean't to say "Hai!" but I dropped a letter. Sorry.

### DIFF
--- a/src/main/java/dan200/computercraft/client/proxy/ComputerCraftProxyClient.java
+++ b/src/main/java/dan200/computercraft/client/proxy/ComputerCraftProxyClient.java
@@ -8,6 +8,7 @@ package dan200.computercraft.client.proxy;
 
 import dan200.computercraft.ComputerCraft;
 import dan200.computercraft.client.gui.*;
+import dan200.computercraft.client.render.ItemPocketRenderer;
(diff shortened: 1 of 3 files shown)


Restioson:
Great! How does one get it into map-like form (keybind)?

SquidDev:
> Just one question - how does one get it into map-like form (keybind)?

It's like that by default. It would be possible to add a config option/keybind/per-computer setting but I'm not sure how user-friendly that is.

thatcraniumguy:

PR #498, opened 2017-12-04
Add -a Option to list.lua

JakobDev:
This PR simple adds a -a Option to list.lua which allows you to see hidden Files, even if the list.show_hidden setting is false.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/list.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/list.lua
@@ -3,7 +3,14 @@ local tArgs = { ... }
 
 -- Get all the files in the directory
 local sDir = shell.dir()
-if tArgs[1] ~= nil then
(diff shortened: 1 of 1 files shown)


apemanzilla:
What happens if you want to list files in the `-a` directory? You'd have to do `ls -a -a` for it to work, which would also show hidden files when you might not want them, or `cd -a` and then `ls`, which adds an extra command.

SquidDev:
@apemanzilla I believe you can just do `ls ./-a` then: that's what one does on Linux at least.

JasonTheKitten:
I waited a day to post this, but I really wanted to post it...

*SELF-PROMOTION *
_[GetOpts](https://pastebin.com/PSFnwfPp)_. An entire 205 line API when you really only needed 5 lines!

It would probably cause more bugs than it is worth, but least it would be reusable.

PR #506, opened 2018-01-14
Fix advanced monitor recipe.

Wojbie:
Fix recipe to create 4 monitors.

It was set to create only 1 monitor in 1.12 port - this fixes it.


SquidDev:
Sorry! I think normal monitors only created one, and I just copied the recipe from that.

Issue #410, opened 2017-08-09
Modem Transmission returns java error


hugeblank:
So I'm not sure if this issue is known or not, but I'm playing with CC on a server in 1.12 and I keep getting this error from time to time. It's inconsistent, and has caused many breakages in computers that would typically act normally. 
Here is a screenshot of the error that gets thrown. Apologies that it's on a pocket computer, it's the only computer in the area that got this error... [error.png](https://user-images.githubusercontent.com/16393543/29113408-c9f644ae-7ca5-11e7-9134-ff42777d3e9b.png)

MineRobber9000:
![Proof that this isn't just exclusive to Pocket Computers](https://user-images.githubusercontent.com/12676241/29113717-e2c768fe-7cbf-11e7-896b-66498776f348.png)

SquidDev:
@hugeblank Can you rebuild with #401 merged, enable `logPeripheralErrors` in the config and post a couple of stack traces from the logs when the error occurs. I'm rather surprised to see this happening on modems.

---

In the case of in-world modems, I've only ever noticed this when the peripheral is broken whilst spamming packets. `transmit` calls [`ModemPeripheral.getPosition()`](https://github.com/dan200/ComputerCraft/blob/579f7443a8c47d71efd71da41744239b5019f320/src/main/java/dan200/computercraft/shared/peripheral/modem/ModemPeripheral.java#L242), which calls [`TileWirelessModem.getDirection()`](https://github.com/dan200/ComputerCraft/blob/579f7443a8c47d71efd71da41744239b5019f320/src/main/java/dan200/computercraft/shared/peripheral/modem/TileWirelessModem.java#L47) which [queries the block state](https://github.com/dan200/ComputerCraft/blob/579f7443a8c47d71efd71da41744239b5019f320/src/main/java/dan200/computercraft/shared/peripheral/modem/TileWirelessModem.java#L73). This is problematic at the best of times, as it occurs on the computer thread - it would probably be better to store the direction of block in the TE (I don't know if the hooking into the `validate` method would work).

It's worth noting that there are several other thread unsafe accesses, which can cause similar problems: see SquidDev-CC/CCTweaks#106 and SquidDev-CC/CCTweaks#113. I believe dan200/ComputerCraft#105 would also be fixed by caching various properties.

hugeblank:
[...]

PR #507, opened 2018-01-16
[WIP] World thread-safety

SquidDev:
This is an attempt at ensuring peripherals/other Lua calls do not access the world from the Lua thread. I'm going to be honest, none of the fixes are especially elegant. I'm expecting several of the fixes to require some level of polling, which is a pain. Currently most peripherals implement `ITickable` so it's not the end of the world,  but it'd be nice to reduce the number which do in the future.

Current progress of the PR:

 - [x] Cache direction of modems. (see #410).
 - [x] Cache turtle family. (see SquidDev-CC/CCTweaks#113).
 - [ ] Cache monitor's terminal and peripheral type.
 - Probably lots of other things I haven't come across yet.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/peripheral/common/TilePeripheralBase.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/common/TilePeripheralBase.java
@@ -24,7 +24,7 @@ public abstract class TilePeripheralBase extends TileGeneric
 {
     // Statics
 
-    private EnumFacing m_dir;
+    protected EnumFacing m_dir;
     private int m_anim;
     private boolean m_changed;
 
@@ -94,6 +94,11 @@ public abstract class TilePeripheralBase extends TileGeneric
 
     @Override
     public EnumFacing getDirection()
+    {
+        return m_dir;
+    }
+
+    public EnumFacing getCachedDirection()
     {
         return m_dir;
     }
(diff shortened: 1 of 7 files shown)


lupus590:
>it's not the end of the world

If it causes a save corruption then, arguably, it is the end of that world. 🤣 

I'll show myself off out...

SquidDev:
As you may be able to tell, the monitor changes are slightly larger than anticipated. As the potential for breakage is larger, I'd recommend some serious testing before merging. We're currently using this on SwitchCraft, which caught a lot of the issues before they ended up in the PR.

If anyone's interested in helping test, please [download from the CI server](http://cc.crzd.me/maven/dan200/computercraft/ComputerCraft-SquidDev-CC-ComputerCraft-feature-world-thread-safety/?sort=time&order=desc).

SquidDev:
Whilst the last commits aren't strictly world/computer threading issues, but they fix various aspects about monitors and it was easier to bundle them here instead. I'm happy to revert them if inconvenient.

There's some [other fixes](https://github.com/SquidDev-CC/CC-Tweaked/pull/15) I'd like to add at some point, but I'd like to confirm they work and are needed first.

thatcraniumguy:

SquidDev:
For what it's worth, I've been running a fork with almost all of my PRs merged on a relatively popular server and not seen any issues (it's actually _more_ stable thanks to this pr).

That being said, there's not really much point bumping PRs. Dan merges stuff when he has time. While it's a little infuriating seeing stuff just sit there, I don't think it's fair to keep pestering Dan either.

thatcraniumguy:
I don't remember adding this comment... I think my Jenkins server pulled them all and auto-commented. My bad.

SquidDev:
Ahhh, that explains a lot. I was rather confused when I woke up this morning and there were ~30 PRs with the same comment!

PR #508, opened 2018-01-19
Fix turtle speaker upgrade's missing texture

SquidDev:
The sprite was not registered into the atlas, meaning it rendered the missing texture instead. I'm going to be honest, I'm not sure why it was working in 1.11.2 - the code hasn't changed since then.


thatcraniumguy:

PR #509, opened 2018-01-19
Improved autocompletion for edit.lua

JakobDev:
At the Moment, edit.lua will only complete global functions. But must Programs have local functions. With this PR, edit.lua will search the code of your file and add all functions to autocompletion.


apemanzilla:
This won't work for anything involving scope or tables, and the pattern will also match things like newlines which could lead to all sorts of strange things, especially with anonymous functions.

thatcraniumguy:

PR #290, opened 2017-05-29
Add build instructions to Readme

JakobDev:
#286

PR #510, opened 2018-01-25
Create setup.md

JasonTheKitten:
This is so that people know how to set it up

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -1,0 +1,55 @@
+you will need to download the following Programs:
+IntelliJ
+Java 8 (You may need to uninstall Java 9)
+Java 8 SDK - You should choose the version that matches your Java
+git bash
+
+Run each of those, and wait for them to install.
+Next, open git bash and run the following:
+git clone you will need to download the following files:
+IntelliJ
+Java 8 (You may need to uninstall Java 9)
+Java 8 SDK - You should choose the version that matches your Java
+git bash
+
+Run each of those, and wait for them to install.
+Next, open git bash and run the following:
+git clone https://github.com/dan200/ComputerCraft.git
+
+When it is done, leave it open, we will use it again.
+
+Open IntelliJ and choose 'import project'. Navigate to the computercraft folder and choose 'build.gradle'
+I think it might ask you where your SDK is, if it is 64x it will be in Program Files/Java, labeled jdk1.80_161 (or whatever patch you have)
+It will start syncing, when the bottom window says it is done syncing, close IntelliJ.
+Now you must type ./setup.bat into git bash
+When that finishes, open IntelliJ again.
+Goto file>project settings>project. Change the project language level from 9 to 8.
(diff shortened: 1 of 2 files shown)


setup.md
```diff
@@ -0,0 +1,55 @@
+you will need to download the following Programs:
+IntelliJ
+Java 8 (You may need to uninstall Java 9)
+Java 8 SDK - You should choose the version that matches your Java
+git bash
+
+Run each of those, and wait for them to install.
+Next, open git bash and run the following:
+git clone you will need to download the following files:
```
lupus590: i think you have a copy past error or something here

SquidDev:
I'm going to be honest, I do feel this is overly complicated for what equates to:

 - Clone the repo
 - Run `./gradlew setupDecompWorkspace` in the directory
 - Import `build.gradle` into your editor of choice.

Have a look at some of Dan's comments on #290 as well - one doesn't need to spoon feed it to the user.

setup.md
JasonTheKitten: Yea I did have a copy/paste error. A few, in fact

SquidDev:
@JasonTheKitten Have you ever used Gradle + a Java IDE before though? Once you've done it once it's pretty much the same for all projects: this repo doesn't need to be Java 101.

I think a better solution would be to have some variation of the above bullets and then link to http://mcforge.readthedocs.io/en/latest/gettingstarted/ for more information.

With respect to CC, we should probably mention `deploy.sh`, at least until we remove the need for it.

JasonTheKitten:
I use windows cmd not linux bash so I can't use sh files

JasonTheKitten:
> once you've used Gradle once, it's the same elsewhere.

What does that mean? If you mean it makes new gradle files, then can't we just ues copy/paste instead of the git command?

SquidDev:
> > once you've used Gradle once, it's the same elsewhere.
> 
> What does that mean? 

Once you've set up Gradle in your editor once, it's pretty much the same procedure in any other project which uses Gradle.

lupus590:
@JasonTheKitten are you aware that you still have this pull request open and you have made a commit against it which seems irrelevant to the pull request?

JasonTheKitten:
That was an accident

lupus590:
https://sethrobertson.github.io/GitFixUm/fixup.html

You will need to force the push, after following the steps in the above link, do this with `git push -f` (assuming you have your default remote set correctly)

apemanzilla:
Force pushing is a bad idea. If you've already pushed erroneous commits (which you have) you should just revert the changes instead.

KnightMiner:
Reverting leaves a revert commit which makes the PR much harder to review. Force push is perfectly valid for PRs

apemanzilla:
The files changed tab will not show the difference because the changes are reverted, and the extra commits can be dropped when merging.

thatcraniumguy:

PR #514, opened 2018-02-01
HTTP.request function upgrade!

JasonTheKitten:
Added important HTTP features used by many real-world applications!
Compatibility is retained with 99.9% of programs!
Interact with servers that require non-get/post methods!

Updates include getting the text that goes with the response code from the server instead of using a lookup table that may not match the server's, the ability to use methods other than post and get, and also the ability to enable or disable automatic page redirects.

I thought about adding a way to get the encryption certificate, but decided against it.
Also, due to limitations of httpurlconnections, there is a whitelist of usable methods. This can probably be changed by writing a script to communicate directly with a socket, but under the same name and with the same methods (so we don't have to change all of our other code), but I decided to let whoever else do that, or maybe I'll do that later, but right now I'm not too worried about it.

**EDIT:** Not sure why travis-ci is erroring, I looked at the log, it was all stuff that was already there. Plus the java code, which is where it is erroring, works on the devices in which I try them on


JasonTheKitten:
Not sure why travis-ci is erroring, I looked at the log, it was all stuff that was already there. Plus this works on the devices in which I try them on

PR #515, opened 2018-02-01
HTTP.request function upgrade!

JasonTheKitten:
**I think that this time it will pass the travis-ci**
**Edit: Yay, it did. Time before this it did not**

Added important HTTP features used by many real-world applications!
Compatibility is retained with 99.9% of programs!
Interact with servers that require non-get/post methods!

Updates include getting the text that goes with the response code from the server instead of using a lookup table that may not match the server's, the ability to use methods other than post and get, and also the ability to enable or disable automatic page redirects.

I thought about adding a way to get the encryption certificate, but decided against it.
Also, due to limitations of httpurlconnections, there is a whitelist of usable methods. This can probably be changed by writing a script to communicate directly with a socket, but under the same name and with the same methods (so we don't have to change all of our other code), but I decided to let whoever else do that, or maybe I'll do that later, but right now I'm not too worried about it.


lupus590:
>REPLY TO SquidDev's COMMENT...

Wrong PR?

src/main/java/dan200/computercraft/core/apis/HTTPAPI.java
```diff
@@ -18,56 +18,45 @@
...
+                    h.whenFinished(m_apiEnvironment);
                     it.remove();
                 }
             }
         }
     }
-    
+
     @Override
-    public void shutdown( )
```
SquidDev: I don't like to nitpick, but it's probably best to keep whitespace changes to a minimum. I know indent styles are rather a personal matter, but it's best to be consistent :). It also makes it easier to review the PR and see what functionality has changed, rather than what's just cosmetic.

src/main/java/dan200/computercraft/core/apis/HTTPAPI.java
```diff
@@ -76,90 +65,81 @@ public void shutdown( )
...
                 boolean binary = false;
-                if( args.length >= 4 )
-                {
-                    binary = args[ 3 ] != null && !args[ 3 ].equals( Boolean.FALSE );
+                if (args.length >= 4) {
+                    binary = args[3] != null && !args[3].equals(Boolean.FALSE);
                 }
 
+                //Get mode
+                String DMODE = "GET";
```
SquidDev: Also minor stylistic choice, but probably shouldn't be UPPER_CASE - that's normally reserved for constants.

src/main/java/dan200/computercraft/core/apis/HTTPRequest.java
```diff
@@ -48,37 +48,47 @@ public static URL checkURL( String urlString ) throws HTTPRequestException
         return url;
     }
 
-    public HTTPRequest( String url, final String postText, final Map<String, String> headers, boolean binary ) throws HTTPRequestException
```
SquidDev: This whole file shouldn't exist - I'm pretty sure it's me mucking up a rebase at some point. It's probably best to not edit it at all and someone can delete it in the future (or just delete it now).

src/main/java/dan200/computercraft/core/apis/http/HTTPRequest.java
```diff
@@ -278,6 +286,11 @@ private static ILuaObject wrapStream( final ILuaObject reader, final int respons
                         // getResponseHeaders
                         return new Object[] { responseHeaders };
                     }
+                    case 2:
+                    {
+                        // getResponseCodeText
+                        return new Object[] { responseCodeText };
```
SquidDev: Instead of adding a separate method, I wonder if we could just return both in `getResponseCode`? Might be a slightly cleaner API.

SquidDev:
Aside from my above comments it looks good. 

As so far as additional methods go, I've been looking into using Netty's HTTP client instead, as it imposes less restrictions (and also drops the need to have separate threads for each request). One'd probably want to wait until #395 has been reviewed, but it might also be worth looking into.

src/main/java/dan200/computercraft/core/apis/HTTPRequest.java
JasonTheKitten: Thing doesn't seem to work without it though...

src/main/java/dan200/computercraft/core/apis/HTTPAPI.java
JasonTheKitten: BTW, it really annoys me that the {'s seem to be on a separate line than the code controlling  the block, but I try to follow that formatting anyways. Don't know why.

src/main/java/dan200/computercraft/core/apis/HTTPRequest.java
SquidDev: Strange. From what I can tell, all references are pointing towards `apis.http.HTTPRequest` instead of `apis.HTTPRequest`. What error do you get?

src/main/java/dan200/computercraft/core/apis/HTTPAPI.java
SquidDev: Yeah, there's lots of things I'm not a fan of as far as CC's formatting goes, but you have to roll with it :).

JasonTheKitten:
Their never seems to be a reply option to lupus's comments...
> Wrong PR?
Branch copy error, I'm worried that deleting it might delete my other branch?

src/main/java/dan200/computercraft/core/apis/http/HTTPRequest.java
JasonTheKitten: Maybe, can I get Lupus's opinion first though?

SquidDev:
> Branch copy error, I'm worried that deleting it might delete my other branch?

It shouldn't - if you're rebasing, it'll just delete those commits on this particular branch.

JasonTheKitten:
Ok, got rid of that :)

src/main/java/dan200/computercraft/core/apis/HTTPRequest.java
JasonTheKitten: I might be wrong, I'll have to check

SquidDev:
Someone added a request similar to this one on the CC:T issue tracker, so I thought I'd have another look at this and raise a few suggestions. One worry I have is that `http.request` (and other `http` methods) are ending up like `read`, with far too many arguments to keep track of.

One possible solution would be to determine if the first argument is a table (falling back to the current system if not) and extract our options from there. This could use a scheme similar to [`luasocket.http.request`](http://w3.impa.br/~diego/software/luasocket/http.html):

```lua
http.request {
  url = string,
  [method = string,]
  [headers = header-table,]
  [body = string],
  [redirect = boolean,]
  [binary = boolean,]
}
```

This means we can add as many arguments and options as we like (well, within reason) whilst still keeping calls readable and maintaining backwards compatibility. Obviously this does make argument parsing more complex (though one could create some helpers for table parsing on the Java side), but I think it's worth it.

thatcraniumguy:

PR #517, opened 2018-02-04
Add ore dictionary support to all recipes

SquidDev:
This allows people to use alternative items, such as Quark chests or dyed glass panes, in recipes instead of the "standard" one. The former is especially useful as Quark makes it much harder to acquire a "standard" chest.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/media/recipes/DiskRecipe.java
+++ b/src/main/java/dan200/computercraft/shared/media/recipes/DiskRecipe.java
@@ -10,19 +10,23 @@ import dan200.computercraft.shared.media.items.ItemDiskLegacy;
 import dan200.computercraft.shared.util.Colour;
 import dan200.computercraft.shared.util.ColourTracker;
 import dan200.computercraft.shared.util.ColourUtils;
-import net.minecraft.init.Items;
 import net.minecraft.inventory.InventoryCrafting;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.IRecipe;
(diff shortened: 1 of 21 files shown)


thatcraniumguy:

src/main/resources/assets/computercraft/recipes/skull_cloudy.json
```diff
@@ -1,7 +1,7 @@
 {
     "type": "minecraft:crafting_shapeless",
     "ingredients": [
-        { "item": "minecraft:skull", "data": 1 },
+        { "item": "minecraft:skull", "data": 32767 },
```
KnightMiner: The wildcard metadata is deprecated by Mojang, it will no longer work in 1.13 and does not show properly in the recipe book in 1.12. Mojang replaced all uses of it with a list of ingredients. Since its reused, male that list of skulls a constant

SquidDev: Ahhh, thanks. I've used it without problem with JEI, but forgot to test the recipe book.

Out of interest, what's the recommended way of dealing with items like wool, where there's no corresponding ore-dict entry? Obviously in 1.13 we can use item tags, but it also seems a little sub-optimal to list all 16 variants.

KnightMiner: Yeah, all recipes that use all 16 variants of wool in 1.12 listed all 16 variants. You could also just add your own oredict entry to the vanilla items to supplement until tags are added in 1.13, as I'm sure other mods add their own wools too for the oredict

PR #518, opened 2018-02-05
Track which player "owns" a turtle

SquidDev:
When a player places a turtle, they are marked as its owner. Any actions they perform (such as breaking blocks, moving, etc...) are performed using this player's game profile.

This allows turtles to work correctly with various permissions/claim mods. Previously you would have to whitelist all turtles in order for them to function within a claimed area.

See SquidDev-CC/CC-Tweaked#8 for the original issue & explanation.

### DIFF
--- a/src/main/java/dan200/computercraft/api/turtle/ITurtleAccess.java
+++ b/src/main/java/dan200/computercraft/api/turtle/ITurtleAccess.java
@@ -6,6 +6,7 @@
 
 package dan200.computercraft.api.turtle;
 
+import com.mojang.authlib.GameProfile;
 import dan200.computercraft.api.lua.ILuaContext;
 import dan200.computercraft.api.lua.LuaException;
 import dan200.computercraft.api.peripheral.IPeripheral;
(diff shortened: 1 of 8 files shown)


lupus590:
I assume turtles which are placed by turtles inherit the placing turtle's ownership?

What happens if I place a turtle using another mod? I would assume the mod's fake player get's ownership. (I may be wrong, but can a mod place blocks without a fake player? What about that?)

SquidDev:
> I assume turtles which are placed by turtles inherit the placing turtle's ownership?

Yep. You can have turtles all the way down!

> What happens if I place a turtle using another mod?

It depends. If other mods are doing things "correctly" and tracking a tile's owner, then it'll inherit the owner from there too. Otherwise it'll either use that mod's fake user\* or CC's fallback fake user.

\* By "fake user" I mean a fabricated `GameProfile`. We're still using a `FakePlayer` (or `TurtlePlayer` to be precise), it's just we're constructing it with the owner's `GameProfile`.

Selim042:
Is there a way to check who owns a given turtle via Lua?

SquidDev:
> Is there a way to check who owns a given turtle via Lua?

No, though that _could_ be exposed in a later PR or through an upgrade.

Selim042:
Alright.  I assumed that is what you were going to say, just curious.

thatcraniumguy:

PR #520, opened 2018-02-12
Prevent turtles moving beyond the world border

SquidDev:
As tiles outside the world border are not ticked, turtles are rendered entirely useless. Furthermore, the turtle animation will never progress resulting in visual glitches.

In order to avoid this, we ensure the target position is within the world border when moving to it.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/turtle/core/TurtleMoveCommand.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/core/TurtleMoveCommand.java
@@ -14,9 +14,9 @@ import dan200.computercraft.api.turtle.TurtleCommandResult;
 import dan200.computercraft.shared.util.WorldUtil;
 import net.minecraft.block.Block;
 import net.minecraft.entity.Entity;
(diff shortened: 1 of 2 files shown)


src/main/java/dan200/computercraft/shared/turtle/core/TurtleMoveCommand.java
```diff
@@ -14,9 +14,9 @@
 import dan200.computercraft.shared.util.WorldUtil;
 import net.minecraft.block.Block;
 import net.minecraft.entity.Entity;
+import net.minecraft.util.EnumFacing;
```
lupus590: Why was this line moved? I assume the move has no effect and just makes the commit a bit messy.

SquidDev: Because my IDE likes to put things in alphabetical order. I'm going to be honest, I can't really muster the energy to fix it.

thatcraniumguy:

Issue #276, opened 2017-05-21
[Suggestion] Turtles get buffs when in range of (vanilla) beacons


lupus590:
Suggested by @SquidDev  [here](https://github.com/dan200/ComputerCraft/issues/274#issuecomment-302953395) on #274 .

Basically, allow turtles to get benefits from beacon effects (and maybe potions?).

Haste (or whatever the speed boost one is called) could either make turtles move faster or reduce the fuel cost (which would make fuel values floats which might cause issues. Maybe this isn't the best example).

SquidDev:
Copying an extract from my follow up comment:

> I'm more interested in exposing the functionality to allow those sorts of features than implementing the features themselves. This way you give other peripheral developers the same power and flexibility as the mod itself.

> Detecting beacons and responding to them requires some rather ugly hacks and I'm not entirely sure how efficient it would be - something which would have to be investigated before jumping in.

There are two solutions I can think of:
 - Iterating over the `World.tickableTileEntities` set and determining whether they are within range and are beacons.
 - Getting all loaded chunks within a 24 chunk radius (max radius of a beacon), and looping through their TEs determining whether they are a beacon and within range

Neither are especially appealing, but I guess are still worth looking into.

KnightMiner:
I am not sure how much it fits with turtles lore wise given they are machine rather than living, plus they are missing a few of the stats beacons boost. Might make more sense to have something dedicated to turtles instead boosting them.

lupus590:
So having them equip a beacon block gives boosts?

Wojbie:
> reduce the fuel cost (which would make fuel values floats which might cause issues. Maybe this isn't the best example).
You could keep fuel integer and just have random chance proportional to haste level for movement event not to use fuel.

KnightMiner:
I was thinking closer to turtle beacon like block to increase efficiency of turtles in the area. Easier to code since it is our own block, but probably would not fit best in base CC.

lupus590:
Random idea, in my opinion definitely not fitting to CC, turtles 'worship' this beacon block by spending their fuel points and give a boost to themselves and other nearby turtles.

Another idea, vanilla beacon but replace the metal blocks with computers/turtles/redstone blocks.

SquidDev:
I feel this is very quickly becoming a separate add on mod, and not something which really fits within CC itself.

lupus590:
@SquidDev agreed, closing

Perhaps if someone wants to make a CC based magic addon mod?

PR #522, opened 2018-02-16
[WIP] Turtle events

SquidDev:
This makes most turtle actions fire an event describing the action. Other mods may then listen to the event, cancelling it or changing the behaviour. Examples of things which may be done:

 - Prevent turtles from attacking specific mobs or players.
 - Block turtles from sucking from inventories (but still allow sucking item entities).
 - Add additional data to `.inspect*()` methods.
 - Disable any turtle action.

Most of these serve little purpose in the main mod, but do allow for greater flexibility for other mods. Whilst this technically works as is, I'm marking "WIP" as I'd like to get some thoughts on the whole implementation. It may be that this whole system is overkill and we can build something much simpler but equally powerful.

I'd also like to flesh out the event system a little more. Potentially adding the following:

 - Support for refuelling from non-fuel sources (and blocking from existing ones).
 - Control over turtle animation/action speed (as mentioned in #276).
 - Add additional data to `.getItemDetail()` method.
 - Allow reading/writing ItemStack NBT (and turtle upgrade NBT) when equipping upgrades. This would allow items to preserve information when being equipped and unequipped.

### DIFF
--- a/src/main/java/dan200/computercraft/ComputerCraft.java
+++ b/src/main/java/dan200/computercraft/ComputerCraft.java
@@ -7,6 +7,7 @@
 package dan200.computercraft;
 
 import com.google.common.base.CaseFormat;
+import com.google.common.base.Converter;
 import dan200.computercraft.api.filesystem.IMount;
 import dan200.computercraft.api.filesystem.IWritableMount;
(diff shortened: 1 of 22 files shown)


thatcraniumguy:

PR #523, opened 2018-02-24
Use an interface instead of method reflection to call the API

SquidDev:
This makes it easier to verify at compile time that API methods are valid. I also think it makes the API code substantially easier to read, but that may just be me.

### DIFF
--- a/src/main/java/dan200/computercraft/ComputerCraft.java
+++ b/src/main/java/dan200/computercraft/ComputerCraft.java
@@ -422,11 +422,6 @@ public class ComputerCraft
         }
     }
 
-    public static String getVersion()
-    {
-        return "${version}";
-    }
-
     public static boolean isClient()
     {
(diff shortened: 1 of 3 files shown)


theoriginalbit:
> I also think it makes the API code substantially easier to read, but that may just be me.

Definitely a much better way to do it! Though if ComputerCraft is using Java 8 then it could be made a little cleaner with Optionals.

apemanzilla:
> Though if ComputerCraft is using Java 8 then it could be made a little cleaner with Optionals.

It could be made even cleaner on top of that using Guava's [`Suppliers.memoize`](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/base/Suppliers.html#memoize-com.google.common.base.Supplier-) (Minecraft/forge already has Guava as a dependency, right)

thatcraniumguy:

PR #488, opened 2017-11-09
Localize ComputerCraft for es_ES

daelvn:
es_ES: Español de España (Castellano)
I created a .lang file for es_ES, it's my native language, so I thought I could do it.
Although untested, it should work fine, just as en_US works.

PR #527, opened 2018-03-05
Localize ComputerCraft for fr_fr

absolument:
Following #488 
Here is my translation in french fr_fr


thatcraniumguy:

PR #528, opened 2018-03-13
Add recipes to upgrade computers

SquidDev:
image

This adds recipes to upgrade computers, turtles and pocket computers using gold ingots. Computers and pocket computers require 7 gold ingots, whilst turtles require 14 (in the form of 1 gold block and 5 ingots). A couple of things worth noting:

 - These recipes are currently shaped. I'd be open to converting it to be shapeless if people would prefer it.
 - The recipes are largely data-driven, which hopefully makes it easier for people to change them - especially when 1.13 (and so data packs) drop.

This also converts the current `TurtleRecipe` over to use the same system: changing the item rather than the family. However, this now requires the computer item to be an actual computer - previously one could use a turtle or pocket computer. This behaviour feels a little strange to me, but I'm happy to revert to that if desired.


thatcraniumguy:

PR #536, opened 2018-04-04
Make wget automatically determine the file name.

Luca0208:
Sometimes when using wget it is pretty annoying having to write the file name again.
This PR makes that unnecessary by automatically determining the file name.
This is done by stripping all trailing slashes and afterwards taking everything after the last remaining slash.
This is either the domain name if used like this: https://domain.tld or the file name if used liked this: https://domain.tld/folder/file.ext

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/help/wget.txt
+++ b/src/main/resources/assets/computercraft/lua/rom/help/wget.txt
@@ -1,5 +1,7 @@
 wget is a program for downloading files from the internet. This is useful for downloading programs created by other players.
+If no filename is specified wget will try to determine the filename from the URL by stripping any anchors, parameters and trailing slashes and then taking everything remaining after the last slash.
 The HTTP API must be enabled in ComputerCraft.cfg to use this program.
-
 ex:
 "wget http://pastebin.com/raw/CxaWmPrX test" will download the file from the URL http://pastebin.com/raw/CxaWmPrX, and save it as "test".
+"wget http://example.org/test.lua/?foo=bar#qzu" will download the file from the URL http://example.org/test.lua/?foo=bar#qzu and save it as "test.lua"
+"wget http://example.org/" will download the file from the URL http://example.org and save it as "example.org"


SquidDev:
Thank you! This is something I've been meaning to do for a while. Instead of using a loop, it might be a little cleaner to use `gsub` and `match` though:

```lua
-- Trims everything after ? or # (means foo#bar or foo?bar=qux reduce to foo), 
-- then trims trailing `/`
sUrl = sUrl:gsub( "[#?].*", "" ):gsub( "/+$", "" )

-- Find everything from the last / to the end of the string
return sUrl:match( "/([^/]+)$" ) or sUrl
```

Note I've only done some limited testing, so this may be more susceptible to breakage.

Luca0208:
foo#bar should be reduced to bar, however I'm not sure if foo?bar=s should be reduced to foo.
Sometimes these URL Parameters matter.

As for the pattern I tested them against a bunch of edge cases and they seem to work.
The `or sUrl` in the last line isn't even needed because the pattern will match the / of "http://" or "https://" which is required for a valid URL.

apemanzilla:
`=` and `?` are forbidden characters on Windows filesystems.

Luca0208:
That's a good point. I don't have access to CC in MC right now, but both CCEmuX and CCEmuRedux also emit `?` in filenames `=` seems to be allowed(However I'm on a mac right now, not sure how it behaves on windows)

apemanzilla:
On NTFS (the most commonly used Windows filesystem) it will fail.

Luca0208:
Ok it will now strip anchors and url parameters and I also applied the suggestion of SquidDev to use patterns and gsub/find

src/main/resources/assets/computercraft/lua/rom/programs/http/wget.lua
```diff
@@ -15,7 +15,12 @@ if not http then
     printError( "Set http_enable to true in ComputerCraft.cfg" )
     return
 end
- 
+
+local function getFilename( sUrl )
+    sUrl = sUrl:gsub( "[#?].*" , "" ):gsub( "/+$" , "" )
+    return sUrl:match( "/([^/]+)$" ) or " "
```
Luca0208: I'm not sure if the `or " "` should be here or further down(See comment on line 52)

src/main/resources/assets/computercraft/lua/rom/programs/http/wget.lua
```diff
@@ -43,7 +48,7 @@ end
  
 -- Determine file to download
 local sUrl = tArgs[1]
-local sFile = tArgs[2]
+local sFile = tArgs[2] or getFilename(sUrl)
 local sPath = shell.resolve( sFile )
```
Luca0208: This could be replaced by `shell.resolve( sFile or " ")`

SquidDev:
On my machine (Windows), `fs.exists(" ")` always returns true, as it's parsed as a directory. Which means `wget google.com` will result in `File already exists` instead of `URL malformed`.

I think it would be better to move the `http.checkURL` call before the file name extraction. This way one doesn't need the `or " "` at all, which makes it substantially less ugly. It might produce a tiny bit of delay, as this may require a DNS lookup, but it should be imperceptible (<50ms).

Luca0208:
Ok in that case I will put the `http.checkURL` before the filename extraction.

Luca0208:
I just noticed that the help page of wget should be changed to reflect this change. Going to do that this evening.

src/main/resources/assets/computercraft/lua/rom/programs/http/wget.lua
```diff
@@ -40,10 +36,20 @@ local function get( sUrl )
...
- 
+
 -- Determine file to download
 local sUrl = tArgs[1]
-local sFile = tArgs[2]
+
+--Check if the URL is valid
+local ok, err = http.checkURL( sUrl )
+if not ok then
+    if err then
```
SquidDev: I wonder if it'd be worth doing `printError( err or "Failed. ")` instead. Though `err` shouldn't ever be nil, so neither is _really_ needed.

Luca0208: If this is done it would probably be better to use "Invalid URL" or something like that instead of "Failed." to distinguish it from the "Failed." message caused when the download failed.

thatcraniumguy:

PR #537, opened 2018-04-07
Make the arguments of os.epoch case-insensitive

Luca0208:
I don't see any reasons why the arguments of os.epoch should be case-sensitive.
This could be a problem especially with UTC when people try `os.epoch("UTC")`, but I think it would also make sense for "local" and "ingame" to be checked case-insensitive.


SquidDev:
It might be worth doing `.toLowerCase(Locale.ROOT)` instead (or `Locale.ENGLISH`) so you don't get strange behaviour on non-English systems. This change should probably be made to `os.time` and `os.day` as well (oh goodness, who thought the time changes were a good idea?).

thatcraniumguy:

PR #540, opened 2018-04-16
Add additional method stubs to the TurtlePlayer

### LINKED ISSUE #539: turtle.place() caused server crash
There was rail with full rail chest with various garbage in it in front of the turtle and 16 vanilla glass panes in turtle's selected slot. I've called turtle.place() from lua console and expected to see false and some kind of error message as second return value in lua console in game. But server just crashed.

Did the same in single player. Client crashed too. Can someone repeat this, please?

SquidDev:
Forge's default fake player implementation doesn't override all methods which use the connection. As it is not set, we get an NPE and thus crash the server. We simply stub those methods out ourselves to prevent such an issue.

Closes #539


JasonTheKitten:
This seems smart. Why did nobody do this before?

SquidDev:
@JasonTheKitten Because they really shouldn't be needed - ideally these methods would be stubbed in Forge instead. The alternative would be to do what Plethora does and provide a "fake" connection, which just voids all packets.

thatcraniumguy:

PR #541, opened 2018-04-23
Remove upper bound on how many bytes/characters can be read

SquidDev:
The limit was added to prevent people creating arbitrarily large buffers (for instance, `handle.read(2^31)` would create a 2GB char array). For "large" counts, we now read in blocks of size 8192, adding to an extendable buffer.

See also SquidDev-CC/CC-Tweaked#29


thatcraniumguy:

PR #543, opened 2018-05-08
amend typo in keys.lua

hugeblank:
Skimming through the CC API for an emulator project and found this bug rendering the game "literally unplayable". I decided to take this game-breaking issue into my own hands and repair it. You're welcome CC.

In other words this is the worst reason to receive a contributor tag.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/keys.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/keys.lua
@@ -16,7 +16,7 @@ local tKeys = {
 	"comma",	"period",	"slash",	"rightShift","multiply",	-- 51
 	"leftAlt",	"space",	"capsLock",	"f1",		"f2",			-- 56
 	"f3",		"f4",		"f5",		"f6",		"f7",			-- 61
-	"f8",		"f9",		"f10",		"numLock",	"scollLock",	-- 66	
+	"f8",		"f9",		"f10",		"numLock",	"scrollLock",	-- 66	
 	"numPad7",	"numPad8",	"numPad9",	"numPadSubtract","numPad4",	-- 71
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/apis/keys.lua
```diff
@@ -53,6 +53,7 @@ for nKey, sKey in pairs( tKeys ) do
 	keys[sKey] = nKey
 end
 keys["return"] = keys.enter
+keys.scollLock = keys.scrollLock
```
lupus590: You might as well change this to `keys["scrollLock"] to be consistent with the line above.

SquidDev: I believe it's only ["return"]` because that is a Lua keyword, and so `keys.return` is invalid syntax.

KnightMiner: Probably should add a comment here explaining why the old name is kept, something like "typo kept to avoid a breaking change"

hugeblank: I thought about it, I wasn’t sure if it was worth commenting though. I’ll add it when I get the chance.

hugeblank:
Found another typo, you can now cîrcûm _FLÊX_ (correctly) on all your friends.

src/main/resources/assets/computercraft/lua/rom/apis/keys.lua
hugeblank: @KnightMiner I should mention I did this in the most recent commit.

thatcraniumguy:

PR #550, opened 2018-06-14
Socket Support

JasonTheKitten:
Adds socket support.

Here is an example script:

```
local host, port = socket.lookup("https://google.com/")
local sock = socket.open(host, port, true)
sock.write("GET /search?q=Google HTTP/1.1\n\n")
print(sock.readAll())
```

Also supports ASYNC requests

There is a small minor bug with sock.length and sock.clear, where Java seems to divide the output into buffers, but it does not matter too much.

### DIFF
--- a/src/main/java/dan200/computercraft/ComputerCraft.java
+++ b/src/main/java/dan200/computercraft/ComputerCraft.java
@@ -121,6 +121,7 @@ public class ComputerCraft
     };
     
     public static boolean http_enable = true;
+    public static boolean socket_enable = true;
     public static AddressPredicate http_whitelist = new AddressPredicate( DEFAULT_HTTP_WHITELIST );
     public static AddressPredicate http_blacklist = new AddressPredicate( DEFAULT_HTTP_BLACKLIST );
     public static boolean disable_lua51_features = false;
@@ -202,6 +203,7 @@ public class ComputerCraft
         public static Property http_enable;
         public static Property http_whitelist;
         public static Property http_blacklist;
+		public static Property socket_enable;
         public static Property disable_lua51_features;
         public static Property default_computer_settings;
         public static Property logPeripheralErrors;
@@ -283,6 +285,7 @@ public class ComputerCraft
                 Config.http_whitelist.setValues( currentProperty.getString().split( ";" ) );
             }
         }
+        
         Config.http_whitelist.setComment( "A list of wildcards for domains or IP ranges that can be accessed through the \"http\" API on Computers.\n" +
             "Set this to \"*\" to access to the entire internet. Example: \"*.pastebin.com\" will restrict access to just subdomains of pastebin.com.\n" +
             "You can use domain names (\"pastebin.com\"), wilcards (\"*.pastebin.com\") or CIDR notation (\"127.0.0.0/8\")." );
@@ -292,6 +295,11 @@ public class ComputerCraft
             "If this is empty then all whitelisted domains will be accessible. Example: \"*.github.com\" will block access to all subdomains of github.com.\n" +
             "You can use domain names (\"pastebin.com\"), wilcards (\"*.pastebin.com\") or CIDR notation (\"127.0.0.0/8\")." );
 
+        
+        Config.socket_enable = Config.config.get( Configuration.CATEGORY_GENERAL, "socket_enable", socket_enable );
+        Config.socket_enable.setComment( "Enable the \"socket\" API on Computers (HTTP must also be enabled)" );
+        
+        
         Config.disable_lua51_features = Config.config.get( Configuration.CATEGORY_GENERAL, "disable_lua51_features", disable_lua51_features );
         Config.disable_lua51_features.setComment( "Set this to true to disable Lua 5.1 functions that will be removed in a future update. Useful for ensuring forward compatibility of your programs now." );
 
@@ -364,6 +372,7 @@ public class ComputerCraft
         http_enable = Config.http_enable.getBoolean();
         http_whitelist = new AddressPredicate( Config.http_whitelist.getStringList() );
         http_blacklist = new AddressPredicate( Config.http_blacklist.getStringList() );
+        socket_enable = Config.socket_enable.getBoolean();
         disable_lua51_features = Config.disable_lua51_features.getBoolean();
         default_computer_settings = Config.default_computer_settings.getString();
         logPeripheralErrors = Config.logPeripheralErrors.getBoolean();
--- a/src/main/java/dan200/computercraft/core/apis/SocketAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/SocketAPI.java
@@ -21,6 +21,7 @@ import static dan200.computercraft.core.apis.ArgumentHelper.*;
 public class SocketAPI implements ILuaAPI, IAsyncObject {
     private final IAPIEnvironment m_apiEnvironment;
     private final List < Socket > m_socks;
+	private AsyncAction m_queue;
 	
 	private String[] methods = {
         "open",
@@ -29,9 +30,8 @@ public class SocketAPI implements ILuaAPI, IAsyncObject {
     };
 
     public SocketAPI(IAPIEnvironment environment) {
-		new AsyncAction().startAsyncAction();
         m_apiEnvironment = environment;
-        m_socks = new ArrayList < Socket > ();
+        m_socks = new ArrayList<Socket>();
     }
 
     @Override
@@ -42,13 +42,16 @@ public class SocketAPI implements ILuaAPI, IAsyncObject {
     }
 
     @Override
-    public void startup() {}
+    public void startup() {
+		m_queue = new AsyncAction("async_socket");
+	}
 
     @Override
     public void advance(double _dt) {}
 
     @Override
     public void shutdown() {
+		m_queue.clear();
         synchronized(m_socks) {
             for (Socket sock: m_socks) {
                 try {
@@ -57,7 +60,6 @@ public class SocketAPI implements ILuaAPI, IAsyncObject {
             }
             m_socks.clear();
         }
-        AsyncAction.clear();
     }
 
     @Nonnull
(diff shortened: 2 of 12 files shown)


apemanzilla:
I think websockets (as implemented by #395 or [CC-Tweaked](https://github.com/squiddev-cc/cc-tweaked)) make more sense, being easier to use, having better security (`wss` out of box) and still being powerful for real-time communication.

JasonTheKitten:
`.getCertificates` is so that a program can know or notify the user how secure the connection is
Perhaps I could remove `.length` and `.clear`, because they are buggy anyways, but I feel it would be best left off to be removed on the lua side and not removed completely from the java side

JasonTheKitten:
> I think websockets (as implemented by #395 or CC-Tweaked) make more sense, being easier to use, having better security (wss out of box) and still being powerful for real-time communication.

I didn't like how CCTweaked didn't let you make your own requests... One of the reasons I wrote this
Because I don't think many sites support websockets (Tried them on google, did not work), but most site support TCP sockets

I've been wanting a raw socket implementation for two years now, and, well, here it is

JasonTheKitten:
Closing this for a little, I will reopen it soon

SquidDev:
> I didn't like how CCTweaked didn't let you make your own requests... One of the reasons I wrote this Because I don't think many sites support websockets (Tried them on google, did not work), but most site support TCP sockets

I suspect we're thinking of using TCP sockets for different things. From what I can tell, you're trying to emulate HTTP requests, which seems rather pointless as we have a rather decent library for that already.

TCP sockets (and websockets) are far more useful when you want to keep connections open for a long period. This means you don't have to poll a website for information - it's streamed directly to you. A great example of this is Krist - you can connect via websocket to the API and it'll send you a message every time there's a transaction.

JasonTheKitten:
Sockets can be used to communicate with a server that uses a constant input stream. In addition, other than the fact that `socket.lookup` does not seem to support wss schemes, we can emulate websockets

apemanzilla:
Websockets can also be used for constant communication, and while re-implementing websockets using a TCP layer is technically possible, it would result in a variety of issues:

- LuaJ is already slow, implementing a high-level protocol like websockets (which are themselves built on top of HTTP) is only going to make it worse
- Security pretty much goes out the window, unless you want to implement modern TLS yourself and then copy all the necessary certificates for HTTPS (and by extension, WSS) to the computers
- While we're on the topic of reimplementing things, you'd also need to either implement or expose DNS, or use an external service to resolve domains

I could go on and on...

SquidDev:
Guys, this shouldn't boil down to some "TCP vs Websockets" debate, as I don't see a reason [we can't have both](https://squiddev.cc/r/both-why.gif). One is easier to use, one is more powerful.

I think what really needs to be discussed, is how we want people to use it, and thus how we want the API to look. One of the complications that TCP has is there is no distinct idea of a packet or message, so we can't really do a rednet-esque implementation like #395 does. I think the best solution is something akin to what this PR does, albeit with a few more events thrown in:


<details>
<summary>A potential <code>socket</code> API, put in a spoiler as it's rather large</summary>

### `socket` library
#### `socket.connect(host:string, port:number[, secure:boolean]):table|nil,string`
Creates a connection to a given host/port combination (note host can be a domain name or IP address). Behind the scenes, this will delegate to some async method which queues a `socket_success` or `socket_failure` once connection succeeds/fails.

### Socket handles

#### `handle.read([len:number]):string`
Read at most `n` bytes from the socket or, if no `n` is specified, as many bytes as are available.

#### `handle.write(msg:string)`
Write `msg` to the socket (as a set of bytes). 

**Issue:** What happens when not all of `msg` can be sent for whatever reason? How do we signal this to the user, and how can we recommend they try again.

#### `handle.close()`
'nough said

#### Event `socket_readable`: `host:string, port:number`
Fired when incoming data arrives, denoting there is information which can be read from the socket.

**Issue: ** What happens if you've multiple connections to the same host/port combination? Should we give each socket some unique identifier instead?

#### Event `socket_closed`: `host:string, port:number`
Fired when the socket is closed. I'm less sure/fussed about this one.

---

This was knocked up pretty early in the morning, so there's definitely some issues, but I think it's a good starting point. Obviously it's possible to add additional methods to socket handles, but I think this is a nice skeleton.

That being said, I do welcome radically different proposals. While I think an "event driven system" is necessary, I do think I run the risk of just re-inventing CCTweaks :/.
</details>

apemanzilla:
I was specifically responding to the idea of re-implmenting websockets using a lower level API.

JasonTheKitten:
I'm going to finish my edits to this problem when I get a chance and then reopen this

JasonTheKitten:
I do not understand how you could have just one thread with non-blocking.

I tried out a single thread but found that that meant you have to wait for one task to finish before you perform the next action, which is especially troublesome to debug if you accidentally enter an infinite loop...

Here is the new code I was trying out (GitHub made me change the file extension):
[AsyncAction.java.log](https://github.com/dan200/ComputerCraft/files/2108562/AsyncAction.java.log)

JasonTheKitten:
I've commited a few modifications

apemanzilla:
There are already many applicable concurrency features in the Java standard library (i.e. `Future`, `Executor`, and various locks/latches), at least use those.

JasonTheKitten:
Why do we need all this complicated stuff, why can't we just create threads and then delete them when done, like I originally did...

I wrote the AsyncAction class because I originally did not include async methods in my initial code, but changed my mind, and I _didn't_ want to have to deal with all this complicated stuff that will turn my code into spaghetti.

Lignum:
> Why do we need all this complicated stuff, why can't we just create threads and then delete them when done, like I originally did, before I modified it to only use a single thread...

Because starting threads is expensive, and can be abused by malicious players, unless you have a thread pool, which we don't.

> [..], and I didn't want to have to deal with all this complicated stuff that will turn my code into spaghetti.

If modeled properly, it won't turn your code into spaghetti. That applies to basically everything.

JasonTheKitten:
This is mostly ready, just that one thing about the threads...

JasonTheKitten:
I tried a thread pool, the threads wouldn't run... );

JasonTheKitten:
OK, GitHub should recognize my latest commit when I reopen this, fixed thread system and socket.lookup

apemanzilla:
The point of using `Future` and other concurrency features is to avoid implementing your own equivalents (`AsyncAction` and co). `FutureTask` should be used for this, and can be submitted directly to an `Executor`.

JasonTheKitten:
> The point of using Future and other concurrency features is to avoid implementing your own equivalents (AsyncAction and co). FutureTask should be used for this, and can be submitted directly to an Executor.

This is easier. I already had my thingy set up to return the values in the method, so it was easiest to do this than create extra methods. Also, if enough functions were to use AsyncAction, it would eventually save us some code.

zardyh:
If enough functions were to use `Future`, it'd save us all of that code and then all of the code for `AsyncAction`!

apemanzilla:
Your `AsyncAction` class is nothing but a wrapper for a `ThreadPoolExecutor` anyways, and if you actually used the features provided by `FutureTask`, it would be completely unnecessary. `Counter` is nothing more than a gimped `AtomicInteger`, minus the thread safety (speaking of which, almost none of your code is thread-safe). Your implementation right now will create a new `ThreadPoolExecutor` for every single computer, each capable of spawning many threads, which as previously mentioned, is quite expensive and abusable. This is a very bad™ idea.

JasonTheKitten:
Only up to 2048 threads per computer. That was intentional

apemanzilla:
> Only up to 2048 threads per computer. That was intentional

That has to be the most absurd statement I've ever heard. 2048 threads is enough to _completely cripple_ a (real) computer, and that's before you start placing more of them down. There should be at _most_ only a couple threads used for this, globally, and with minimum priority.

apemanzilla:
> The HTTP API uses Integer.MAX_VALUE for it's thread count. This is much lower.

That's a different issue that should also be addressed, but at the very least it uses one `ThreadPoolExecutor` instead of one for every single computer.

JasonTheKitten:
Are you saying that I should revert to 1 thread?
https://github.com/JasonTheKitten/ComputerCraft/blob/08d221ef3fc7adcb48b365f692b33377e0899e27/src/main/java/dan200/computercraft/core/apis/socket/AsyncAction.java

apemanzilla:
No, I'm saying you should properly make use of `ThreadPoolExecutor` and `FutureTask`: One shared instance of the executor, sane thread limits (_not_ 2048), using methods provided by `FutureTask` rather than a weird wrapper system, and ditch `AsyncAction` entirely.

JasonTheKitten:
Today's Windows 8.1 cpu's are fast. They should _easily_ be able to handle 2048 very small threads

apemanzilla:
> I really don't see what you have against AsyncWrapper. I thought it was better than futures, and even thought about changing the whole HTTP API to use (a modified version of) them instead of futures

It's an unnecessary layer of abstraction, considering `CompletableFuture` and `Execturo` provide all the necessary facilities.

SquidDev:
> The HTTP API uses `Integer.MAX_VALUE` for it's thread count. This is much lower.

Yeah, I'm lazy and didn't want to choose a good maximum. Sorry. I honestly don't think the thread limit is a stopping factor, as both the HTTP, websocket and socket functionality is highly abuseable in other ways.

> Are you saying that I should revert to 1 thread?

ape has rather lead you up the wrong alley here. It's not that what he's said is incorrect, it's just that it's the correct solution to an entirely different problem. The problem with having one thread for all of them (or 10, or 100) is that reading from one socket will block all the others. The only solution here is to have one thread per socket, which then leads onto the same sort of issues as [C10k](https://en.wikipedia.org/wiki/C10k_problem). 

The solution here, as I may have mentioned before, is to look into non-blocking IO. It's possible to get this running with 4 threads _across the entire system_. This way you don't need to worry about resource limits, nor one socket blocking the others.

JasonTheKitten:
Ok, but quick question, does the nio class support SSL?

Hmm, GitHub is saying I cannot comment, I'll try again

Lignum:
> Ok, but quick question, does the nio class support SSL?

[Yes](https://stackoverflow.com/questions/9118367/java-nio-channels-and-tls).

JasonTheKitten:
Do whatever you want with this, feel free to fork it if you really want changes (I will be notified), I don't want to deal with this any more

apemanzilla:
You should probably close the PR then

PR #561, opened 2018-07-07
Give adventure some TLC

atenfyr:
Just a few changes to adventure that address a couple bugs and some grammar issues.


Selim042:
I like having the fake score at the end still as a reference to the [long-standing Minecraft bug](https://bugs.mojang.com/browse/MC-66).

atenfyr:
That's a good point. I'll put that back.

PR #566, opened 2018-08-12
A couple of small improvements to CraftOS

SquidDev:
- Make window.reposition's argument validation a little more strict. Previously it would accept `window.reposition(x, y, width)` (no height argument), just not act upon it.
 - Use select instead of table.unpack within `pastebin run`.
 - Use `parallel.waitForAny` instead of `waitForAll` within the dance program.
 - Pipe the entire help file into `textutils.pagedPrint`, rather than doing it line by line.
 - Remove bytecode loading disabling from bios.lua. This never worked correctly, and serves little purpose as LuaJ is not vulnerable to such exploits.

I realise many of these changes are incredibly minor, and so don't really seem worth it. _However_, CraftOS is a great reference to link people to when they're looking for help, and so it's always good if we can show "best practices" for using CC & Lua's features.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/window.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/window.lua
@@ -421,8 +421,10 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     function window.reposition( nNewX, nNewY, nNewWidth, nNewHeight )
         if type( nNewX ) ~= "number" then error( "bad argument #1 (expected number, got " .. type( nNewX ) .. ")", 2 ) end
         if type( nNewY ) ~= "number" then error( "bad argument #2 (expected number, got " .. type( nNewY ) .. ")", 2 ) end
-        if nNewWidth ~= nil and type( nNewWidth ) ~= "number" then error( "bad argument #3 (expected number, got " .. type( nNewWidth ) .. ")", 2 ) end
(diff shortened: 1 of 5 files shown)


src/main/resources/assets/computercraft/lua/rom/apis/window.lua
```diff
@@ -421,8 +421,10 @@ function create( parent, nX, nY, nWidth, nHeight, bStartVisible )
     function window.reposition( nNewX, nNewY, nNewWidth, nNewHeight )
         if type( nNewX ) ~= "number" then error( "bad argument #1 (expected number, got " .. type( nNewX ) .. ")", 2 ) end
         if type( nNewY ) ~= "number" then error( "bad argument #2 (expected number, got " .. type( nNewY ) .. ")", 2 ) end
-        if nNewWidth ~= nil and type( nNewWidth ) ~= "number" then error( "bad argument #3 (expected number, got " .. type( nNewWidth ) .. ")", 2 ) end
-        if nNewHeight ~= nil and type( nNewHeight ) ~= "number" then error( "bad argument #4 (expected number, got " .. type( nNewHeight ) .. ")", 2 ) end
+        if nNewWidth ~= nil or nNewWidth ~= nil  then
```
Rems19: Isn't that a typo?

SquidDev: It is, good catch!

PR #567, opened 2018-08-14
Upgrade the event

TagFreettle:


Wojbie:
Any screenshots? Description?

SquidDev:
I'll echo Wojbie here - the commit messages do contain _some_ details, but it's quite hard to tell what something will look like. See #390 and #396 for some examples of other PRs which touch edit's functionality.

A couple of other points on the code:
 - There's been a lot of aesthetic changes to the code which make it rather hard to tell what's _actually_ changed. For instance, lots of variables have been renamed and whitespace has been rather messed around. While the existing whitespace is definitely inconsistent, fixing that is best left for another PR!
 - I'm really not sure what `repairString` is for. I can see what it does, but I don't understand the need.
 - Similarly, `eliminatePoint` - this seems to do little more than `tostring` the input. I suspect you're trying to strip the decimal point.
 - There's some behaviour which I'm sure makes sense, but seems a little odd, such as the status scrolling code. It may be worth adding a few comments within the code to explain how it works.

PR #568, opened 2018-08-14
worldfs - shared drives between all computers

timia2109:
Worldfs brings you additional shared drives between all computers which a running at the same Minecraft world. It mounts the worldfs drive as a folder into the computer of your choice.

The files are located at ```.../saves/{worldname}/computer/worldfs/{channelname_hash}```

Only when you know the channelname you can mount the drive, because the name is hashed with SHA-256 and some string transformation. ```"Test"``` would be ```"vUyfT26nd4jk0abSGCur9QIPxn8Wp0GL"```

worldfs is disabled by default and can enabled in the config file.

New API:
 - worldfs.list() : returns table, contains WorldFSWrapper classes
 - worldfs.mount( channelname: string, mountpath: string )
 - worldfs.unmount( channelname: string )

New Program:
 - worldfs.lua

WorldFSWrapper contains the following Lua methods:
 - getChannel() : returns string with the channelname
 - getPath() : returns string, the local path to this worldfs drive on this computer
 - getKey() : returns string, the hash of the channelname

----
I'm open for comments and improvements.

### DIFF
--- a/src/main/java/dan200/computercraft/ComputerCraft.java
+++ b/src/main/java/dan200/computercraft/ComputerCraft.java
@@ -154,6 +154,9 @@ public class ComputerCraft
 
     public static int maxNotesPerTick = 8;
 
+    public static int worldfsLimit = floppySpaceLimit;
+    public static boolean worldfs_enable = false;
+
     // Blocks and Items
     public static class Blocks
     {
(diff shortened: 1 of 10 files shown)


SquidDev:
This seems something which would be better done as an external mod (using something like #491 ). I'd personally consider a little too "magical" to be part of base CC - it rather overshadows any existing system, such as rednet - you don't even need a modem!

PR #575, opened 2018-09-21
Rewrite file systems to use ByteChannels

SquidDev:
This replaces the existing `IMount` `openFor*` method with `openChannelFor*` ones, which return an appropriate byte channel instead. I've put a little more detail in the commit message about the implications of this, but there's a couple of user-facing benefits:

 - Add support for seekable handles. This works for any filesystem or HTTP handle opened in _binary_ mode.
 - Rewrite the io library to more accurately emulate PUC Lua's
   implementation. This passes a large portion of the the test suite - the remaining issues mainly involve their use of `os.date` and `setvbuf`, neither of which are implemented.


SquidDev:
Possibly worth noting that seekable binary handles do not work for Jar mounts - namely the `rom` and any treasure disks. 

It would be possible to implement this by reading the entire contents of the file into memory and exposing it as an `ArrayByteChannel`, but I think it's worth discussing whether it's worth it.

PR #577, opened 2018-09-23
Add .getCursorBlink to monitors and terminals

### LINKED ISSUE #576: term.getCursorBlink
To go with `term.getCursorPos` and `term.getTextColor` and `term.getBackgroundColor`, this would be good.

It would return whether the cursor is set to blink or not, obviously.

SquidDev:
Adds it to the `term` API, `window` API and monitor peripherals. Closes #576.


SquidDev:
@Wilma456 We need to implement this method on all terminal redirects, otherwise you may get `Redirect object is missing method getCursorBlink` errors. We do the same for most getters - see `.getBackgroundColour`, `.getTextColour`.

PR #578, opened 2018-10-10
Change speakers to use the SPacketCustomSound packet instead

SquidDev:
The method to register new `SoundEvent`s is private, which means that few (if any) mods actually register them. Consequently, one could not use the speaker to play any modded sound, as they weren't registered on the server side.

Using `SPacketCustomSound` does mean we can no longer determine if a sound exists or not, but I think a price I'm willing to pay in order to allow playing modded sounds.


lupus590:
What happens if I try to play a sound that doesn't exist?

SquidDev:
It just returns true and does nothing on the client.

PR #588, opened 2019-03-24
Add some lua stuff

codescratchblocks:


lupus590:
We already have a hello world program, we don't need another hello world program

PR #589, opened 2019-04-09
Wang ting zheng patch 1

WangTingZheng:


TorinKwok:
Why did you do that (on master branch directly)? A partially translated README.md does not make sense.

PR #601, opened 2019-10-30
Change height checks to integrated MC method

Shad0wlife:
This effectively still checks if the target position is 0<=y<256, just as separate checks for 0 and world.getHeight() would do.
But by using this, mods like [CubicChunks](https://github.com/OpenCubicChunks/CubicChunks) which modify the world's height limits can change the result of the world.isOutsideBuildHeight() for easy compatibility.


vico93:
Could make a copy of this PR for [CC:Tweaked](https://github.com/SquidDev-CC/CC-Tweaked) too (if this bugfix already wasnt added there)?

SquidDev:
CC:T should have these patches already. That said, I've never actually tested whether it's sufficient to ensure compat with CubicChunks.

Shad0wlife:
You are right, thank you. I did not know about CC:Tweaked though, is there any web presence other than your github? When looking for the current version of CC, I did not find anything about it either.

